PR #261, opened 2015-01-14
Small Performance Optmizations

fsbatista:
Since performance is such a big thing in trading bots, I thought it would be nice to do some small code refactoring, like changing the variable declaration style (http://jsperf.com/variable-declaration5) and replacing if/else expressions for ternary operator. (http://jsperf.com/if-else-vs-arrays-vs-switch-vs-ternary/24). Let me know what you think!


fsbatista:
These are small modifications in the coding style that, if expanded towards the whole app, could result in interesting (even if small) gains of performance. I do agree that ternary operators are not as straightforward as if else statements though, it's a tradeoff.

askmike:
Hi @fsbatista,

First of all, thanks for the PR! Always great to see some new ideas.

As for the performance improvements, I think it is very dangerous to rely on micro benchmarks for profiling in most interpreted languages. See [this great talk](https://webrebels.23video.com/crooked-mirrors-of-performance-by) by mraleph (a V8 developer). So the speed improvements found in jsperfs might not even be applicable in Gekko because of the way V8 is able to optimize the code.

Gekko works in ticks: after everything is loaded and Gekko is runs ticks after any amount of time ([60 seconds after last tick on master branch](https://github.com/askmike/gekko/blob/master/core/tradeFetcher.js#L121), every 20 seconds [in the upcoming version](https://github.com/askmike/gekko/blob/candle-rewrite/core/heart.js#L42)).

If those numbers would still be 100% the same in Gekko, we could maybe slice of < 5ms every tick. If your goal is to make Gekko respond as fast as possible to market events, just is probably not the area we should look at.

During this tick we fetch new trade data from exchanges, and use that to calculate 1m candles, empty candles, and bigger candles. And we feed those to the trading methods. As for speeding things in Gekko up I think we should focus on:

1. Reducing this 20 seconds interval (in the end you just want Gekko to respond faster to market events). This mostly depends rate limits imposed by the exchanges. (Should be looked at at a per exchange basis).
2. Making ticks as fast as possible and making sure we defer everything like persisting data to disk to after we handed it over to the tradeAdvisor (make it some kind of priority object). In other words: make the time between getting new data and feeding it to the tradeAdvisor as short as possible.

For 2. I am pretty sure we could do it in pretty much half the time if use [streams](http://dailyjs.com/2013/04/01/streams-streams-streams/) instead of waiting the exchange to send all trade data before we start processing it. However by changing the 20 seconds number in point 1 to 10 seconds Gekko would still respond almost twice as fast to market events. However this should be properly benchmarked before we start refactoring [Gekko's  internals](https://github.com/askmike/gekko/blob/candle-rewrite/docs/internals/architecture.md) to streams.

AritzaSm:
i cant wait to see the new release ; )

fsbatista:
Thanks for the great feedback @askmike . That was a great talk. I'll keep playing with Gekko and see if I can contribute on anything else. Have you begun doing the benchmarking you talked about?

askmike:
No not yet, more important priorities for me are:

- Finishing new core (see the candle-rewrite branch)
- Stabilising everything (add tests, make sure everything can run on
*every* supported exchange for weeks without leaking memory)
- Solid TA-lib integration (stable API)
- Solid Backtesting

Performance is a very delicate subject in automated trading. The trading
methods Gekko supports are not focused on anything like HFT, but currently
act more on a hourly basis. The indicators try to detect trends in a way
where they lag behind[1] the realtime price movement (at minimum a couple
of hours) to "verify" the trend anyway, slicing of ms here will not make a
lot difference.

In that way Gekko is more like a human trader looking at charts every hour
instead of a real trading algo.

[1] "The reference line is a 9-day EMA of the signal line and, as such, is
designed to lag behind the signal line. " (
http://www.traderslog.com/insidemacd/ )

---

If you **really** want to have a bot that can respond as fast as possible I
would suggest to look at alternatives to nodejs: you will be doing
computation (from simply aggregating raw market data into more useful data,
to pattern detection math) everywhere in the chain. Nodejs is pretty slow
when it comes to computation.

PR #269, opened 2015-02-19
Implemented support for meXBT exchange.

williamcoates:
Currently only supports BTCMXN pair. Will add support for other pairs when volume picks up there.


askmike:
awesome thanks!

PR #291, opened 2015-12-16
Can add exchange bx.in.th Thailand Bitcoin

btcfan:
API Documentation and acceptable use from this friendly exchenge. Every south-west Asia connect.
https://bx.in.th/info/api/


askmike:
It looks like you didn't actually include any code in this PR, I created an issue regarding adding this exchange.

PR #294, opened 2016-02-02
Added new exchange Zaif.jp

naofum:


askmike:
I merged this is in anyway, assuming that's fine :)

PR #296, opened 2016-03-25
allow exchange providesHistory

thiagof:


thiagof:
testing with https://github.com/askmike/gekko/pull/298

tonycapone:
@thiagof  I'm trying to use the new poloniex exchange but it looks like it requires this PR to be merged in before it is functional. Is that correct?

thiagof:
Not sure.. this PR allows to fetch historical data, but poloniex should work without it.

Whats your problem?

tonycapone:
@thiagof   The default for poloniex in `exchanges.js` is `providesHistory: 'date',`, which causes gekko to blow up, because the supporting code isn't in master, currently. When I pull in your other PR, it works correctly.

thiagof:
True.. glad it is working for you

askmike:
The new version of gekko (0.2) uses a entirely different system for fetching data. The new version automatically fetches historical data (when the exchange has `providesHistory: 'date'`) like so:

    # run gekko with a required history of 100 of 5 min candles
    2016-06-03 14:55:10 (DEBUG):	Exchange has data spanning 30283 seconds, we need at least 30000 seconds.
    2016-06-03 14:55:10 (DEBUG):	Exchange returns more data than we need, shift required history to 100 .

thiagof:
Cool mike. Does Poloniex api compatible with the new history (date) gekko api?

askmike:
What do you mean exactly? New Gekko is able to fetch older history at poloniex if required now.

Issue #308, opened 2016-06-06
Update project's dependencies


bugron:
| Package | Current | Wanted | Latest |
| --- | --- | --- | --- |
| async | 0.2.10 | 0.2.10 | 2.0.0-rc.5 |
| bitstamp | 0.1.10 | 0.1.10 | 0.2.1 |
| btc-e | 0.0.16 | 0.0.16 | 1.0.5 |
| chai | 2.3.0 | 2.3.0 | 3.5.0 |
| lodash | 2.4.2 | 2.4.2 | 4.13.1 |
| moment | 2.4.0 | 2.4.0 | 2.13.0 |
| semver | 2.2.1 | 2.2.1 | 5.1.0 |
| sqlite3 | 3.1.3 | 3.1.3 | 3.1.4 |

raffus:
Hey guys,

I'm running gekko on backtest mode with no problems, until I try to run it in live mode.

When running in live mode, this error keeps comming up:

ekko encountered an error and can't continue

```
Error:

The plugin sqlite adapter expects the module sqlite3 to be installed. However it is not, install it by running: 

        npm install sqlite3@3.1.4 

Meta debug info:

Gekko version: v0.4.10
Nodejs version: v7.10.1
```

And when I try to install sqlite again:

```
# npm install sqlite3@3.1.4 

> sqlite3@3.1.4 install /home/infra/gekko/node_modules/sqlite3
> node-pre-gyp install --fallback-to-build

[sqlite3] Success: "/home/infra/gekko/node_modules/sqlite3/lib/binding/node-v48-linux-x64/node_sqlite3.node" is installed via remote
gekko@0.4.10 /home/infra/gekko
└── sqlite3@3.1.4  invalid

```

This does not happen when running in backtest mode.

Any idea?

Thanks

Herklos:
Hey !

Try to install it by
`npm install sqlite3` 
(newer version)

Worked for me ;)

raffus:
Thanks @Herklos.

The problem was that I was running "nodejs" instead of just "node".

Thanks once again.
BR.
R

doepat:
@raffus i have the same issue probably - what did you change exactly?

askmike:
@doepat the issue with sqlite3? Make sure you have the nodejs LTS version before trying to install it.

greenbigfrog:
@askmike is this still a issue? I believe this can be closed.

askmike:
[...]

PR #311, opened 2016-06-06
Update most dependencies to their latest versions

bugron:
Seems I don't brake anything with this update. Needs to be QA'd and tested properly though.
`Lodash` needs to be upgraded separately because this project relies on it largely and there is a problem with `.bindAll` after upgrade. I had to `bind` some methods explicitly and even after that I couldn't make the bot to work. I think this, particularly, is a good reason to refactor whole codebase to use ES2015.
Refers to https://github.com/askmike/gekko/issues/308.


askmike:
Awesome, will give some proper testing soon.

> Lodash needs to be upgraded separately because this project relies on it largely and there is a problem with .bindAll after upgrade. (...)  I think this, particularly, is a good reason to refactor whole codebase to use ES2015.

The current codebase styling relies heavily upon creating constructor objects (similar to classes), we could refactor everything to ES6 classes (with fat arrow methods for auto context binding). But honestly I don't see a lot of reasons of why we should do that anytime soon (sounds like a lot of work and I don't see any real advantage).

askmike:
This is now auto closed since I deleted the 0.2 branch (and unable to reopen). I will cherrypick this in after testing.

(So this means I also cannot delete the "master branch" until open PRs against that branch are resolved).

PR #313, opened 2016-06-07
Remove Makefile, run tests directly with npm scripts

### LINKED ISSUE #310: Make running test easier
Currently I have to install `make` to run tests but it's easier to move `./node_modules/.bin/mocha -u tdd --reporter spec` to npm scripts. `npm test` will still work and I don't think there will be problems with Travis CI too. Any reason not to do this?

bugron:
Closes #310.


askmike:
Awesome,  thanks!

Issue #288, opened 2015-11-04
node-gyp rebuild / error: redefinition of '_NanEnsureLocal'


cfrancomme:
Versions : "node":"4.2.1","npm":"3.3.9", bufferutil@1.1.0 This is happening straight after getting actual version from github

Anyone have had this issue with recent modules ?

> bufferutil@1.1.0 install /Users/x/Documents/myworkspace/gekko/node_modules/bufferutil
> node-gyp rebuild

  CXX(target) Release/obj.target/bufferutil/src/bufferutil.o
In file included from ../src/bufferutil.cc:16:
../../nan/nan.h:261:25: error: redefinition of '_NanEnsureLocal'
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
                        ^
../../nan/nan.h:256:25: note: previous definition is here
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Handle<T> val) {
                        ^
../../nan/nan.h:661:13: error: no member named 'smalloc' in namespace 'node'
    , node::smalloc::FreeCallback callback
      ~~~~~~^
../../nan/nan.h:672:12: error: no matching function for call to 'New'
    return node::Buffer::New(v8::Isolate::GetCurrent(), data, size);
           ^~~~~~~~~~~~~~~~~
/Users/x/.node-gyp/4.2.1/include/node/node_buffer.h:31:40: note: candidate function not viable: no known conversion from 'uint32_t' (aka 'unsigned int') to
      'enum encoding' for 3rd argument
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
                                       ^
/Users/x/.node-gyp/4.2.1/include/node/node_buffer.h:43:40: note: candidate function not viable: 2nd argument ('const char *') would lose const qualifier
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
                                       ^
/Users/x/.node-gyp/4.2.1/include/node/node_buffer.h:28:40: note: candidate function not viable: requires 2 arguments, but 3 were provided
NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
                                       ^
/Users/x/.node-gyp/4.2.1/include/node/node_buffer.h:36:40: note: candidate function not viable: requires 5 arguments, but 3 were provided
[...]

PR #315, opened 2016-06-08
Add most of missing npm packages

bugron:
`okcoincn` is required in `exchanges\okcoin.js` but doesn't even exist in npm registry.
npm couldn't install `node-xmpp`, it recommends to use `node-xmpp-client / node-xmpp-server / node-xmpp-component` instead.
`talib` package failed to install because of my not supported OS (Windows, obviously).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,9 @@
     "async": "0.2.x",
     "bitfinex-fork": "0.3.4",
     "bitstamp": "0.1.x",
+    "bitx": "^1.5.0",
     "btc-e": "0.0.x",
+    "btcchina": "0.1.x",
     "cexio": "0.0.x",
     "kraken-api": "0.1.x",
     "lakebtc_nodejs": "0.1.x",
@@ -24,8 +26,7 @@
     "poloniex.js": "0.0.6",
     "semver": "2.2.1",
     "sqlite3": "3.1.3",
-    "zaif.jp": "^0.1.4",
-    "btcchina": "0.1.x"
+    "zaif.jp": "^0.1.4"
   },
   "devDependencies": {
     "chai": "^2.0.0",


askmike:
As for exchanges:
- mtgox-apiv2 - I only want to leave the code there for historical purposes (maybe I should comment it out completely?). Mt. Gox was the biggest exchange but has been bankrupt / down since 2014. So loading the API wrapper doesn't do anything.
- mexbt: see [#288](https://github.com/askmike/gekko/issues/288#issuecomment-223810974)
- okcoin: I just found this code in one of the forks and it looks like it was usable. I have configured Gekko so that as soon as someone tries to enable okcoin (which isn't documented anyway) Gekko will log [this message](https://github.com/askmike/gekko/blob/0.2/exchanges.js#L264-L265). I also couldn't find the right npm dependency straight away, but as soon as we do we can test this existing code.

bitx & btcchina indeed need to be added.

As for plugins:

Gekko uses a custom way of handling dependencies for plugins. Plugins are very easy to add and leightweight, however their dependencies might not be (for example `talib` or `redis` on windows). Therefor Gekko uses [a custom way](https://github.com/askmike/gekko/blob/0.2/core/pluginUtil.js#L20-L45) of checking for the existence of these dependencies. The idea of plugins is that they can be enabled/disabled easily and core does not depend on any. If you are missing dependencies for a plugin that you have enabled in the config Gekko will tell you like so:

```
2016-06-08 09:20:57 (INFO): Gekko v0.2.0 started
2016-06-08 09:20:57 (INFO): I'm gonna make you rich, Bud Fox. 


2016-06-08 09:20:57 (INFO): Setting up Gekko in realtime mode
2016-06-08 09:20:57 (INFO): 
2016-06-08 09:20:57 (INFO): Setting up:
2016-06-08 09:20:57 (INFO):    SQLite Datastore
2016-06-08 09:20:57 (INFO):    Store candles, trades and advices in a SQLite database

 The plugin sqliteWriter expects the module sqlite3 to be installed. However it is not, install it by running: 

  npm install sqlite3@3.1.3 
```

bugron:
i didn't know about that custom handler, nice. `btcchina` actually was there, npm just sorted the list.
Amended this commit, removed `mexbt` and `mtgox-apiv2`.
And I think it's safe to remove all outdated code entirely because it's going to be in git history anyways.

askmike:
> i didn't know about that custom handler, nice.  

Thanks, so to be clear that means I rather not put all plugin dependencies (irc, mailjs, math-js, redis, etc.) in `package.json`. When running npm install in the gekko directory I rather have it only install dependencies for:
- Gekko core
- Plugins that are enabled on default (sqlite, etc)

bugron:
OK, honestly that custom handler is unusual for me (great idea, though), I prefer to install all deps at once to be able to use them whenever I want.
Amended one more time, only adds `bitx`.

askmike:
You really don't want all dependencies for all possible Gekko plugins: If you are unable to install one of them correctly (like #288) afraid what went wrong where and what the implications are. I also want Gekko to have windows support (without requiring Visual Studio), but plugins like redis and TA-lib definitely need to be possible for everyone else.

The proper way to do this is to require plugins to be created as sub projects (on npm as `gekko-talib-trader` or on github). But that raises the bar for adding/configuring/downloading a plugin. Right now it's just a simple script in the plugins folder (that followes [this API](https://github.com/askmike/gekko/blob/0.2/docs/internals/plugins.md)).

PR #322, opened 2016-06-10
Update kraken.js

Piega:
Updated crypto_currencies and fiat_currencies


askmike:
Thanks for the commit, I actually had it working already on the develop branch. There were actually more changes needed for Gekko to get all markets working, see 780c1668.

Issue #317, opened 2016-06-09
Add Bitstamp EUR/BTC and EUR/USD markets


Piega:

askmike:
Requires update [in dependency](https://github.com/askmike/bitstamp/issues/29).

askmike:
Dependency is now updated, however the signatures of the API calls have changed (see [here](https://github.com/askmike/bitstamp/blob/master/example.js)), also:

> Please note that API v2 endpoints rounding is different, than the one used on the old endpoints.

In order to get this into gekko, we need:
1. Add market info in exchanges.js ([done](https://github.com/askmike/gekko/pull/323/commits/09656f01fad9c834f25a81deba1b00592966176b))
2. Use [bitstamp@0.3.x](https://github.com/askmike/bitstamp)
3. Update all signatures in bitstamp.js ([example](https://github.com/askmike/gekko/blob/stable/exchanges/bitstamp.js#L131)) to add market pairs.
4. Test.

PR #323, opened 2016-06-10
Update exchanges.js

Piega:
Bitstamp EUR/BTC EUR/USD


askmike:
This PR wouldn't be enough, see #317.

askmike:
the commit 006c90010c3 has everything needed for Bitstamp's Euro markets.

PR #330, opened 2016-06-17
Updated bitfinex node module to 'bitfinex-api-node'

xhad:
I haveing issue with npm bitfinex, so I changed to bitfinex-api-node according to their new API docs and everything seems to be working fine.


askmike:
Thanks! Looks like you worked on top of an older codebase, the current stable uses a [different npm module](https://www.npmjs.com/package/bitfinex-fork). As I have some problems with the new version I would very much like to see your version with using `bitfinex-api-node`. Would you mind opening a new PR against stable (there are loads of really old gekko commits in here)?

You can just replace the contents of `/exchanges/bitfinex.js` with yours and update `package.json`.

askmike:
Closed in favor of #331.

PR #331, opened 2016-06-18
updated Bitfinex node module

xhad:
There may still be issues with trading methods that I have not been able to test yet. However, Trader.getTrades is working well. 

Kind Regards, Mike! 

Is there anything you would like me to work on?

### DIFF
--- a/exchanges.js
+++ b/exchanges.js
@@ -179,10 +179,9 @@ var exchanges = [
       }
     ],
     requires: ['key', 'secret'],
-    // TODO: should be possible to enable this for Bitfinex?
-    providesHistory: false,
+    maxHistoryFetch: 10000,
     tid: 'tid'
-    // fetchTimespan: 60
+
   },
   {
     name: 'meXBT',
@@ -308,7 +307,7 @@ var exchanges = [
   // ,
   // ---- Keeping this here for historical purposes. ----
   // {
-  // 
+  //
   //   name: 'MtGox',
   //   slug: 'mtgox',
   //   direct: true,
--- a/exchanges/bitfinex.js
+++ b/exchanges/bitfinex.js
@@ -1,5 +1,5 @@
 
-var Bitfinex = require("bitfinex-fork");
+var Bitfinex = require("bitfinex-api-node");
 var util = require('../core/util.js');
 var _ = require('lodash');
 var moment = require('moment');
@@ -47,7 +47,7 @@ Trader.prototype.retry = function(method, args) {
 }
 
 Trader.prototype.getPortfolio = function(callback) {
-  this.bitfinex.wallet_balances(function (err, data, body) {
+  this.bitfinex.rest.wallet_balances(function (err, data, body) {
     var portfolio = _(data).filter(function(data) {
       return data.type === 'exchange';
     }).map(function (asset) {
@@ -62,7 +62,7 @@ Trader.prototype.getPortfolio = function(callback) {
 }
 
 Trader.prototype.getTicker = function(callback) {
-  this.bitfinex.ticker(defaultAsset, function (err, data, body) {
+  this.bitfinex.rest.ticker(defaultAsset, function (err, data, body) {
     callback(err, { bid: +data.bid, ask: +data.ask })
   });
 }
@@ -80,8 +80,8 @@ function submit_order(bfx, type, amount, price, callback) {
   amount = Math.floor(amount*100000000)/100000000;
 
   bfx.new_order(defaultAsset, amount + '', price + '', exchangeName,
-    type, 
-    'exchange limit', 
+    type,
+    'exchange limit',
     function (err, data, body) {
       if (err)
         return log.error('unable to ' + type, err, body);
@@ -91,22 +91,22 @@ function submit_order(bfx, type, amount, price, callback) {
 }
 
 Trader.prototype.buy = function(amount, price, callback) {
-  submit_order(this.bitfinex, 'buy', amount, price, callback);
+  submit_order(this.bitfinex.rest, 'buy', amount, price, callback);
 
 }
 
 Trader.prototype.sell = function(amount, price, callback) {
-  submit_order(this.bitfinex, 'sell', amount, price, callback);
+  submit_order(this.bitfinex.rest, 'sell', amount, price, callback);
 }
 
 Trader.prototype.checkOrder = function(order_id, callback) {
-  this.bitfinex.order_status(order_id, function (err, data, body) {
+  this.bitfinex.rest.order_status(order_id, function (err, data, body) {
     callback(err, !data.is_live);
   });
 }
 
 Trader.prototype.cancelOrder = function(order_id, callback) {
-  this.bitfinex.cancel_order(order_id, function (err, data, body) {
+  this.bitfinex.rest.cancel_order(order_id, function (err, data, body) {
       if (err || !data || !data.is_cancelled)
         log.error('unable to cancel order', order_id, '(', err, data, ')');
   });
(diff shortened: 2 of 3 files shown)


askmike:
Awesome thanks. I will test the live trading soon (in another country atm). For me the hard part was first making sure the funds are in the "right" wallet on bitfinex itself, and then select only funds in this wallet (the only working wallet for me was the "exchange" wallet).

As for bitfinex todo's:
- Testing the trader (I can also do this, I have some funds there for testing)
- Asking for more historical data on first fetch, by upping the "limit" in the trades call. (example at bitstamp: [here](https://github.com/askmike/gekko/blob/stable/exchanges.js#L86) and [here](https://github.com/askmike/gekko/blob/stable/exchanges/bitstamp.js#L147-L150)).

And at some point we could start thinking about other trading types (margin trading, etc.), would be handy if by then we had the API calls ready.

---

> There may still be issues with trading methods that I have not been able to test yet. 

specific to bitfinex? Should be completely separated.

xhad:
I will try to clean this up.. such that it can be set in excahnges.js as:
`providesHistory: true,
    maxHistoryFetch: 'day',
    fetchTimespan: 60,`

askmike:
Looks great, thanks! Note that the exchanges.js file is very outdated and both `providesHistory` and `fetchTimespan` are currently not used.
- I want to work on adding all functionality for `providesHistory`, but this flag means: Gekko can request all data since a specific date (if you configure it in a way that Gekko needs data from the last month, all that data can be fetched). Gekko is currently unable to handle that properly anyway.
- `fetchTimespan` had information about the time between fetch calls in a time where some exchanges guaranteed this (mostly Mt. Gox), this is redundant now. You can ignore this property for now.

xhad:
I tested the bitfinex live trading and Gekko gave me an error:

Bitfinex requires "(api secret)" to be set in the config. However, I have
them both set in the config.json and the exchange.js.

Have you had this issue before? Is it a Gekko issue or Bitfinex issue?

askmike:
The idea is that everything specific to any exchange lives in both the `/exchanges` directory (glue code to call the API) and the `exchanges.js` file (settings / limitations Gekko has to take into account). Users that just want to use Gekko never touch any of those files, they can do everything they want from the config.

> Bitfinex requires "(api secret)" to be set in the config.

That is an error from Gekko, it shows that the settings in `exchange.js` tell Gekko that for real trading Bitfinix requires both an API key and secret (The `requires` prop in that file), other exchanges sometimes need more or less (so their `requires` property [lists more](https://github.com/askmike/gekko/blob/stable/exchanges.js#L84)). But Gekko cannot find the key/secret in the config file (Normal users never touch the `exchange.js`, only people working on exchange integrations.).

Could you double check if you configured those keys exactly [here](https://github.com/askmike/gekko/blob/stable/sample-config.js#L149-L150) in the config?

---

I can also test / fix trading integration after pulling in your PR if you don't have time :)

xhad:
ok cool. I made a new API key and secret but there is a delay before it gets activated... so i went for a walk by the river, came back and Bud Fox is working hard at the BFX desk. 

BTW, I made two trades today (manually) from the advice Gekko gave me, and both trades made money on the short side. 

Getting back to the historical data from BFX.. from the rest api there is a limit of 1 day's worth of trades. So if we are going to get full trade history.. it might be necessary to scrape data off the Websocket market events. and save that data on a server.. then serve it to Gekko from that server (monetization?)

Or, it's possible to get daily close data, but I haven't been able to find historical market_events / trade data. 

Does Gekko want only 'trade' data? Thoughts?

askmike:
> BTW, I made two trades today (manually) from the advice Gekko gave me, and both trades made money on the short side.

Awesome to hear! So this is also a feature I am looking to implement: now Gekko is only able to place limit orders, which means Gekko cannot short (most exchanges do not support shorting, but still).

> Getting back to the historical data from BFX.. from the rest api there is a limit of 1 day's worth of trades. 

This sounds like a good start, this is how bitstamp is currently implemented as well. This is fine if you want to trade or paper trade. And you can only backtest after running Gekko for a few weeks (Gekko saves all data, after a few weeks you have some data to test against).

> So if we are going to get full trade history.. it might be necessary to scrape data off the Websocket market events. and save that data on a server.. then serve it to Gekko from that server (monetization?)

I am currently running a bunch of Gekkos on my server (to store data, but also to check stability etc), this is an easy way to store trade data. I will probably offer this data for free (servers are dirt cheap) somewhere in the future. But since right now nobody is doing this (AFAIK) you can start offering data for money.

> Does Gekko want only 'trade' data? Thoughts?

Currently Gekko itself only deals with 1 minute "candles" which hold information about trades that happened on an exchange during 1 minute. If you have 1 minute candles you can pipe them into Gekko (no documentation, but this happens [here](https://github.com/askmike/gekko/blob/stable/gekko.js#L174-L180)). Backtesting is simply [piping all locally stored 1 minute candles](https://github.com/askmike/gekko/blob/stable/core/backtestMarket.js) into Gekko.

So a super easy way to offer more historical data would simply be to run a Gekko and publish the local database files online. People can just place them into their history folder and everything works.

Because all exchanges offer past trades Gekko uses [an internal module called BudFox](https://github.com/askmike/gekko/blob/stable/docs/internals/budfox.md) that retrieves past trades and transforms it into 1 minute candles. If you want Gekko to respond to a websocket instead, just crate a small module that streams 1 minute candles and pipe them into Gekko.

xhad:
Thank you very much for taking the time to help me understand Gekko. I have Bitfinex running now and it made a trade fine from the exchange account. My first Gekko Trade. :)

How should the Margin/Exchange config happen? Should you set the type in config.js? 

```
type: (margin or exchange)
```

and then handle type in exchanges/bitfinex.js?

Again, thank you so much for your time. This is really exciting to learn.

askmike:
Yes. And `exchange.js` will have an array per exchange that says what types are supported on that exchange.

---

As that is a big change, let's keep all discussion about it in this issue: #335. As it's not directly related to this PR.

---

What is the status with this PR?

xhad:
This PR is closed. The bfx module is working well.

askmike:
As in, the one currently in Gekko? I have to pull the PR in if I need your commits (but need to check them out first since you want to merge them into the `stable` branch).

xhad:
There is still more that needs to be done here.. but I would like to get your advice. There are two api endpoints: 

https://api.bitfinex.com/v1/trades/btcusd?limit_trades=10000 
&
https://api.bitfinex.com/v1/trades/btcusd?timestamp=1464739200 

And in exchanges.js:

``
Trader.prototype.getTrades = function(since, callback, descending) {
  var self = this;

  // Bitfinex API module does not support start date, but Bitfinex API does. 
  var start = since ? since.unix() : null;
  this.bitfinex.trades(defaultAsset, start, function (err, data) {
...
``

Where is getTrades() being called? I don't know where 'since' is coming from...

askmike:
Right now Bitfinex is down so I am unable to test anything, but this call:

https://api.bitfinex.com/v1/trades/usdbtc?limit_trades=10000 

Seems to be super right now. The since is going to be either:
1. `undefined` for when Gekko just needs a limited batch of trades.
2. [this value](https://github.com/askmike/gekko/blob/stable/exchanges.js#L67) (but for bitfinex, not the one for btc-e). So if you set the `maxHistoryFetch` on bitfinex to `?limit_trades=10000` this will be the since. You can also set it to `10000` if that's easier for your code (if you need to pass limit_trades in an object or so).

xhad:
Thanks for your patience. This is my first PR.

askmike:
it looks great now (just note, that raw API link you posted has the `usdbtc` in the wrong order), I will check and test it tonight (I am at work now) and will merge it in if everything works.

xhad:
It was a pretty wild day for me working on the Bitfinex module. 

Gekko was able to cancel the trade that it had sent to Bitfinex while the market was live. It then re-entered long at a lower price when that market opened. I'm quite impressed with how Gekko handled this crazy market situation at Bitfinex.

askmike:
Thanks for the pull request, I was rather busy so I was unable to test the real trader. But I had to change your code a little bit (see 5e19e4a1f): Gekko sometimes needs loads of trades (your 10000) but other times just a limited batch, at which Gekko will pass a since of `null`. In this case Bitfinex apparently did not return any trades. That is fixed now. I will test the trader later!

PR #333, opened 2016-06-20
updated cexio exchange

valeriominetti:
cexio exchange is not limited to ghs/btc pair anymore and underlying cexio library can handle more asset/currency pairs. this patch should enable gekko to use them.

### DIFF
--- a/exchanges.js
+++ b/exchanges.js
@@ -92,12 +92,19 @@ var exchanges = [
     slug: 'cexio',
     direct: false,
     infinityOrder: false,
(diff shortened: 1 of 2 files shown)


askmike:
looks great, thanks! Did you by any change test/use the real trader for cexio (using gekko)?

valeriominetti:
not yet, sorry. i just ran gekko to get advices. 
exchanges.js still need some work, in fact minimum quantities should be updated aswell.

PR #340, opened 2016-06-23
added talib module to deps

xhad:
I'm getting an error that talib is not found when turning on tradingAdvisor


xhad:
When I changed node to v4.. I got an error that node modules talib and sqlite3 were not installed...

PR #349, opened 2016-06-26
fix api

Niller2005:


askmike:
Darn, we got everything working in another branch. But merging ended up with problems, nice catch! If you have more problems let me know 👍

PR #350, opened 2016-06-26
webserver was on by default

Niller2005:
Was missing dep, so crashed. And it's been off by default since the begining i think :)


askmike:
yes, this should be off by default. Nice catch!

PR #351, opened 2016-06-27
Added nodejs-websocket as a dependency and new advice field datetime for webserver plugin

jhnferraris:


core/baseTradingMethod.js
```diff
@@ -196,15 +196,20 @@ Base.prototype.addIndicator = function(name, type, parameters) {
...
+  if(newPosition) {
+    advice = newPosition;
+  }
 
   this.emit('advice', {
-    recommandation: newPosition,
-    portfolio: 1
+    recommandation: advice,
+    portfolio: 1,
+    datetime: moment().format('YYYY-MM-DD HH:mm:ss')
```
askmike, line 213: Would you mind broadcasting the moment object itself instead? This makes it easier for other consumers to use. Actually it would probably be even better to use the time of the candle that "caused" the advice. (I can also do this after pulling this in).

package.json
```diff
@@ -25,6 +25,7 @@
     "lakebtc_nodejs": "0.1.x",
     "lodash": "2.x",
     "moment": "2.4.x",
+    "nodejs-websocket": "1.6.0",
```
askmike, line 28: So the webserver is a Gekko "plugin". Plugins dependencies are optional (you only need them if you actually use them. I see you already put them in `plugins.js`, so that means it can be removed here).

askmike:
Looks great, would you mind looking at the comments? Also please create a pull request against the `develop` branch (instead of stable)?

askmike:
This is merged in #355.

PR #354, opened 2016-06-29
Develop

wupeng1211:


askmike:
I need to do a little bit more testing, you can already play around with it by downloading the develop branch:

```
git clone -b develop https://github.com/askmike/gekko.git
```

PR #355, opened 2016-06-29
Added moment object in `advice` data - Used the correct branch

jhnferraris:


askmike:
Thanks! So I did use the time of the candle instead of the system time (b0e7433b2), this so that the advice can also be used in backtests.

PR #370, opened 2016-07-07
okcoin exchange

xhad:
Okcoin is working well here, for me. Still need to add child-accounts. Please take a look

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "lakebtc_nodejs": "0.1.x",
     "lodash": "2.x",
     "moment": "2.4.x",
+    "okcoin-china": "0.x.x",
     "poloniex.js": "0.0.6",
     "semver": "2.2.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "poloniex.js": "0.0.6",
     "semver": "2.2.1",
     "sqlite3": "3.1.3",
+    "websocket": "latest",
```
askmike, line 32: unfortunately this does not work on windows. I'll try to find an alternative which is more cross platform.

askmike:
thanks a lot! Will give this some proper testing.

LobodaCrypto:
Hello, askmike
Could you add usd/btc, usd/eth, btc/eth, btc/neo, trading pair in Okcoin exchange?

LobodaCrypto:
can I use gekko ui for okcoin exchange?

PR #405, opened 2016-08-06
trades, mytrades, orderbookupdates

0b01:


askmike:
sorry for the late response, I am currently traveling and my laptop broke down. Will look at it as soon as possible!

PR #450, opened 2016-09-18
Added percent of amount for buy or sell core/portfolioManager

Pascal66:


askmike:
I wanted to merge this but it seems you are missing some stuff: 

You are passing in percent to portfolioManager trade:

```
Manager.prototype.trade = function(what, percent) {
```

However you never pass this value in anywhere. If you fix this, I'll merge this in :)

Pascal66:
@askmike  I pass the value in the closed #448 in BaseTradingMethod

Pascal66:
I used the value in the other closed #449 in Trader.prototype.processAdvice.
I've made 3 patches instead of just one sorry

Lorkshem:
I could not figure out how to enforce this percentage limit.
Could someone explain better?

smurfy:
@askmike I like this feature. Everything is there, just in 3 pull requests. Also now with conflicts.
Should i create a combined PR?

7flash:
I think it is very important feature for the most strategies. Really I have no idea how to create a successful strategy without opportunity to set amount of lot in advice method.

askmike:
@7flash in percentage or fixed amount?

7flash:
I think the best option is to know available portfolio size in strategy and set concrete amount for every advice, for example based on calculated percentage.

7flash:
Can you explain why now amount always depends on ticker.ask ?

askmike:
> I think the best option is to know available portfolio size in strategy

Within Gekko the strategy is completely decoupled from the portfolio, Gekko deals with TA algorithms and not with execution algorithms. Whether it is the best or not, it is not the goal for Gekko.

What do you mean? That only happens when calculating the order to send to the exchange.

7flash:
Why calculates amount based on ticker.ask before send to the exchange?

askmike:
I don't understand your question. The exchanges require an amount to be send, we need to calculate what the amount will be.

7flash:
Of course. I mean I don't understand why amount size relates on "ticker.ask" value?

7flash:
An effective strategy need an opportunity to choose a lot size based on it's confidence in success, or any inner logic of money management.

askmike:
> An effective strategy need an opportunity to choose a lot size based on it's confidence in success, or any inner logic of money management.

This is outside the scope of Gekko, once you need something with more advanced order execution Gekko is not for you. If you have any other questions, please create a new issue, as I have no idea on what code you are commenting (this PR, stable brach, dev branch, trader branch)? Also all the other people in this thread will get emails.

huemulnasillo:
Hi everyone! Is this feature working already?

askmike:
Its not working at the moment, and with the current statistics calculations
(roundtrips and risk metrics) not feasible.

askmike:
For that reason I am closing this issue for now. The concept might be revisited later.

LooOOooM:
That is realy poor that this ticket just got closed and gekko still has no possibillity to deal with portions, as this is soooo important to riskmanagement. Especially for Stoploss if the stoploss is too far away from the entrypoint, letz say 10%. I do not want to risk 10% but i.e jsut 3% of my total portfolio. Therefore it is absolutly needed to set the amount of the trade just much less that the 10% stoploss differnece to entrypoint is not equal to the total risk of my portfolio
please reopen this issue

rcoenen:
Can this be configured anywhere in the WebUI?

PR #454, opened 2016-09-19
Added basic postgresql support

tinyrobot:


askmike:
Could you create a PR for this into askmike:develop?

askmike:
Great start, I'll try to refactor this into a seperate adapter (so we can keep dependencies seperated, eg. I am pretty sure this won't build on windows, but for those Gekko could use sqlite..)

Sorry for not providing documentation on how to do this!

askmike:
> should I have done that differently?

No that's great for now! Closing this in favor of #455.

PR #461, opened 2016-09-24
fix trade and profit report

niklasbuschmann:
Changes the look of the profit simulator

Also fixes (PROFIT REPORT) ... months _days_ bug
Logs currency/asset _before_ zeroing it

See this screenshot (`| grep 'Profit simulator'`)
screen shot 2016-09-24 at 11 26 18


askmike:
ty!

PR #462, opened 2016-09-24
Add PPO_RSI strategy

niklasbuschmann:
Adds a mixed PPO/RSI strategy which adjusts the RSI thresholds according to the PPO trend.

Had a 4.8 ROI on the BTC/ETH pair on Poloniex

screenshot


askmike:
awesome :)

Issue #457, opened 2016-09-20
Stranges tradings advices


Pascal66:
Trader say **received advice to short selling**
AND result error is : **wanted to buy ?!**

2016-09-20 11:45:19 (INFO):     Trader Received advice to go short Selling  ETH

2016-09-20 11:45:19 (INFO):     We have new trading advice!
2016-09-20 11:45:19 (INFO):              Position: short
2016-09-20 11:45:19 (INFO):              Market price: 13.77
2016-09-20 11:45:19 (INFO):              Based on market time: 2016-09-20 09:43:00

2016-09-20 11:45:20 (INFO):     wanted to buy but insufficient ETH (1e-8) at BTC-E

askmike:
What market is this?

Pascal66:
USD ETH

PR #464, opened 2016-09-28
Sell Buy correction

Pascal66:
Correcting Issue #457


askmike:
sorry for the late response, but looks great!

PR #468, opened 2016-10-03
Generalize PPO_RSI to work with other momentum indicators

niklasbuschmann:
Allows PPO to work together with any of the momentum oscillators added in #465


askmike:
:+1:

niklasbuschmann:
It would make sense to also merge #465 as it contains the indicators. Or is there any problem?

PR #493, opened 2016-11-04
Add mongodb support

S7evinK:
Tested with mongodb >= 3.3

### DIFF
--- a/plugins/mongodb/handle.js
+++ b/plugins/mongodb/handle.js
@@ -1,0 +1,46 @@
+var mongojs = require('mongojs');
+var mongoUtil = require('./util');
+
+var util = require('../../core/util.js');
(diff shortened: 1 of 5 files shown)


askmike:
great stuff! Thanks.

I am working on a UI for gekko (see branch web-wrapper), and that requires DB readers to implement a method called `tableExists` which will pass a boolean into a callback (whether the specified table exists or not). [See here](https://github.com/askmike/gekko/commit/e3a5c71239561956d0f806b164b9ede3e28133b0#diff-62f8ba0b7b3b83bd10470a0cdd70f21c).

I can also look at adding this whenever I have some time.

S7evinK:
Added tableExists in d9e4a108ef43fb2 and 41648fe62421

PR #504, opened 2016-11-19
Update UIconfig.js

0b01:
When this launched under an IP address as opposed to localhost. The host name should be updated.


askmike:
this breaks the [development flow](https://github.com/askmike/gekko/blob/develop/docs/internals/gekko_ui/frontend.md#developing-for-the-gekko-ui-frontend) where the location of the API is different from where the frontend is served from.

The idea is that if you need to host it somewhere else you update the UIconfig!

PR #508, opened 2016-11-22
Corrected loaders and typo

mattbrun:
New webpack requires loaders to be defined with their full name.
Thus I moved the `json` loader require to the webpack config as well.
There was also a vue component require which had a name typo.

I don't know how it could happen, but the first commit author is wrong because it's not me...
I've corrected the commit author with the second one.
Anyway, I did both commits ;)

Cheers,
M.


askmike:
Awesome stuff, thanks!

PR #510, opened 2016-11-27
Update server.js

luisan00:
rename 'routes/liveGekkos' to 'routes/livegekkos' in server.js or rename the file livegekkos.js to liveGekkos.js in the path /web/routes/


askmike:
thanks, the live gekko is a work in progress. Don't expect to much right now!

luisan00:
The live gekko is in early stage, but in general gekko is a good job.

PR #513, opened 2016-12-01
Ui/market watchers [WIP]

askmike:
this PR contains functionality for running gekkos using gekko-UI that will watch a market in realtime (using [the budfox realtime market](https://github.com/askmike/gekko/blob/stable/docs/internals/budfox.md)) and are able to run [strats](https://github.com/askmike/gekko/blob/stable/docs/Trading_methods.md) against this market.

----

When using Gekko normally (not using the UI but the commandline), it is quite common to just run one strat over one market. This is very simple using [a single pipeline](https://github.com/askmike/gekko/blob/stable/docs/internals/architecture.md#gekkos-architecture): configure a market and pipe the data into the strat.

However the UI makes it easy to run a lot of these parts in parallel, so a new architecture called "market watching" is necessary.

> screen shot 2016-12-01 at 15 33 41

When you configure the UI to run a strat over a market the UI will actually create two gekko instances. One that simple implements the budfox markets and dumps all market data into the database. And one "leech" gekko that will pull the data out of the database and run the strat over it.

Advantages:

- Very simple to run multiple strats over 1 market (gekko only needs to watch the market once).
- Very simple to run the same strat over 2 markets.
- Easier to gather market data (by not running a strat but simple a "market watcher").

this is currently still a WIP.


askmike:
Watchers implemented, it currently looks like this: https://wizb.it/gekko/static/watcher1.gif

Note that I will implement a price chart so you can "see" what the market looks like.

askmike:
And here is the market chart:

screen shot 2016-12-04 at 14 39 25

I would like to at some point use a candle chart, however I want to focus on shipping the main func first.

askmike:
closed in favor of #613

PR #519, opened 2016-12-10
Fix talib bbands calculation (returned undefined result)

osnwt:
The same for develop branch.
Probably there are other similar places, did not check.


askmike:
👍

PR #520, opened 2016-12-10
Add a telegram bot similar to xmpp and irc ones

osnwt:
It requires to register a bot using Telegram's BotFather first, then put the token into configuration.
gekko_telegrambot


askmike:
this is awesome!

PR #532, opened 2016-12-28
Update package.json

BarnumD:
Add talib to packages as required to talib indicators.


askmike:
Talib is very hard to compile on windows, you only need talib for methods that use talib indicators. I rather keep it out of `package.json` because else Gekko cannot be installed by npm on windows (without proper compile toolchain). If you right now try to run Gekko with a talib method, it will prompt you to install talib manually.

ghost:
My thoughts exactly

PR #545, opened 2017-01-13
Incorrect string format

aclymer:
The string was formed using the wrong apostrophe characters and threw and error using Node.js tools for Visual Studio

### DIFF
--- a/core/util.js
+++ b/core/util.js
@@ -70,8 +70,8 @@ var util = {
     }
   },
   logVersion: function() {
-    return  `Gekko version: v${util.getVersion()}`
-    + `\nNodejs version: ${process.version}`;
+    return  'Gekko version: v${util.getVersion()}'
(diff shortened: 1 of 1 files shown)


S7evinK (commented):
I think it's just fine to use back-ticks ([Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)) here.
With this change you'll output `Gekko version: v${util.getVersion()}` instead of something like `Gekko version: v0.3.4`

aclymer:
The compiler and/or tools I was using threw an error. Simple fix to a tiny typo.

askmike:
@S7evinK is right, changing this would change the output so that the version will not be printed. Please use tools that support [template literals](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals).

PR #552, opened 2017-01-17
Comment Clarification and Trader Sell fix

aclymer:
Clarified Method Comments
Fixed issue with Sell
 - amount < available causing insufficient funds message (should be >)


askmike:
Great stuff, but could you please PR it against develop branch? Thanks!

Issue #548, opened 2017-01-16
Backtesting not working when postgresql backend enabled


p4u:
Method Reader.prototype.tableExists does not exist in plugins/postgresql/reader.js 

So if backend postgressql is enabled the backtesting does not work

PR #553, opened 2017-01-18
tableExists fix for postgresql and mongodb

S7evinK:
fix for #548
+ small change for sqlite reader (use parameter instead of hardcoded "candles")


askmike:
👍

PR #562, opened 2017-01-22
Revert "Bitstamp added Ripple (XRP) last week."

SkaTeMasTer:
Reverts askmike/gekko#558


askmike:
I haven't been paying attention to Bitstamp, but did they remove these markets again? Or was something wrong?

PR #564, opened 2017-01-24
Added GDAX Support

smurfy:
This PR adds GDAX Support.

I do not 100% like the "import" feature, but it it works. It scans back the history until it finds the start date and then fetches every trade after this point.

This could be slow, mainly because gdax only supports fetching 100 trades at once. 
It works OK on ETH-BTC but for example BTC-EUR  takes quite a while.

Trades have been tested on the sandbox.

Not sure where to document the configs but here they are:

    config.trader = {
       enabled: true,
       key: xxxx,
       secret: xxxx,
       passphrase: xxxx,
       sandbox: false,  //defaults to false
       post_only: true //defaults to false, if true all trades will be POST Only which means all orders will be maker trades or will be rejected by the system
    }

Also the import date should be reduced for importer runtime:

    config.importer = {
         daterange: {
             from: "2016-07-01 12:00:00"
         }
    }

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Gekko is not built for HFT or anything related to being the fastest (like arbitr
 | Exchange        | Monitoring | [Trading](https://github.com/askmike/gekko/blob/stable/docs/Plugins.md#trader) | [Importing](https://github.com/askmike/gekko/blob/stable/docs/Importing.md) | Notes |
 | --------------- |:----------:|:-------:|:---------:|-------|
 | [Poloniex](https://poloniex.com/)      | ✓ | ✓ | ✓ | |
+| [GDAX](https://gdax.com/)      | ✓ | ✓ | ✓ | |
 | [BTCC](https://btcc.com/)      | ✓ | ✓ | ✓ | (=BTCChina) |
 | [Bitstamp](https://bitstamp.com/)      | ✓ | ✓ | ✗ | |
 | [Kraken](https://kraken.com/)      | ✓ | ✓ | ✗ | |
@@ -126,7 +127,6 @@ If you installed Gekko via git you can easily fetch the latest updates by runnin
 
 *Better exchange support:*
 
-- add GDAX exchange (supports [importing](https://docs.gdax.com/#get-historic-rates))
 - support importing at bitfinex ([here](http://docs.bitfinex.com/#trades)).
 - add okcoin China
 - add okcoin
--- a/exchanges.js
+++ b/exchanges.js
@@ -634,6 +634,27 @@ var exchanges = [
     providesHistory: false,
     tid: 'tid'
   },
+  {
+    name: 'GDAX',
+    slug: 'gdax',
+    direct: false,
+    infinityOrder: false,
+    currencies: ['USD', 'EUR', 'GBP', 'BTC'],
+    assets: ['BTC', 'LTC', 'ETH'],
+    markets: [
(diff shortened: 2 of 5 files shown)


askmike:
This is awesome stuff! Has been on my list for way to long!

👍

askmike:
Would you mind creating a PR against develop branch?

smurfy:
Will do, i probably will squash the commit too.

smurfy:
@askmike Not sure if i should create a ticket for this, or not. but i just noticed that with the post only option it is possible that the order will be rejected for a long time. Since gekko has no limit in how long it will retry to post the order this could be quite stupid, because the price will change and it makes no sense to do the order after a while.  So i think gekko should have a limit on how long it cancels/retries to post an order.

I also noticed (probably another ticket) the exchange backend class is "stateless" in normal mode and "state full" in importer mode. Aka a new instance is created every time its called. I wanted to have some "persistant" properties but this is not possible atm.

askmike:
>  but i just noticed that with the post only option it is possible that the order will be rejected for a long time. Since gekko has no limit in how long it will retry to post the order this could be quite stupid, because the price will change and it makes no sense to do the order after a while. So i think gekko should have a limit on how long it cancels/retries to post an order.

The idea is that once an order is not executed (fully), it will recreate an order after rechecking the current market (by [updating ticker information](https://github.com/askmike/gekko/blob/develop/plugins/trader/portfolioManager.js#L165-L168)). So if the price is crashing and bot is trying to sell, it will constantly try to sell for a lower price.

I do think it's important that the [portfolioManager](https://github.com/askmike/gekko/blob/develop/plugins/trader/portfolioManager.js) tries to take a certain position (as adviced by the strat) no matter what.

Is this not working properly for you?

----

> I also noticed (probably another ticket) the exchange backend class is "stateless" in normal mode and "state full" in importer mode. Aka a new instance is created every time its called. I wanted to have some "persistant" properties but this is not possible atm.

Import mode is a basic market within Gekko:

Gekko architecture

The actual market (node stream object) is [a wrapper](https://github.com/askmike/gekko/blob/develop/core/markets/importer.js) around your own [gdax importer file](https://github.com/smurfy/gekko/blob/42fefc5067230a047e70c886bc0c3390e021fe9d/importers/exchanges/gdax.js)

You can store state in your own importer file ([like so](https://github.com/smurfy/gekko/blob/42fefc5067230a047e70c886bc0c3390e021fe9d/importers/exchanges/gdax.js#L15-L17)) or in the exchange base class (https://github.com/smurfy/gekko/blob/gdax/exchanges/gdax.js), this last one only gets created in the beginning ([here](https://github.com/smurfy/gekko/blob/42fefc5067230a047e70c886bc0c3390e021fe9d/importers/exchanges/gdax.js#L19)).

If you have any questions feel free to ask!

smurfy:
Ok, squashed the commits and changed them against develop.
I kept different stages in separate branches. 

```gdax_stable``` stable branch, squashed commits
```gdax_stable_old``` stable branch, separate commits

smurfy:
Fee differs based on trading unit/asset combination. Maybe its possible adding the base fees to ```exchanges.js```  ?

ghost:
when a ORDER is REJECTED on GDAX after an advice gekko stop making order https://github.com/askmike/gekko/issues/2067
Does it means we cannot use post_only on gdax ?

PR #556, opened 2017-01-20
HOTFIX fixed readme docker image

ivan1993spb:
- [Dockerfile](https://github.com/ivan1993spb/docker-gekko/blob/master/Dockerfile)
- [Repo](https://github.com/ivan1993spb/docker-gekko)
- [Image](https://hub.docker.com/r/ivan1993spb/gekko/)

PR #565, opened 2017-01-25
Add ability to run gekko in a docker container using docker-compose.

pygeek:
Rationale: Adding docker-compose makes it easier for contributers and
end-users to interact with their docker containers. For example,
setting up and running gekko using docker-compose could be as easy as
`docker-compose up gekko`. Simply starting, stopping or restarting it as
a daemon could be `docker-compose start`, `docker-compose stop`, or
`docker-compose restart` respectively.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,1 @@
+FROM node:7.4-onbuild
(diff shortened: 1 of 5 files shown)


askmike:
See #556

askmike:
Can you create a new PR without the extra commits?

Issue #578, opened 2017-02-01
period is not defined


danielsalles:
EDIT: #580 --- bug fixed
I'm trying to create my custom methods, and I'm having problems.
https://gist.github.com/danielsalles/ccd65832cd333a34906087f87ae05acf

It seems the Talib is charging me something that does not exist in context. I do not get it right, I'm still not familiar with gekko.

```/home/dbastos/botbitcoins/gekko/node_modules/sqlite3/lib/trace.js:28
                    throw err;
                    ^

ReferenceError: period is not defined
    at Object.run (/home/dbastos/botbitcoins/gekko/core/talib.js:343:30)
    at _.each.indicator (/home/dbastos/botbitcoins/gekko/core/baseTradingMethod.js:181:30)
    at forOwn (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:2106:15)
    at Function.forEach (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:3303:9)
    at Base.tick (/home/dbastos/botbitcoins/gekko/core/baseTradingMethod.js:179:7)
    at Base.bound [as tick] (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Actor.processCustomCandle (/home/dbastos/botbitcoins/gekko/plugins/tradingAdvisor.js:68:15)
    at CandleBatcher.bound (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at emitOne (events.js:96:13)
    at CandleBatcher.emit (events.js:188:7)
--> in Database#all('\n    SELECT * from candles_BTC_DASH\n    WHERE start <= 1475328239 AND start >= 1475325240\n    ORDER BY start ASC\n  ', [Function])
    at Reader.get (/home/dbastos/botbitcoins/gekko/plugins/sqlite/reader.js:78:11)
    at Reader.bound [as get] (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Market.get (/home/dbastos/botbitcoins/gekko/core/markets/backtest.js:56:15)
    at Market.bound [as get] (/home/dbastos/botbitcoins/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Market._read (/home/dbastos/botbitcoins/gekko/core/markets/backtest.js:47:8)
[...]

PR #580, opened 2017-02-03
Mini-bug talib-ema fixed

danielsalles:
Ref: https://github.com/askmike/gekko/issues/578


askmike:
Could you please submit a PR to merge this into develop? Thanks!

Issue #538, opened 2017-01-03
Says buy successful on bitstamp but no trade or order made


joelbrennan0:
Hi,

Everything was fine until today and made several trades on BitStamp without any issue.  However something seems to have gone wrong today..

Here is the terminal output:

2017-01-03 14:09:03 (INFO):     (PROFIT REPORT) original simulated balance:      71.01750 USD
2017-01-03 14:09:03 (INFO):     (PROFIT REPORT) current simulated balance:       71.32758 USD
2017-01-03 14:09:03 (INFO):     (PROFIT REPORT) simulated profit:                0.31008 USD (0.43663%)
2017-01-03 14:09:03 (INFO):     Attempting to BUY 0.06852757002924836 BTC at Bitstamp
2017-01-03 14:10:04 (INFO):     BUY was successfull

But nothing happened on BitStamp ... No order or anything.

Trader is enabled, and to my knowledge I haven't changed anything apart from DEMA / MACD config settings.

Can anyone offer any advice please?

askmike:
you enabled a trade simulator. Enable the real trader!

askmike:
Is this closed because it did buy? Or?

joelbrennan0:
Hi,

Thanks for getting back Mike, sorry for being a pain, its more than likely I'm doing something wrong here and getting into a panic.

I have deleted everything and cloned anew to be sure..

The old version was selling but not buying on advice.

I can't work out if the new version is buying or not.

I'm still having weirdness happening, sells fine but says it's attempting to / buy successful  but never see any evidence of the buy order on bitstamp:
cant figure this out.. ( please see attached output:)
[gekko1mCandle.txt](https://github.com/askmike/gekko/files/684911/gekko1mCandle.txt

After this run I'm still in USD, is that correct?

Sorry I should try and get this on my own but haven't had time to look under the hood yet.  I'm really not familiar with node apps either.

Is there any good way of debugging you can recommend to me?  could I catch the query string data and see whats happening?  Are you able to help point out where I'm going wrong?

Please let me know if you need any further information.

Cheers...  I'll keep trying.. It's probably something silly I'm doing this end..

joelbrennan0:
Also I f**kd up and didn't back up my sqlite history so I'm starting again there :/  any body have a bitstamp file they may be able to share for the last couple of days?  At this stage I'm not too worried about the settings I just want to confirm that buy / sell orders are being executed as intended.

Thx

joelbrennan0:
[...]

PR #589, opened 2017-02-16
Workaround for #538 (silent failure when buying on Bitstamp)

jjongman:
I ran into issue #538 and after some research I found that:
1) The bitstamp.js file didn't handle errors coming from the server correctly
and
2) It failed because it was trying to buy too much. As a workaround I have decreased the amount to buy by 1% but obviously the original calculation goes wrong somewhere.


askmike:
Could you please PR this against develop? I am not sure if Gekko is compatible with the new version of the bitstamp npm module.

jjongman:
Done, but please consider addressing this in the main branch as well as Gekko is currently unusable on Bitstamp (for me anyway).

PR #592, opened 2017-02-17
Please add a stop loss

ghost:
Just need to add stop loss like if i'm getting 5% and i want stop(Close/Sell my order) if they take 1.5% of my profit. And also when open order why isnot implemented yet? i think it's possible.


Momohime:
Stop loss is relatively easy to implement and can be done so in your strategy. It's up to you to implement the correct calculations according to your strategy and your profit / loss margins.

ghost:
I'm using RSI so, how i do that?

PR #600, opened 2017-02-25
sell fix bitstamp

jjongman:
Sorry Mike, there was a mistake in my previous pr. Fixed here.


askmike:
👍  will test out as soon as I can spend some time on Gekko...

PR #613, opened 2017-03-08
Ui/market watchers rebase

askmike:
see details in #513


askmike:
Merging this now just as a prerequisite to get stable up to date, it's way too old..

PR #616, opened 2017-03-08
to fix: toFixed is not a function

Niller2005:
log.info('\t', fund.name + ':', fund.amount.toFixed(12));
                                                                     ^

TypeError: fund.amount.toFixed is not a function


askmike:
👍

PR #617, opened 2017-03-09
put parseFloat() on all toFixed()

Niller2005:
Had the same problem on line 255 as my last pull request. So I just did them all :)


askmike:
Hmm, these are error messages indicating that because an amount is less than what is available a trade could not happen. But if you need to parse it as float that means the if check that came before also didn't work properly.

Would you mind creating a PR against develop instead? I still want these changes though, will figure the if check out later.

Also:

- what exchange (and market) are you using?

Niller2005:
I'm using bitfinix

PR #611, opened 2017-03-05
[WIP] Release new stable version (0.4)

askmike:
First focus is stability (this PR), with experimental UI. Next release will be stable UI.

More info about the UI can be found in #489 and #513.

We can merge after:

CLI (legacy) interface:

- [x] make sure live market works
  - [x] make sure live trading works
  - [x] make sure paper trading works
- [x] make sure backtesting works
- [ ] verify whether all adapters work.

UI interface (new default):

- [x] merge ui-watchers into develop (even though it's WIP)
- [x] finish UI watchers:
  - [x] finish strat runner interface (show trades in price graph)
  - [x] show profit report (for paper trader)
- [x] make sure backtesting works
- [x] make sure backtesting talib works (UI, see [here](https://github.com/askmike/gekko/issues/489#issuecomment-285776325))
- [x] rewrite docs, default to UI
- [x] Add basic UI feedback
  - [x] modal when gekko backend is down.
  - [x] message when a live gekko crashes or cannot start.
- [x] force sqlite adapter (for now, see #625).

PR #619, opened 2017-03-09
put parseFloat() on all toFixed()

Niller2005:
#617


askmike:
👍 

I'll try to merge develop into stable soon (see #611)

askmike:
I pushed a new commit into develop that should prevent the parseFloat from happening in the first place. Could you let me know if you run into more problems?

PR #635, opened 2017-03-19
talib fix

danielsalles:
https://github.com/askmike/gekko/pull/580


askmike:
This includes fixes already merged in develop, as well as fixes to files that don't exist anymore (in develop, which will be merged into stable soon). After trying to solve merge conflicts I couldn't see any relevant changes/fixes in these commits.

If this is supposed to solve something specific? If so could you create a PR for what is is supposed to solve?

PR #644, opened 2017-03-22
Fix postresql adapter

RatkoR:
Postgresql adapter seems to be broken since commit 837d18e3e1dec54c. This commit changed `config.adapters.sqlite` to `config.sqlite` for sqlite and missed to do so for postgresql.

If running gekko on empty postgresql server, a new DB poloniex is created. Gekko then fails with `relation '"candles_usdt_btc" does not exist`. The second commit fixes it in a way it is already fixed in sqlite.

### DIFF
--- a/plugins/postgresql/handle.js
+++ b/plugins/postgresql/handle.js
@@ -8,13 +8,13 @@ var dirs = util.dirs();
 
 var log = require(util.dirs().core + 'log');
 
-var adapter = config.adapters.postgresql;
+var adapter = config.postgresql;
 
 // verify the correct dependencies are installed
 var pluginHelper = require(dirs.core + 'pluginUtil');
 var pluginMock = {
   slug: 'postgresql adapter',
-  dependencies: config.adapters.postgresql.dependencies
+  dependencies: config.postgresql.dependencies
 };
 
 var cannotLoad = pluginHelper.cannotLoad(pluginMock);
@@ -30,10 +30,10 @@ var dbName = config.watch.exchange.toLowerCase();
 
 var mode = util.gekkoMode();
 
-var connectionString = config.adapters.postgresql.connectionString+"/postgres";
+var connectionString = config.postgresql.connectionString+"/postgres";
 
 var checkClient = new pg.Client(connectionString);
-var client = new pg.Client(config.adapters.postgresql.connectionString+"/"+dbName);
+var client = new pg.Client(config.postgresql.connectionString+"/"+dbName);
 
 /* Postgres does not have 'create database if not exists' so we need to check if the db exists first.
 This requires connecting to the default postgres database first. Your postgres user will need appropriate rights. */
--- a/plugins/postgresql/handle.js
+++ b/plugins/postgresql/handle.js
@@ -62,6 +62,8 @@ checkClient.connect(function(err){
         });
       }else if(mode === 'backtest') {
         util.die(`History database does not exist for exchange ${config.watch.exchange}.`);
+      }else{
+        util.die(`Start gekko first in realtime mode to create tables. You are currently in the '${mode}' mode.`);
       }
     }else{ //database exists
       log.debug("Database exists: "+dbName);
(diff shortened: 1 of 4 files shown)


askmike:
Hey @RatkoR,

Thanks for this PR, the reason I failed to include anything for postgresql is that I am working on a new UI for gekko, this UI will require more/changed calls to database adapters. I was planning on finalizing this API first before updating adapters (besides sqlite).

Currently the following things need to be changed:

- the error catch you already included.
- gekko expects the reader to expose a tableExists method (see [here](https://github.com/askmike/gekko/pull/611/files#diff-62f8ba0b7b3b83bd10470a0cdd70f21cR79))
- gekko expects the adapter to expose a scanner that when called will return a list of tables that may hold market data (see [here](https://github.com/askmike/gekko/pull/611/files#diff-2c99872c0a9f71c7106e7e60cd3e31df))

But I think more will change:

- for example I really want to take all logic from [mostRecentWindow](https://github.com/askmike/gekko/blob/2c134598538c1ecff8632146adacd075b3479215/plugins/sqlite/reader.js#L16) outside the adapter and into code that actually uses it.
- right now the UI is not persisting anything to database (trades, backtest results, etc), this might change.


---

EDIT: if you can update this PR with `Currently the following things need to be changed` I will merge this PR, we just need a simple way to configure what adapter the UI will use.

askmike:
> Postgresql adapter seems to be broken since commit 837d18e.

Woa, I didn't notice I broke all adapters on stable! We might need to merge a hotfix into stable, as it will take some more time before we can merge develop (see #611).

RatkoR:
Ok, thx for the explanation. I'll try adding those two additional points to the PR.

---

A side question.

We have some additional PRs in work: like adding more columns to the store or changing postgresql so that all gekko tables are in one DB. How do you feel about it. For example, having one DB instead of multiple seems more manageable to us, specially when multiple users are playing with it.

askmike:
>  changing postgresql so that all gekko tables are in one DB.

This is fine, I don't have any opinions about how postgresql should persist data and one database is probably better. As long as Gekko get's the right data in the calls.

This looks like a small change and it would be best to include the next major (which has the UI). But it might be handy to add a small migration script that is able to turn the old style into the new style (for advanced users having months of history in the old format).

> adding more columns to the store

Please let me know what you had in mind. It might be good to keep this in sync across adapters, I've been getting some requests to store additional information. It would be very confusing if we expose functionality that is only available in certain adapters.

> manageable to us, specially when multiple users are playing with it.

I am very intrigued! Please tell me more :)

RatkoR:
Ok, I've added some more code, trying to get those `Currently the following things need to be changed` parts done.

Feel free to comment the code, I'll fix whatever you find strange or missing.

Do you have tests that cover `scanner` and `tableExists` functions?

RatkoR:
>  it might be handy to add a small migration script

Hmm, I don't really like migration scripts. I thought about doing a config variable: if (for example) user has `config.postgres.database = 'my_db_name;'` variable set, then let's put all tables into this database. And if user does not have that config, he'll see no changes, his gekko will work as before (multiple databases).

> It might be good to keep this in sync across adapters

Sure!

> specially when multiple users are playing with it.

Well it happens even with a single dev. :) When I play with gekko and if I decide that I need to start from scratch, it's easier to just delete one database. Also all my changes are only in my database. So if the other dev has his tables in his database I don't trash his work if I delete my database.

askmike:
> Do you have tests that cover scanner and tableExists functions?

Unfortunately not :/ Would very much like to add tests for adapters (open to PRs).

> (...) his gekko will work as before (multiple databases).

Great!

> When I play with gekko and if I decide that I need to start from scratch, it's easier to just delete one database.

Gekko soon requires more calls to the adapter to delete datasets (starting out with all data from a market). See #654 for reasoning.

askmike:
Is this PR finished? Will test when you give the signal.

RatkoR:
Yes, go ahead and try it. Keep me posted if you find anything.

askmike:
👍  will merge already (usage is undocumented at the moment), will test & document & add option to use postgre as a backend for Gekko's UI soon.

Thanks :)

PR #648, opened 2017-03-25
Remove defaults for computed values

zeroX-tj:
I had issues the asset and currency didn't update when i switched the default market. (And because i was using Kraken it complained about the default BTC value)
Removing the default values for the computed fields seem to fix this.


askmike:
Thanks! 👍

PR #652, opened 2017-03-27
Fixes logging:  Starting to watch the market: Kraken / 

zeroX-tj:
Currently logging shows this: Starting to watch the market: Kraken /
With this fix it shows asset and currency properly like this: Kraken ETH/EUR


askmike:
thanks a lot, this has been bugging me for quite a while!

PR #656, opened 2017-03-30
Mongodb adapter fixes + scanner

zeroX-tj:
This works for me. The config still needs to be fixed as well as some config values for the exchange are not there when running from the ui. So i had to add a workaround for now.

### DIFF
--- a/core/markets/backtest.js
+++ b/core/markets/backtest.js
@@ -5,7 +5,7 @@ var dirs = util.dirs();
 var log = require(dirs.core + 'log');
 var moment = require('moment');
 
-var adapter = config.adapters[config.adapter];
+var adapter = config[config.adapter];
 var Reader = require(dirs.gekko + adapter.path + '/reader');
 var daterange = config.backtest.daterange;
 
--- a/core/markets/backtest.js
+++ b/core/markets/backtest.js
@@ -5,7 +5,7 @@ var dirs = util.dirs();
 var log = require(dirs.core + 'log');
 var moment = require('moment');
 
-var adapter = config[config.adapter];
+var adapter = config.adapters[config.adapter];
 var Reader = require(dirs.gekko + adapter.path + '/reader');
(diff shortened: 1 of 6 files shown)


core/markets/backtest.js
```diff
@@ -5,7 +5,7 @@ var dirs = util.dirs();
 var log = require(dirs.core + 'log');
 var moment = require('moment');
 
-var adapter = config[config.adapter];
+var adapter = config.adapters[config.adapter];
```
askmike: The config format has actually changed, you changed it to the old format.

core/tools/dateRangeScanner.js
```diff
@@ -10,7 +10,7 @@ var config = util.getConfig();
 var dirs = util.dirs();
 var log = require(dirs.core + 'log');
 
-var adapter = config[config.adapter];
+var adapter = config.adapters[config.adapter];
```
askmike: dito

plugins/mongodb/scanner.js
```diff
@@ -0,0 +1,40 @@
...
+    this.db = handle;
+
+    let markets = [];
+    async.waterfall([
+        (cb) => {
+            handle.getCollectionNames(cb)
+        },
+        (collections, cb) => {
+            async.each(collections, (collection, cb) => {
+                let [exchange, type] = collection.split('_');
```
askmike, line 18: Do you know if node v6.0 (minimum as required by Gekko) supports this?

askmike (changes requested):

plugins/mongodb/scanner.js
zeroX-tj: Works for 6.0 (tested)

askmike:
👍  I'll ping you if there are issues with mongo at some point.

askmike:
thanks a lot :)

PR #665, opened 2017-04-04
Adds configuration key for postgres schema.

RatkoR:
A simple PR that adds a schema key for postgresql plugin. Let's have it in config since we're using it in postgresql/util.js. [see here](https://github.com/askmike/gekko/blob/develop/plugins/postgresql/util.js#L16).


askmike:
👍

PR #666, opened 2017-04-04
Postgres Single Database Patch

RatkoR:
Using this patch, user can have all tables in one postgresql DB.

There is a new `config.postgresql.database` key. By default this key is missing (or set to null), which means there will be no changes for existing users. Only if user sets it to some value, eg:

```
config.postgresql = {
  path: 'plugins/postgresql',
  ...
  database: 'gekkodb'
  ...
}
```
only then will gekko try to save all tables into `gekkodb` database.

To prevent table collision, tables in this DB setup are named with additional prefix. eg: `poloniex_candles_usdt_btc`.


askmike:
👍  I'll ping you if someone has problems!

Issue #664, opened 2017-04-04
SQLITE_BUSY: database is locked


hoerbi1000:
When im running gekko, im nearly getting always this error after about 1 -6 hours

`2017-04-04 21:34:28 (DEBUG):    In downtrend since 7 candle(s)
events.js:161
      throw er; // Unhandled 'error' event
      ^
Error: SQLITE_BUSY: database is locked`

does anyone else also know about the bug? 
i have nothing special run in background.

tneele:
I experience the same when I run a market watcher at the same time as running an import. In that case, this error seems like the logical result: two workers want to write to the database.
Because of this, I wanted to switch to MongoDB completely, but unfortunately, it is not yet supported in the UI.

askmike:
@hoerbi1000 

> When im running gekko, im nearly getting always this error after about 1 -6 hours

Are you only running a single Gekko (through the commmandlie)?

@ThomasNeele 

> I experience the same when I run a market watcher at the same time as running an import.

I will try to reproduce, but I'm not sure how much we can do about it. Are you interested in different markets on the same exchange, or simply one market?

mongodb support was updated by @zeroX-tj (I have reviewed it but untested), could you try that out on the develop branch (and let me know how that goes)?

tneele:
At that time, I was tracking as well as importing one market.

There were still some issues with MongoDB for me in the development branch as well. I have tried to fix those in a PR.

askmike:
@ThomasNeele Thanks a lot for that.

@hoerbi1000 I am able to reproduce, but it's a hard problem to fix at the moment. You can workaround this for now by not importing + running live gekkos against the same exchange..

hoerbi1000:
@askmike 
Okay thanks for the Support ;)

darklow:
@askmike I am experiencing same problem as well. Since you were able to reproduce and know why the problem appears - is it sqllite specific and would moving to postgres database help to solve this issue?
I am also running commandline trader plus I am running few market monitors to store data to analyze in backtest later.

askmike:
>  is it sqllite specific and would moving to postgres database help to solve this issue?

Yes, judging by your use case (running multiple gekkos at the same time) postgresql would be a better option.

xyzzy529:
I start a fresh Server instance from Windows CMD prompt. Then start a Live Gekko Paper Trader.
After about 20-30 minutes, get this error.
Running fresh install with no changes. Installed per Instructions on Document site.

```
[...]

PR #669, opened 2017-04-05
Fixed MongoDB handler to work in UI

tneele:
As per issue #664.

### DIFF
--- a/web/vue/src/components/gekko/singleStratrunner.vue
+++ b/web/vue/src/components/gekko/singleStratrunner.vue
@@ -84,7 +84,6 @@ import { post } from '../../tools/ajax'
 import spinner from '../global/blockSpinner.vue'
 import chart from '../backtester/result/chartWrapper.vue'
 import paperTradeSummary from '../global/paperTradeSummary.vue'
-import config from '../../../../routes/baseConfig.js'
 // global moment
 
 export default {
@@ -109,14 +108,6 @@ export default {
     },
     data: function() {
       return _.find(this.stratrunners, {id: this.$route.params.id});
-    },
-    baseCandleConfig: () => {
-      return {
-        adapter: config.adapter,
-        sqlite: config.sqlite,
-        postgresql: config.postgresql,
-        mongodb: config.mongodb,
-      }
     },
     chartData: function() {
       return {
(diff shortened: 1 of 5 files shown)


askmike:
👍 

So by setting the adapter to mongodb in `web/routes/baseConfig.js` everything works?

tneele:
There seems to be one more issue with the market watcher that still tries to read the sqlite database. I will see if I can come up with a fix.

askmike:
Thanks! Let me know if you need any directions in the codebase.

tneele:
The configuration for market watchers and strategy runners was separated and still set to sqlite. The changes I made in the last commit are somewhat ugly, because the frontend refers to files outside the vue directory. It seems to work though.

I have not included a rebuild of build.js in the commits.

web/vue/src/components/gekko/singleStratrunner.vue
```diff
@@ -84,6 +84,7 @@ import { post } from '../../tools/ajax'
 import spinner from '../global/blockSpinner.vue'
 import chart from '../backtester/result/chartWrapper.vue'
 import paperTradeSummary from '../global/paperTradeSummary.vue'
+import config from '../../../../routes/baseConfig.js'
```
askmike: You don't actually need to include it here since in the route the we already merge whatever is send over by the frontend: https://github.com/askmike/gekko/blob/stable/web/routes/startGekko.js#L19

So you can just not send it in the this file instead (and gekko will pick up what is set in the baseConfig).

askmike:
See my comment, you can just leave them out the frontend and the API will merge it with baseConfig :)

If you fix that I'll merge (you can recompile the frontend, but not necessary)

tneele:
Thanks for the feedback. I will have a look at it again tomorrow.

tneele:
I think I now implemented it in the way you envisioned. Thanks for letting me figure it out myself, I like that.

On a sidenote: there seems to be some more places where sqlite is configured (config/general.toml and config/plugins/candleWriter.toml), but that config is not used I think.

askmike:
> So by setting the adapter to mongodb in web/routes/baseConfig.js everything works?

This should be the case now.

----

Thanks for sticking with me, it looks great. If anyone has a problem with the mongo adapter I'll ping you!

BlackHoleDevs:
@askmike I think that paper trader do not work with mongodb one more time ...

Issue #672, opened 2017-04-05
Shared config makes it hard to run behind a proxy


zeroX-tj:
Hi

at the moment the shared config is a bit of a problem when running behind a proxy. I had to hardcode the port (3000) in the backend. This way i could configure 443 (ssl) for the frontend in the gekko config.

Splitting the config in a server and frontend part would solve it. What also would be cool is if we could bind the backend to an interface (localhost) so it must be accessed via the proxy.

Another thing running behind a proxy that bugged was my websocket connection getting killed. Probably because of inactivity. So i disabled that too in the fronted for now because it isn't used? Or at least i don't use it atm :) Loggers are running in seperate process, i only use the ui to run backtests. (Because i noticed loggers aren't stored and respawned on a restart?) 

I can do a PR, just would like some input first ;)

askmike:
> I had to hardcode the port (3000) in the backend.

Ahhh yes we should split it up.

> What also would be cool is if we could bind the backend to an interface (localhost) so it must be accessed via the proxy.

👍  even better / safer than relying on users to configure a firewall properly.

> Another thing running behind a proxy that bugged was my websocket connection getting killed. 

Interesting, have not noticed this before (I assume this happened when you were not running either an import or a live gekko?). We can easily add some heartbeat in there (I thought websockets had this as part of the spec).

> So i disabled that too in the fronted for now because it isn't used?

Websockets are not used for backtests, but they are for live gekkos and imports. In the future I might use them for backtests too: if the backtest lasts very long (if you are testing over a year or so) the connection might time out.

> (Because i noticed loggers aren't stored and respawned on a restart?)

What exactly do you mean here? But nothing is stored after a backtest (simply pushed to the frontend which will discard after you backtest again).

--------

I'd love to see a PR. We should split up the ports / hosts the backend is running on VS what the frontend is trying to connect to. But I wouldn't split up the config itself: The idea is that you don't have to touch the config for normal usage (you run gekko on your own computer), and if you want to run it in the cloud it would be best to only have to edit "one" config (with a section for frontend and one for backend.

I'm kind of struggling with the config: if you don't want to use the UI you need to use gekko over the commandline with the old config, which is completely unrelated from the UIconfig (makes sense to me, but might be complex for new users - the commandline is also a "UI - User Interface").

zeroX-tj:
> :+1: even better / safer than relying on users to configure a firewall properly.

Well, i would let the user make that decision. But still it could be a default.

[...]

PR #677, opened 2017-04-07
Splitted config

zeroX-tj:
Shared config with ui/api split as discussed in #672 

Let me know if there are any issues. 

There is one change. Gekko used to be accessible on all interfaces, now it is set to 127.0.0.1(localhost) by default. We can easily change the default to 0.0.0.0 so we have old behaviour.


askmike:
Looks great, so I see you opted for a single config file still (that has both backend and frontend configuration)? You had me convinced but both options are fine to me.

zeroX-tj:
If the proxy is properly secured there should be no issue. Now it is easier to edit in one place.

askmike:
👍  thanks :)

PR #688, opened 2017-04-15
Adds config option for tickrate - exchange watch interval.

RatkoR:
The tick interval was hardcoded to 2 seconds for okcoin exchange
and to 20 seconds for any other exchange. This patch adds config variable
`config.watch.tickrate` which can override this defaults.

If tickrate option is not set, the default values will still be used - current
gekko apps will work as before.


askmike:
👍  thanks a lot :)

This will stay an option only available for commandline usage though. For the UI I don't want users to have to worry about this (I will work on a websocket market, but that will take a while - this is great for in the meantime).

Issue #678, opened 2017-04-08
Insane Simulation -- Can this happen for real?? Please just tell me that i am wrong please...


nozyqubis:
![simulation_1](https://cloud.githubusercontent.com/assets/26750756/24826660/b4c8f6fc-1c43-11e7-86bf-ec3d38c4f57f.png)

2017-04-08 01:15:18 (INFO):	(PROFIT REPORT) amount of trades:		 1136
2017-04-08 01:15:18 (INFO):	(PROFIT REPORT) original simulated balance:	 0.03500000 BTC
2017-04-08 01:15:18 (INFO):	(PROFIT REPORT) current simulated balance:	 171.76453140 BTC
2017-04-08 01:15:18 (INFO):	(PROFIT REPORT) simulated profit:		 171.72953140 BTC (490655.80401150%)
2017-04-08 01:15:18 (INFO):	(PROFIT REPORT) simulated yearly profit:	 381.70510050 BTC (1090586.00143971%)

![sim2](https://cloud.githubusercontent.com/assets/26750756/24826840/85496ea8-1c47-11e7-9138-3f241173ffd5.png)

askmike:
Woah, not sure. Some questions:

- What version (branch) of gekko are you running?
- Did you change the paper trade settings (especially the fee & slippage)? If not, what are they set to?

nozyqubis:
Its the Latest Updated Version, i didnt change anything on paper trading, the fee is 0.25% as per poloniex.
I have let running live simulation on one of my VM, will update with the results.I am getting similar results fro different coins, for different timespan.

askmike:
I mean the fee within gekko: if on the backtest page you click on "Change paper trader settings" they should exactly look like this:

![screen shot 2017-04-08 at 11 03 07](https://cloud.githubusercontent.com/assets/969743/24827933/186e763a-1c4b-11e7-89be-76b362fda6e9.png)

Please let me know if they look different.

nozyqubis:
they are the same, i have set to 0 the asset and 0.025 the currency.The only i can think is that i have found the correct numbers :).

askmike:
Could you please send me your database (file inside the history folder) to
mike@mvr.me

And let me know what strat you ran with what settings.

Virkis:
[...]

Issue #607, opened 2017-03-04
Historical data from bitcoincharts.com?


Suko:
Hi, did everyone tried to import historical data from bitcoincharts.com, they have it from biggest markets and it is updated twice a day.

Data can be found here:
http://api.bitcoincharts.com/v1/csv/

askmike:
we can add an importer for bitcoincharts (though some time ago the data was not considered to be 100% accurate, I am not sure what the current status is).

thegamecat:
I've done just that for btce. Took a while but worth it. Take the data, put it into 1m candles then import it into sqlite.

askmike:
I'd love to make that part of Gekko at some point, should be pretty easy to create a [new market file](https://github.com/askmike/gekko/tree/develop/core/markets) that will import a csv export from bitcoincharts and pipes into a gekkostream (even simpler/faster would be a script that inserts it straight into sqlite, but you'd lose flexibility for all of gekkos supported adapters (postgre, mysql, etc).

thegamecat:
Yes, really you just need the candle maker from raw data then spits into the 3 supported dbs.

tuxitor:
I support the idea of a file importer.  At least a couple of sites are offering historical data in JSON and/or CSV format. That would leave the responsibility of fetching data to the user and possibly not pose any licence issues for this project. Another example is cryptocompare.com that has history data. Here is an example:
https://min-api.cryptocompare.com/data/histominute?fsym=BTC&tsym=USD&limit=60&aggregate=1&e=Bitstamp

A file importer could specify a format or be configurable. The obvious choices would be "JSON import" and "CSV import". Mapping a JSON/CSV format to another is usually a trivial task. 

A file importer may possibly be of use for filling holes in the history database.

The dropdown menu should only show supported exchanges.

askmike:
@eigilb that looks great.

> A file importer could specify a format or be configurable. The obvious choices would be "JSON import" and "CSV import".

I'd say we need importers based on providers. We already have a "native" one that doesn't use any providers but the exchanges itself (requires no trust besides the exchange, this is a service given only by some exchanges). We can add additional ones for bitcoincharts and cryptocompare.

This is also handy because the logic needed for each provider is different: bitcoincharts gives trades, which Gekko would need to aggregate into candles, while cryptocompare already gives candles (just pipe them into a database).

> Mapping a JSON/CSV format to another is usually a trivial task.

[...]

PR #697, opened 2017-04-17
Bigger candles

RatkoR:
This PR is about new fields in candles.

We need a couple of new [fields](https://github.com/RatkoR/gekko/blob/bigger_candles/core/budfox/candleCreator.js#L43) in candles. I'd like you to look over the changes and say what is OK and what needs to change to get this beast merged.

The idea is simple: add 4 new fields to candle data and to the DB schema.

The patch is a bit 'ugly', because I didn't want existing users to worry about this change. I added a `version` [config](https://github.com/RatkoR/gekko/blob/bigger_candles/sample-config.js#L335) variable: no version id (or version 1) means candles work as they work now. Version 2 means candles have 4 additional fiels.

If user changes candle versions, he'll have to add these new fields manually to his existing DB. User is intentionally changing config, so he probably knows what he's doing and a short notice or a wiki page would be enough to help him sort out the DB. Normal users won't change this config and will be unaffected. Also, if you change version before creating DB, tables will be created for you correctly.

Since this `version` variable is a numeric one, you can just increment it  to `version: 3` and add other new fields - if you'll ever feel you need some more. What I want to say is that this version thing is expandable.

But.. looking at the patch maybe there's another (better) way. What if we don't add `version` parameter, but rather an array of additional fields user can have in candles. Fields that are in candle data now are mandatory and cannot be removed. But other fields can be just uncommented and you get them in your next candles.

Maybe something like:

```
/**
 * Uncomment to enable additional fields
 * in candles. Note that you'll also have to
 * add new fields to DB if you uncomment it later.
 */

config.candleWriter.additionalFields = [
//  'buyVolume',
//  'buyTrades',
//  'lag',
//  'raw'
]
```

Tell me if you know a better position for `version` or `additionalFields` in config. I didn't find any better place but `candleWriter` node.

----

One last thing: CandleBatcher class had a single argument (`candleSize`). I had to change it to take a second parameter (candleVersion).

Calling it looked strange, because you cannot see immediatelly what those parameters mean. Eg:

```
var cb = new CandleBatcher(2, 1);
```

so I changed it to take options parameter. Eg:

```
var cb = new CandleBatcher({candleSize:2, candleVersion:1});
```

I changed all class invocations and added/changed tests also.

### DIFF
--- a/core/budfox/candleCreator.js
+++ b/core/budfox/candleCreator.js
@@ -16,7 +16,16 @@
 //   ]
 // }
 //
-// Emits 'new candles' event with:
+// Emits 'new candles' event.
+//
+// There are two candle versions currently. Which
+// one is sent depends on the config.candleWriter.version variable.
+//
+// Version 1 is the default.
+// Version 2 has additional fields, see below which ones.
+//
+// Candle version 2 (and above) has also a 'version' field in candle
+// data for quick checking.
 //
 // [
 //     {
@@ -27,7 +36,14 @@
 //       low: (float),
 //       close: (float)
 //       volume: (float)
-//       vwp: (float) // volume weighted price
+//       vwp: (float) // volume weighted price,
+//       trades: (integer), // nb of all trades
+//       // ---- v2 fields -----
+//       version: 2,
+//       buyVolume: (float), // volume of buy trades
+//       buyTrades: (integer), // nb of buy trades,
+//       lag: (integer), // exchange lag,
+//       raw: (array) // array of all trades
 //    },
 //    {
 //       start: (moment), // + 1
@@ -38,6 +54,13 @@
 //      close: (float)
 //       volume: (float)
 //       vwp: (float) // volume weighted price
+//       trades: (integer), // nb of all trades
+//       // ---- v2 fields -----
+//       version: 2,
+//       buyVolume: (float), // volume of buy trades
+//       buyTrades: (integer), // nb of buy trades,
+//       lag: (integer), // exchange lag,
+//       raw: (array) // array of all trades
 //    }
 //    // etc.
 // ]
@@ -45,6 +68,7 @@
 
 var _ = require('lodash');
 var moment = require('moment');
+var config = require('../../core/util.js').getConfig();
 
 var util = require(__dirname + '/../util');
 
@@ -56,6 +80,9 @@ var CandleCreator = function() {
 
   // This also holds the leftover between fetches
   this.buckets = {};
+
+  // exchange lag
+  this.lag = 0;
 }
 
 util.makeEventEmitter(CandleCreator);
@@ -66,11 +93,15 @@ CandleCreator.prototype.write = function(batch) {
   if(_.isEmpty(trades))
     return;
 
+  this.lag = batch.lag;
+
+  var candleVersion = config.candleWriter.version;
+
   trades = this.filter(trades);
   this.fillBuckets(trades);
-  var candles = this.calculateCandles();
+  var candles = this.calculateCandles(candleVersion);
 
-  candles = this.addEmptyCandles(candles);
+  candles = this.addEmptyCandles(candles, candleVersion);
 
   // the last candle is not complete
   this.threshold = candles.pop().start;
@@ -101,7 +132,7 @@ CandleCreator.prototype.fillBuckets = function(trades) {
 }
 
 // convert each bucket into a candle
-CandleCreator.prototype.calculateCandles = function() {
+CandleCreator.prototype.calculateCandles = function(candleVersion) {
   var minutes = _.size(this.buckets);
 
   // catch error from high volume getTrades
@@ -110,7 +141,7 @@ CandleCreator.prototype.calculateCandles = function() {
     var lastMinute = this.lastTrade.date.format('YYYY-MM-DD HH:mm');
 
   var candles = _.map(this.buckets, function(bucket, name) {
-    var candle = this.calculateCandle(bucket);
+    var candle = this.calculateCandle(bucket, candleVersion);
 
     // clean all buckets, except the last one:
     // this candle is not complete
@@ -123,7 +154,7 @@ CandleCreator.prototype.calculateCandles = function() {
   return candles;
 }
 
-CandleCreator.prototype.calculateCandle = function(trades) {
+CandleCreator.prototype.calculateCandle = function(trades, candleVersion) {
   var first = _.first(trades);
 
   var f = parseFloat;
@@ -139,11 +170,48 @@ CandleCreator.prototype.calculateCandle = function(trades) {
     trades: _.size(trades)
   };
 
+  if(candleVersion === 2) {
+    candle = _.merge(candle, {
+      version: candleVersion,
+      buyVolume: 0,
+      buyTrades: 0,
+      lag: this.lag,
+      raw: trades
+    });
+  }
+
+  /**
+   * Returns trade type ('buy' or 'sell').
+   *
+   * If current trade has higher price then the previous one, it must
+   * have been a buy action. Or if previous trade was a 'buy' and
+   * price hasn't changed, we assume users are still buying.
+   */
+  var getTradeType = function (trade) {
+    var newTradePriceIsHigher = (this.lastTradePrice < trade.price);
+    var newTradeRemainsBuy = ((this.lastTradePrice == trade.price) && (this.isBuyAction));
+
+    return (newTradePriceIsHigher || newTradeRemainsBuy) ? 'buy' : 'sell';
+  }
+
+  this.lastTradePrice = 0.0;
+  this.isBuyAction = false;
+
   _.each(trades, function(trade) {
     candle.high = _.max([candle.high, f(trade.price)]);
     candle.low = _.min([candle.low, f(trade.price)]);
     candle.volume += f(trade.amount);
     candle.vwp += f(trade.price) * f(trade.amount);
+
+    if(candleVersion === 2) {
+      this.isBuyAction = getTradeType(trade) === 'buy';
+      this.lastTradePrice = trade.price;
+
+      if (this.isBuyAction) {
+        candle.buyVolume += f(trade.amount);
+        candle.buyTrades++;
+      }
+    }
   });
 
   candle.vwp /= candle.volume;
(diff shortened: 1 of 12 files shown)


askmike:
Hey, great idea!

I have been thinking about updated marked data for a while and there are a few things to note:

- Gekko [already has a versioning system for persisting data](https://github.com/askmike/gekko/blob/stable/plugins/sqlite/handle.js#L31), it would be better if we could leverage that instead. The version is driven from the config so Gekko will just use whatever the config has configured.
- If we go to a new version, we really need spread information in there! (bestAsk & bestBid). See #678.
- If you need different behaviour for the candleBatcher, I rather make a new class (candleBatcherv2 or something). They can point to specific specs about fields. Assuming v0 when a version key can not be found does not sound ideal to me (when config is misconfigured and Gekko assumes you want the old version, things like that are very hard to debug).

askmike:
Also a tiny nitpick thing (which you could not have known) is terminology wise: **Bigger candles**

Gekko used to have a concept of smallCandles and bigCandles for over a year. smallCandles used to be 1 minute candles, and bigCandles had the same content but they covered the period configured in candleSize. Everyone coming back to Gekko will be left very confused.

Bigger candles also makes it hard to ever go to v3, will we call them *Biggest candles?* And what about v4?

RatkoR:
Hey, thanks for the feedback, I appreciate it!

I think I pushed too much too quick this time. Let's just talk and forget about the PR for now.

Candle has currently [these](https://github.com/askmike/gekko/blob/develop/core/budfox/candleCreator.js#L23) fields:

```
     {
       start: (moment),
       end: (moment),
       high: (float),
       open: (float),
       low: (float),
       close: (float)
       volume: (float)
       vwp: (float) // volume weighted price
    },
```

Let's say I wanted to add `number of buy trades` that happened in this candle's timeline. What would be the preferred way to do this? Note that we probably shouldn't disturb current users, because 99% of them are not interested in this additional field. But some (1%) are interested and would like to enable it somehow for them also.

This would be the candle emited *and saved in the DB* after the change.

```
     {
       start: (moment),
       end: (moment),
       high: (float),
       open: (float),
       low: (float),
       close: (float)
       volume: (float)
       vwp: (float) // volume weighted price,
       buyVolume: (integer) // <-- new field
    },
```

Thanks! (I know I should have asked all this before doing the PR...)

askmike:
Yes let's have a quick discussion about this! Sorry for not being able to accept it like it is now...

> Let's say I wanted to add number of buy trades that happened in this candle's timeline.

Just to be on the same line: I guess you mean the volume of trades where asks (sell offers) were taken from the orderbook (eg. bought)? Because in every trade there is a buyer as well as a seller.

> This would be the candle emited and saved in the DB after the change.

Yes that looks great, we can add extra fields to the database. But we need to keep the following scenarios in mind:

- What if someone has been running Gekko (without `buyVolume`) for a long time and they update to this. Would we then ignore all old data since the DB table doesn't have this `buyvolume` field? Or should we "migrate" the data (and do a) assume it's half the volume b) set to 0 or c) allow NULL values).
- Gekko can connect to 17 different exchanges, do all these exchanges give data about `buyVolume`? What do we do if an exchange does not? (we could use [this file](https://github.com/askmike/gekko/blob/develop/exchanges.js) were we already store related information).
- If `buyVolume` what we should go for right now? (Not saying it is not, maybe we should come up with a list of properties that might be useful - I am thinking mainly of spread related properties).

RatkoR:
> Yes let's have a quick discussion about this! Sorry for not being able to accept it like it is now...

Absolutely nothing wrong with this.

> Just to be on the same line: I guess you mean the volume of trades where asks (sell offers) were taken from the orderbook (eg. bought)? Because in every trade there is a buyer as well as a seller.

Yes, exactlty!

> What if someone has been running Gekko (without buyVolume) for a long time and they update to this. Would we then ignore all old data since the DB table doesn't have this buyvolume field? Or should we "migrate" the data (and do a) assume it's half the volume b) set to 0 or c) allow NULL values).

I'd go with a notice. If someone is changing his config (assuming that this will be a config variable) then he probably has such knowledge to go look in the wiki and change his tables accordingly. What do you think?

I wouldn't change user tables automatically - I think too many problems can happen. Let user handle his tables.

> Gekko can connect to 17 different exchanges, do all these exchanges give data about buyVolume? What do we do if an exchange does not?

We're [checking](https://github.com/RatkoR/gekko/blob/bigger_candles/core/budfox/candleCreator.js#L190) for every trade if it is a buy or sell action. We don't look for specific fields in exchange return because (as you said) not every exchange returns this info. If price is higher then before then it must have been a sell.

> If buyVolume what we should go for right now?

I was thinking of array of all possible fields. If someone wants to add another one, he'd have to add it to this list and add a getter function that returns value for it.

RatkoR:
How about this new config variable?

```
config.candleExtraFields = [
  'buyVolume',
  'buyTrades',
  // 'allTrades'
]
```

As I wrote before, the objective is to allow user to add additional fields to candles: if field is in `candleExtraFields` array it will be emitted in candle data.

In this example, `buyVolume` and `buyTrades` fields are part of candle data, `allTrades` is commented out and hence it is not part of candle data.

If this seems reasonable we'll move to next discussion (about DB changes). For now, let's agree on config change first.

askmike:
Sorry for the late reply, swamped with work atm!

>> Just to be on the same line: I guess you mean the volume of trades where asks (sell offers) were taken from the orderbook (eg. bought)? Because in every trade there is a buyer as well as a seller.
>
> Yes, exactlty!

To know this correctly we would need to watch the orderbook in realtime. Something that comes at a big performance price (though as soon as we do this we can of course provide a lot of other very useful information). Or am I missing something obvious?

Your proposal is build on a main assumption `Or if previous trade was a 'buy' and price hasn't changed, we assume users are still buying.` which I think is a pretty dangerous one to make. Especially in high volatile times when some exchanges don't provide trades in the order they were executed (I've only seen this in smaller exchanges, but still).

> I'd go with a notice. If someone is changing his config (assuming that this will be a config variable) then he probably has such knowledge to go look in the wiki and change his tables accordingly. What do you think?
> 
> I wouldn't change user tables automatically - I think too many problems can happen. Let user handle his tables.

The thing is that I am really pushing towards UI usage, where configuring this is a lot harder. The config was great and still is a great way to tweak almost everything Gekko is doing. But I want to go more in a direction where the "default" settings get you pretty far. Turning Gekko a bit more into a "magic box" (where you can still tweak everything, but only if you really need to).

People who want to tweak everything that happens without Gekko would probably just open the code and start hacking anyway.

> How about this new config variable?
> As I wrote before, the objective is to allow user to add additional fields to candles: if field is in candleExtraFields array it will be emitted in candle data.

The reason why I am a bit hesitant is:

Right now people can simply exchange history databases and everything continues to work. The only reason to add props to the candles is to be able to create strategies that rely on more market data (or not?). But as soon as you do this you need a way to handle strategies trying to run that rely on more data than currently configured. And if they configure their "config" (or UI) to use extra props, these might not be available in already imported datasets.

On top of that there are some discussions (example: #607) about importing data from other sources than exchanges. There are a lot of sources that have candle (OHCL) data, but I don't know any that also include buyTrades. So we need to think about handling cases where people have strategies relying on buyVolume but want to import from a source that does not include this data.

I'm all for more props in candles, but if you make it configurable it will become very hard to support imo.

askmike:
And about including all the trades in the candles. I'm really against this. We should just expose all trades in a way ([budfox](https://gekko.wizb.it/docs/internals/budfox.html) already has this data). I wouldn't see why you want to batch these into candles.

RatkoR:
> Your proposal is build on a main assumption Or if previous trade was a 'buy' and price hasn't changed, we assume users are still buying. which I think is a pretty dangerous one to make. Especially in high volatile times when some exchanges don't provide trades in the order they were executed (I've only seen this in smaller exchanges, but still).

Yes, that's true, it could really be a problem if trades are not in the correct order.

> People who want to tweak everything that happens without Gekko would probably just open the code and start hacking anyway.

That's what we're doing now. But it's hard to keep code in sync between the forks.

> Right now people can simply exchange history databases and everything continues to work

Good one, I didn't think about this one.

> The only reason to add props to the candles is to be able to create strategies that rely on more market data. (or not?)

Yes, definitely!

> And about including all the trades in the candles. I'm really against this. We should just expose all trades in a way (budfox already has this data). I wouldn't see why you want to batch these into candles.

Our strategy calculates some additional data from those raw trades. And instead of adding those weird fields to candles, so we just pass all trades. But I see your point, we should stuff trades somewhere else.

RatkoR:
Ok, so what do we do now?

Another idea: what about some kind of plugin support? The data for these additional fields (buyVolume, etc...) would be in other DB table(s), so it wouldn't broke main candle table.

And the strategies would then load plugin data with something like:

```
var pluginData = candle.loadFromPlugin('pluginId');
```

In the main gekko, we'd only need to add code to support these kind of plugins. I'd have a lot less work keeping branches in sync.

askmike:
> Ok, so what do we do now?

As you've noticed I am quite reluctant to adding code and complexity to hot code paths that won't be needed except for by a few people.

> Another idea: what about some kind of plugin support? The data for these additional fields (buyVolume, etc...) would be in other DB table(s), so it wouldn't broke main candle table.

Great idea! So if you look at how Gekko currently works:

There are two main parts: a market and a gekkostream (stuff that deals with candles, strategies etc)

gekko architecture

In the picture above we are looking at the live market (called budfox). This is the part that connects to exchanges and makes sure everything else gets fed candles.  Read more about budfox here: https://gekko.wizb.it/docs/internals/budfox.html

Budfox uses a bunch of modules to transform trades into candles (tradeBatcher, candleCreator, , writing candles to DB, etc.) and wraps the whole thing in one node stream that only exposes candles.

-----

The best in this ecase would be to add a new market (that reuses most modules but monkey patches say the candleCreator by copying it to `gekko/core/bigCandles/candleCreator` or whatever you want to call it and add whatever props you want. You can also choose to not only expose a list of candles but also trades (as they happen for example). If you change what is broadcasted (not just candles but also X) you do need to rewire a lot of other stuff, so this is not recommended.

----

TLDR: If you want custom behaviour that almost no one else wants I don't want to add that to code being run by everyone (unless it's very trivial). If we can get support for new candle props by more of people we can of course still add it to the main code!

If you have any questions about any part of the codebase feel free to ask :)

RatkoR:
Ok, agreed, I'm fine with this. I'll let you know if I have any questions with the new approach.

askmike:
👍  Sorry for not being able to accept this. And yes please! If you have any questions or ideas just let me know.

I understand how frustrating it is to have to maintain a fork, so I can definitely merge stuff that is not used on default (just like adapters), but I want to do it in a way so that it doesn't complicate / slow down gekko for everyone else!

Issue #705, opened 2017-04-20
Different select boxes for assets and currencies for the importer


tudorconstantin:
When selecting the market (asset/currency pair), there's currently a single select box, which makes it quite annoying to scroll and search for BTC/ETH for example.

If we'd have 2 select boxes, then we'd be able to quickly select BTC as the currency and then click on the second select box, press 'e' a few times until ETH gets selected

askmike:
You are talking about both importing as well as a live market?

This was actually a part of a very early version of the UI, you had to select currency & asset. ~That code is still there and can be found in [configbuilder/marketpicker](https://github.com/askmike/gekko/blob/stable/web/vue/src/components/global/configbuilder/marketpicker.vue).~ I might have removed that code!

tudorconstantin:
Was there a reason why you chose back then to use a single select box? 

Do you think the same reason still exists with my updates also? Or, is there something more I can do to overcome it?

PS: I like it how it looks with the select boxes on the same line

tudorconstantin:
I think this can be closed, since you already merged the changes that implement this

PR #706, opened 2017-04-20
Different select boxes for assets and currencies for the importer

tudorconstantin:
This PR is the implementation for issue #705 that I just opened.


tudorconstantin:
I think the only way to get rid of the conflict in the build.js file is to re-generate the file after merging the pull request

askmike:
Thanks for this!

funcOK but the commits are not really that clean, would it be possible to fix that by either rebasing or cherrypicking into a new branch? That should also solve the conflict in the compiled frontend

tudorconstantin:
I'm glad you also find this useful.

I submitted a totally [new PR](https://github.com/askmike/gekko/pull/707), containing 2 commits:
- one with the modified source files
- one with the overwritten build.js file.

askmike:
closed in favor of #707 .

PR #707, opened 2017-04-21
Different select boxes for assets and currencies for the importer

tudorconstantin:
I've re-done the PR for implementing issue #705


askmike:
Hey commits look great now! Sorry to be so nitpicky but this does changes the UI a lot:

screen shot 2017-04-21 at 10 26 43

I'd rather see currency and asset next to each other so that the marketpicker is the same height as the typepicker (in live gekko) and daterange selector (in import).

But if you don't feel like doing it let me know, in that case I'll merge the PR and update it myself.

tudorconstantin:
I've updated the pull request with the new layout:

gekkomarket

tudorconstantin:
Is there something that prevents this PR to be merged or rejected?

I don't want to be too pushy, I like the project and I want to contribute further - in a helpful manner, not to bother with noisy/useless functionality

askmike:
Sorry did not have any time over the weekend! Looking at it right now :)

askmike:
It's currently not working because the asset and currency are flipped (gekko should have complained about this, but it isn't). So for me it's not working as is: since on Poloniex the USDT is the currency and BTC the asset, and this set's it as the other way around.

I can also fix this myself if you'd rather, just let me know :)

EDIT: it appears just the defaults are wrong, you are populating the correct lists for the dropdown.

tudorconstantin:
I had doubts regarding what's an asset and what's a currency myself, but somehow I was able to determine it from the exchanges.js file.

Thanks for accepting the PR. It feels great to me

PR #709, opened 2017-04-22
include secure reverse proxy setup

generalectric:
Updated gekko ui server install docs to include easy secure reverse proxy setup so people can securely access their gekko ui remotely from schoo.. i mean home or wherever =) I spent hours banging my head against a wall so figured I'd save people some hair as it wasn't so straight forward... for me anyways. Happy backtesting!


vietlongn:
@generalectric Well done !
There is a missing  "}" at the end of your nginx config thought :)

askmike:
Sorry for the delayed response!

Looks very good :)

One question: why the section on self signed certs? If people end up doing this they will only see flashing warnings in their browsers.

generalectric:
I added it so people that dont have a fqdn can still use dyndns with a self signed cert though it does produce a warning occasionally

askmike:
Live here as well:

https://gekko.wizb.it/docs/installation/installing_gekko_on_a_server.html#Configuring-NGINX-as-a-reverse-proxy

rcoenen:
We should have some instruction for EC2 / AWS too dealing with *unsceure* Nginx setups... AC2 makes it impossible to generate the SSL certificates!

PR #713, opened 2017-04-26
Bugfix: mongodb reader query sort order and history size

freshpatze:


askmike:
👍 

Quick question: how are you using the mongoDB adapter? I don't think all features of the UI work since this adapter is a bit outdated.

freshpatze:
Was going to use it without the UI actually, but I think I am going for postgresql now ;-)

Issue #716, opened 2017-04-28
[feature request] UI improvements


tudorconstantin:
I think that implementing the below features would be **highly appreciated** by the ones who are using gekko
- add the option to close the modal when the connection is re-established
- add the ability to click on any elements in the row when selecting the dataset in backtesting

tudorconstantin:
I've implemented the above and submitted the PR #717

askmike:
I commented on the PR!

> add the option to close the modal when the connection is re-established

The reason for the modal after reconnection is that the state in the web UI is broken:

When you load the UI in your browser the UI will fetch state of current gekkos (live gekkos, imports, etc), after that websockets are used to keep the state up to date. But in case the UI is reconnected you might have missed a ton of websocket messages (maybe the UI was disconnected for X hours), so all state displayed in the UI can be out of date. To fix we need to refetch state and the easiest way to do this (right now) is refresh the browser.

askmike:
Let's keep the discussion about the reconnection modal in the PR instead of having it in 2 different places!

askmike:
Both of improvements have been merged (into develop branch, will merge into stable branch very soon). So I am closing this already :)

PR #717, opened 2017-04-28
Implementation for issue #716

tudorconstantin:
I think that implementing the below features would be **highly appreciated** by the ones who are using gekko

- add the option to close the modal when the connection is re-established
- add the ability to click on any elements in the row when selecting the dataset in backtesting

The overlay looks like this with the Close button added:
close_button


web/vue/src/components/global/mixins/dataset.js
```diff
@@ -21,13 +21,16 @@ var mixin = {
 
         response.datasets.forEach(market => {
           market.ranges.forEach(range => {
+            let num = 1;
```
askmike, line 24: You don't have to manually count. The forEach function provides a second argument called index which will increment once per iteration.

askmike:
I really like this idea! Could you actually instead of closing just refresh the browser when hitting the button (and call it "refresh")?

Also the commits are not super clean (preventing me to merge this), but I can manually cherry-pick the commits of interest as well into the develop branch.

web/vue/src/components/global/mixins/dataset.js
tudorconstantin: great observation, I totally forgot about the second param of the forEach. Will modify it and resubmit

tudorconstantin:
I can add a second button for refresh.

I liked the idea for closing the overlay, because it annoyed me its appearance when I was in the middle of something, like backtesting: chose the currency/asset pair, customize some of the strategy params and then, right before customizing the candle size and hitting the backtest button, the overlay appeared, making the refresh throw away the whole customizations. 
And the same would happen when customizing an import, after I'd spend some time in carefully setting up the from:/to: params, the overlay would appear and in order to make it go away I would have to refresh, having to stat all over again.

As a hack, I used to manually get rid of the overlays with the help of Chrome Inspector Tools and I didn't had any issues with the app working as it should from there on.

I was curious, why do you prefer the refresh? are there some potential issues with Vue's internal state if the overlay is closed?

askmike:
> because it annoyed me its appearance when I was in the middle of something, like backtesting: chose the currency/asset pair, customize some of the strategy params and then, right before customizing the candle size and hitting the backtest button, the overlay appeared, making the refresh throw away the whole customizations.

Woa it sounds like you are seeing this error a lot. You should only really see this if either gekko goes down or if you lost connection to Gekko. Are you randomly getting this modal? If so how are you running Gekko?

tudorconstantin:
I installed an instance of it on a dedicated server that I have at Hetzner. 
In the first instance I connected to gekko directly and then I configured it behind nginx. 
In both cases the disconnect/reconnect issue comes up quite frequent, but I wasn't able to either find a pattern, nor to identify the real issue (I haven't put too much time in investigating it, since I can get rid of the overlay quite easy).

It reconnects quite fast (a few seconds) and the node process is not showing any errors - I assume there must be some network related error.

askmike:
> It reconnects quite fast (a few seconds) and the node process is not showing any errors - I assume there must be some network related error.

Woah that is pretty bad.. So the UI uses websockets to keep a constant connection between the UI and Gekko. (I explained why this is needed [in the issue](https://github.com/askmike/gekko/issues/716#issuecomment-298174906)).

So for some reason your browser is unable to keep a constant connection to Gekko, under normal network conditions this should rarely happen. Do you also have other problems when using the web? Almost all websites that have some realtime aspect (facebook, twitter, slack, discord, etc) use websockets for this, do you also have problems with those?

To answer your question:

> I was curious, why do you prefer the refresh? are there some potential issues with Vue's internal state if the overlay is closed?

Yes, state might not be complete or up to date (see comment in #716). Refreshing this is a hacky way to sync full state from Gekko. A better way would be:

**On reconnect just fetch & sync the data from gekko, not displaying the modal nor requiring a refresh.**

Syncing state is as easy as calling `initializeState` from `web/vue/src/store/init` (see [here](https://github.com/askmike/gekko/blob/stable/web/vue/src/main.js#L42)). We can simply run that function on reconnect to fix the underlying problem. That would be a much simpler PR :) Let me know if you want to do this.

What that does is fetch state from gekko and set it in vuex (causing the full UI to have updated data).

EDIT: if you do create a PR for this please also remove the reconnect modal, since we would never use that anymore.

tudorconstantin:
I think the overlay is useful in the case the app doesn't reconnect fast. So I think it would be even better to do it like:

1. on disconnect - show the overlay
2. on reconnect - call `initializeState` and hide the overlay

askmike:
Yes that is exactly what I meant! We are on the same line :)

tudorconstantin:
I've submitted another PR, just for the part with the overlay.

I'm adding the modifications for the second part, the one with transforming the table cell content in `<label for=..."">` in another PR

askmike:
closing this for PRS #720 and #721.

PR #720, opened 2017-04-29
implementing first point for #716, as discussed in #717

tudorconstantin:


askmike:
Great stuff! I don't want to be too nitpicky, but I was thinking of syncing state outside the modal ([around here](https://github.com/askmike/gekko/blob/stable/web/vue/src/components/global/ws.js#L40)). I can also do that after this PR.

tudorconstantin:
indeed it looks a bit out of its place to have a call to `initializeState` in the modal, 
but I think that if I'd have to debug things like:
- _why is the modal not closing automatically anymore?_
- _why is the data refreshed when the modal is closing?_

The modal component would be the first place to look into as a dev and I'd see the call to `initializeState` which would help me investigating further. 

Also, I think it makes it clearer what happens when the modal is closing: the state is reinitilized

But, as you said, it seems ultra logic to have the re-initialization happen in the reconnect method itself.

So, what would be your final decision on this?

askmike:
Good questions:

Both ways would definitely work, below is simply my opinion:

> why is the modal not closing automatically anymore?
> why is the data refreshed when the modal is closing?

If they end up looking at the modal code they would see a reference to `this.$store.state.warnings.connected` which hopefully gets them far enough as well. But this is the whole idea behind vuex/redux: You want to decouple state from UI in every possible way. If we would leave it in the modal but later build some other logic that doesn't use modals but a notification system (just an example), we would not resync state because this is dependent on the modal being loaded. If we want to couple everything happening directly to things visible on the screen we should not use vuex.

> So, what would be your final decision on this?

Put it in the network code (`ws.js`). So in ws.js onconnect sync state (and remove the initializeState in `main.js` - else we do it twice when the UI loads). Simplify the vuex store so we only have a `this.$store.state.warnings.connected` (not the reconnected anymore since UI won't care about it anymore). Remove the reconnected modal (you already did this) and simply toggle the modal based on `this.$store.state.warnings.connected` (you also already did this).

If you have any trouble with any of this (or don't want to do it fully) let me know and I can do it :) I can accept the PR but I would change the code until it is like above (unless you have some convincing arguments).

tudorconstantin:
Indeed, the decoupling argument is quite strong. 
I've implemented the changes and updated the PR.
Hopefully I haven't missed anything :smiley:

askmike:
That is everything, great PR :)

-----

Another reason I implemented the reconnected modal is so that if your Gekko restarts you'll see a message and have to manually take action. Where now if you are running a live Gekko and your Gekko restarts the UI will show this:

screen shot 2017-04-30 at 00 04 04

(Because we don't persist running gekkos and therefor the restarted gekko does not "remember" anything from before the crash). But the reconnected modal was a bad way to make this clear to the user, a notification or so (without the user having to take action) would make more sense. I might add that in the future.

PR #721, opened 2017-04-29
implementing second point for #716

tudorconstantin:
The `build.js` file needs to be re-generated. 

I haven't included it here because it would cause conflicts if other UI updates are made between now and the moment of the merge.


askmike:
👍  thanks :)

PR #722, opened 2017-04-30
add ipv6 to example nginx config

generalectric:
ipv6 addition to example nginx setup config as per #710


askmike:
That's fast haha. Great thanks!

Issue #746, opened 2017-05-14
Config file for exchanges


RatkoR:
A question about an [exchanges.js](https://github.com/askmike/gekko/blob/stable/exchanges.js) file.

Would it make sense if every exchange in the `/exchanges` dir has it's info there? For example, the [poloniex](https://github.com/askmike/gekko/blob/stable/exchanges/poloniex.js) exchange file would have it's capabilities listed in it's .js file.

Currently, exchange data is located in two places (in `exchanges.js` and in `exchanges/poloniex.js`). Adding a new exchange is troublesome, because it's not enough to just add a new file in the exchanges dir, you have to change the exchanges.js file. Which is bad, because this file is managed by the upstream branch. What I want to say is that one cannot just drop a new exchange in the `exchanges/` dir and play with it.

In case of poloniex.js (and every other exchange), it could have a `capabiliteis()` method with the exact same info as is now in exchanges.js:

``` 
Trader.prototype.capabilities = function() {
    return {
        currencies: ['BTC', 'ETH', 'XMR', 'USDT'],
	assets: [
          '1CR', 'ABY', 'AC', 'ACH', 'ADN', 'AEON', 'AERO', 'AIR', 'AMP', 'APH', ...
        ],
        markets: [
          { pair: ['BTC', 'YC'], minimalOrder: { amount: 0.0001, unit: 'asset' } },
          ....
        ],
        requires: ['key', 'secret'],
        tid: 'tid',
        providesHistory: 'date',
        providesFullHistory: true
    }
}
```

askmike:
Interesting idea! I like the exchanges.js file a lot since you can see and compare (and copy) different exchanges at a glance, but it's getting out of hand with the gazillion markets supported for Poloniex.

But if we are going for this it would require changing this for all supported exchanges (not something I feel like doing). Also note that: the UI imports this file to render the interface, if you are going to split it up, also provide a way to retrieve this data from the frontend (API call I guess).

Talking about the UI:

> Adding a new exchange is troublesome, because it's not enough to just add a new file in the exchanges dir, you have to change the exchanges.js file.

You also need to recompile the frontend, so if you want this to be more like "drop a file into `/exchanges` and it works" make it fully dynamic so the frontend redraws the UI based on files in this folder (or possibly on refresh).

askmike:
discussion moved to #750.

PR #750, opened 2017-05-16
Exchange Capabilities

RatkoR:
Mike, this is just a preview regarding the #746 issue.

I wanted to give you a head start on this and ask if it's still OK to do it. I feel it is, since every exchange now has it's own data inside it's own js file. But it's your call.

This patch series adds the `/exchanges` API which returns the same json as was in the `exchanges.js` file. It reads values from each exchange js file (by invoking [getCapabilities](https://github.com/RatkoR/gekko/blob/capabilities/exchanges/bitstamp.js#L163) method. UI works OK with it - see [here](https://github.com/RatkoR/gekko/blob/capabilities/web/vue/src/components/global/configbuilder/marketpicker.vue#L41).

There more work: fix all exchanges, fix exchangeChecker, etc... I'll do it if you're OK.


askmike:
Nice and clean implementation! Though it is limited to 3 exchanges. I'll leave this open for a while until I figure out what (I think) is best for Gekko :)

RatkoR:
No, no, this PR is not ready yet :) I added `getCapabilities` method to those three exchanges because I wanted to show you how an exchange would look like. And so that you can stop me before I loose too much time on it.

I'll change **all** exchanges before submitting this PR for real. I'll also fix `exchangeChaecker.js` and any other file that uses the `exchanges.js` file, since it is not going to be there any more.

askmike:
If you do all that & also update this documentation: https://gekko.wizb.it/docs/extending/add_an_exchange.html (by changing this file: https://github.com/askmike/gekko/blob/stable/docs/extending/add_an_exchange.md ) I have no choice but the accept :)

core/exchangeChecker.js
```diff
@@ -12,10 +14,15 @@ Checker.prototype.notValid = function(conf) {
...
+  return Trader.getCapabilities();
+}
+
 // check if the exchange is configured correctly for monitoring
 Checker.prototype.cantMonitor = function(conf) {
   var slug = conf.exchange.toLowerCase();
-  var exchange = _.find(exchanges, function(e) { return e.slug === slug });
+  var exchange = this.getExchangeCapabilities(slug);
 
   if(!exchange)
```
askmike, line 35: this if check won't get hit anymore, because [this line](https://github.com/askmike/gekko/pull/750/commits/02a62896e4d27adc2e04bc29f1382da9680461c7#diff-2a878ed84810bccfebe021d0e89d7ca8R18) will crash if the file is not found.

RatkoR: True, and we don't want crashes - I'll add a try / catch block. Good catch!

RatkoR:
Allright Mike, I've tried as many ways to run gekko as I could: importer, ui, paper trading, npm tests - they all seemed to work OK. I didn't do realtime trading (I'm not trading real time with gekko yet). :)

I'd be glad if you can check the code and tell me what to change.

askmike:
Thanks a lot! Will try to check it out asap and let you know :)

askmike:
Sorry for the delay! Tested a bunch of features and everything seemed to work :) Will merge into develop and will do more thourough testing (together with #737) before it hits stable.

PR #766, opened 2017-05-31
added currency to mailer.js

PhillRob:


PhillRob:
Makes sense to not only say value and exchange but currnecy as well.

askmike:
Nice catch :) thanks!

PR #777, opened 2017-06-08
Retry certain operations before crashing

andreialecu:
This PR adds retrying to certain requests so that gekko doesn't instantly crash when there's a network error. The mailer was the most problematic, but I've also seen it crash due to the exchange ticker not being available.


askmike:
Sorry for the late reply! Looks great :) We should probably move all retries here (so all the implemented retries within the exchange implementations)

PR #802, opened 2017-06-25
Add GNT (Golem) as an Poloniex asset

PW00X:
Added GNT (Golem) as an Poloniex asset so data import is possible.


askmike:
In the new version of Gekko (the development branch as well as all feature branches) these attributes have been moved into individual exchange files, see #750. In order for this to work you now need to edit this line instead:

https://github.com/askmike/gekko/blob/develop/exchanges/poloniex.js#L193

askmike:
I am closing this, if you would like to see this change but don't see what you'd need to change. Just let me know :)

PW00X:
Changes done in #867

Issue #628, opened 2017-03-16
buying ETH on GDAX


brian000:
I'm running the development branch with default configuration (except GDAX API key) and a couple bucks for testing, however gekko doesn't seem to actually buy any ETH.  It logs the attempted buy but eventually says the order wasn't filled, cancels it, and doesn't attempt to buy again.  If I view GDAX in a browser, I don't see the order in my fills.

It doesn't seem to have any problems selling, but buying is problematic.  I understand why after a minute it doesn't necessarily create a new buy order, as the price has probably changed, but the initial buy just never happens.

```
 2017-03-15 13:02:32 (INFO):     Attempting to BUY 0.07866869444150902 ETH at GDAX
 *CHOP*
 2017-03-15 13:03:33 (INFO):     BUY order was not (fully) filled, cancelling and creating new order
```

Anyone else seeing issues like this? 

Edit: nodejs v7.7.2, commit 7988cce

askmike:
Strange.. Will have a look asap! After #611 I'll go over all bugs regarding exchange integrations.

smurfy:
The problem probably is ```post_only: true``` try disabling it.
it probably makes sense to default to ```false```, while adding support i noticed the problem too, if you debug the error message the order has the status "rejected" while mentioning post_only.

@askmike feel free to mention me for gdax related problems so i'm aware of them.

smurfy:
Ah just noticed the default (if you remove post_only from the config) it actually defaults to false.
I also mentioned the reject problem at #564 
If its NOT post_only  what are you using ETH/USD or ETH/BTC   (i only tested BTC/USD on sandbox and ETH/BTC on production)

brian000:
I'm using the ETH/USD pair, tried adding `post_only: true` to my `config.trader`.   (Wasn't in the default config file.)  The bot's off and running, I'll chime back in with a trip report once it tries to buy something.

Edit: no good, it did the same thing as my OP, it said it was attempting to buy but it never actually happened.

boxofdeath:
I am having the same issue using ETH/BTC. However, it attempts to buy repeatedly.

smurfy:
@brian000 @boxofdeath post_only   should  false.  If the problem still occur please tell me.
I will test the trader tonight with sandbox mode.

brian000:
Gave it a crack with `post_only: false` and it still doesn't buy.  It tries, gives up, then keeps trying a couple more times, just like my initial report.

boxofdeath:
[...]

PR #853, opened 2017-07-16
Fix undefined errors in gdax.js

nourspace:
- Changed error to err
- Unified errors checking and displaying
  `err || data.hasOwnProperty('message')`
- Added missing bind to the newly added result callbacks
- Added missing args to retry in sell

Refs #628


askmike:
Woah this is great, thanks!

nourspace:
No problem. I am testing GDAX and it seems there are few cases that could be handled in a better way.
I will make a list of my findings and create a PR with possible fixes.

PR #867, opened 2017-07-20
Add GNT (Golem) as an asset

PW00X:


askmike:
thanks for updating the previous PR 👍

PR #765, opened 2017-05-30
added passphrase, irc module version update

PhillRob:

PR #874, opened 2017-07-23
gekko v0.5.0

askmike:
This new version adds a few upgrades, most notably:

- number of exchange fixes + supports live traders in the UI (#737)
  - new performanceAnalyzer plugin
- remove legacy exchanges.js file and move it into getCapabilities methods per exchange (#750 by @RatkoR )
- plugin fixes (#765 and #766 by @PhillRob)
- GDAX fixes (#853 by @mochawich)
- twitter plugin (#877 by @ztnark)

And fixed a few async race conditions in the backtester.


askmike:
Finally! 🍕

PR #889, opened 2017-07-27
Corrected certbot installation command typo

tomvlk:
(missing install in apt-get command).


askmike:
Awesome thanks!

Quick question: what is your opinion on certbot (letsencrypt) VS self signed ssl certs?

tomvlk:
Certbot is quite nice, but requires to have open access and default ports. Eg. you can't run it behind firewall (without having to open it for a short period).
Btw, please take a look at the new nginx plugin for certbot, it has auto configuration stuff like it has for apache. See certbot site :)

askmike:
Will have a look!

PR #901, opened 2017-07-31
PR - Improving DOCS + telegram plugin

raffus:
- New documentation examples about creating strategies.
- Fixing telegram plugin so mobile devices can call the methods using upper or lower case (now its case-insensitive). When using mobile phone devices, the first input was always upper case and the robot could not identify the requested command. Now this has been fixed.

### DIFF
--- a/docs/strategies/creating_a_strategy.md
+++ b/docs/strategies/creating_a_strategy.md
@@ -66,6 +66,70 @@ If you find out in the check function that you want to give a new advice to the 
     this.advice('short');
     // or
     this.advice('long');
+    
+### candle variables when creating your strategy
+
+You can use different candle variables when writing your strategies, depending on what you have in mind, using additional variables might help you to improve your results.
+
+Possible candle variables are:
+ - candle.close: the closing price of the candle
+ - candle.high: the highest price of the candle
+ - candle.low: the lowest price of the candle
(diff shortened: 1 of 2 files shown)


docs/strategies/creating_a_strategy.md
```diff
@@ -66,6 +66,70 @@ If you find out in the check function that you want to give a new advice to the
...
+
+    // For debugging purposes.
+    strat.log = function() {
+      // your code!
+    }
+
+    // Based on the newly calculated
+    // information, check if we should
+    // update or not.
+    strat.check = function() {
```
askmike, line 113: this function is passed the candle, if you don't specify it as an argument you cannot access `candle.close` on the next line. Could you add that?

askmike:
Awesome, could you quickly fix the comment in the strategy example?

raffus:
Done! :)

2017-07-31 15:24 GMT-03:00 Mike van Rossum <notifications@github.com>:

askmike:
Great, thanks :)

raffus:
Hey Mike, wouldn't it be good to merge the docs into stable?
That might reduce some issues requests.

askmike:
It is already merged into stable, but as a rule all PRs (some hotfix
exceptions) need to to into develop first :)

PR #903, opened 2017-08-01
Updating DOCS broken link

raffus:
FIX in the broken link for the ta-lib indicators at the creating strategies doc page.


askmike:
Hey! Thanks for finding and fixes those. Two things about this PR:

1. Can you (instead of hard linking the URL) link like so `[talib docs](./name-of-file.md)`? This works in markdown docs, and when I render the website (gekko.wizb.it) those links will be converted to static links.
2. Can you remove the 3 commits that are already part of Gekko? Easiest would be by creating a new branch on Github and fixing the links there.

raffus:
Hi Mike,

I can do it, but, would you tell me how to do item 2?

Thanks!

2017-08-02 9:33 GMT-03:00 Mike van Rossum <notifications@github.com>:

askmike:
> would you tell me how to do item 2?

If you have the git repo on your own computer (and git installed) run:

    git checkout -b docs/fix-links
    # edit the files
    git commit -am 'updated broken links'
    git push

And then create a new PR from the new branch `docs/fix-links`. If you don't use git on your PC just push the requested changes here and I will make sure your commits will get added to Gekko.

raffus:
Mike,

I'm trying to work on this.

Please, help me out:

I'm editing file:
docs/strategies/talib_indicators.md

So, if I understood it correctly, I should change this:
`[your own trading methods](https://gekko.wizb.it/docs/strategies/creating_a_strategy.html)`

to this:
`[your own trading methods](./docs/strategies/creating_a_strategy.html)`

Is that correct?

Note 1:  I'm using .html extension, since the remote link has this one. Should I change it to .md?
Note 2: Should I use "." before "/"? Or just "/docs/strategies/creating_a_strategy.html" like a relative path?

Also, when editing the file "create_a_strategy.md"

We have:
`[TA-lib indicators](https://github.com/askmike/gekko/blob/stable/docs/trading_bot/talib_indicators.md
`
In this case, I was unable to find this file inside the project, just at the remote URL.

Thanks

askmike:
So the idea about the docs is that they also work if you download gekko on your computer and look into the docs folder (either as plain text or with a markdown editor).

> I'm using .html extension, since the remote link has this one. Should I change it to .md?

Please change it to .md! When the docs go on the website I rewrite all links from `.md` to `.html` already, see here: https://github.com/askmike/gekko.wizb.it/blob/master/md.js#L34-L35

> Should I use "." before "/"? Or just "/docs/strategies/creating_a_strategy.html" like a relative path?

Yes please use the relative link (with the `.`), that makes more sense when not using the website.

> We have: [TA-lib indicators (https://github.com/askmike/gekko/blob/stable/docs/trading_bot/talib_indicators.md
In this case, I was unable to find this file inside the project, just at the remote URL.

That doc has moved to: https://gekko.wizb.it/docs/strategies/talib_indicators.html (so you can still make it relative).

askmike:
Let me know if it is to much hassle, I can also do this myself. I just want to give others the change to contribute as well :)

raffus:
Ok, give me some time and I'll be happy to try and contribute. ;)

askmike:
Hey @raffus! Do you need more time or you need a push in the right direction?

askmike:
I think all of these changes have been merged into Gekko now :)

PR #915, opened 2017-08-04
Update documentation 

wakandan:
* add configuration section

### DIFF
--- a/docs/configuration/debug.md
+++ b/docs/configuration/debug.md
@@ -4,8 +4,17 @@ If you haven't done so already, first thing you will need to do is to copy `samp
 
 # Enable debug log in command line
 
-Set `config.debub = true;` (should already be the default setting) to see debug logs when using the command line
+To see debug logs when using the command line, set below setting in `config.js`
 
-Set `config.silent = false;` and `config.debug = true;` in `gekko/web/routes/baseConfig.js` to see debug logs when using ui mode
+    config.debub = true;
 
-With those settings enabled, you can log anything anywhere using `console.log('some message');`
+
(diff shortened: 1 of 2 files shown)


askmike:
Hey @wakandan so there are two ways of running Gekko:

- UI
- CLI

I want the UI to be the default, if users don't know Gekko and are looking for configuration options I want everything to be aimed at configuring Gekko for UI usage. So I don't want to add anything to a default configuration page that talks about `sample-config` and setting up that config.

The idea is: We need to keep everything as simple as possible for new users, and remove anything that might be unrelated and add confusion. The UI was designed to be this simple starting point so all current documentation (not found in the commandline folder) is aimed at this, as well as installation instructions (and docker files). You really only need to use the CLI for:

- Running a Gekko 24/7 for one of the plugins (IRC bot, telegram bot, etc).
- Developing new strategies (not necessary, but probably easier).
- Running a tradebot and requiring advanced logging.

I consider all these uses very advanced, so I don't want to throw configuration for that in people face when they have no experience with Gekko.

wakandan:
Ok I will update the document to something along the line "for starter, UI mode can be run without any extra configuration. However if you wish to see the logs in UI mode or take advantage of advanced features of the CLI mode, you can do so by modifying the config file". Sounds good?

askmike:
> Sounds good?

Hope you were not waiting for me!

Sounds great, but note the overlap between the configuration folder and the installation folder now, not sure what the difference is (since "installing gekko" is downloading files from github).

> UI mode can be run without any extra configuration

This is true, but the UI can also be configured (port, host, etc).

askmike:
Closing this for inactivity.

PR #917, opened 2017-08-04
Add `tradable` capability to Bitfinex trader

lipka:
Couldn't get the trader to work without this capability. Not sure if it was removed/not added intentionally?


askmike:
I think I forgot this, thanks! Will push into develop and test to make sure it works.

PR #923, opened 2017-08-05
Add EURETH and EURLTC pairs to GDAX exchange file

pelletencate:
GDAX now has EURETH and EURLTC markets that were previously not supported by Gekko. This PR will add support for them.


askmike:
Thanks!

PR #924, opened 2017-08-06
Update talib.js - fixed talib callback signature

metachris:
Fixed callback signature of talib (needs `err` and `result` parameters, see https://github.com/oransel/node-talib).

### DIFF
--- a/core/talib.js
+++ b/core/talib.js
@@ -7,9 +7,9 @@ var talibError = 'Gekko was unable to configure talib indicator:\n\t';
 var execute = function(callback, params) {
     return talib.execute(
         params,
-        function(result) {
-            if(result.error)
-                return callback(result.error);
+        function(err, result) {
+            if(err)
+                return callback(err);
 
             callback(null, result.result);
         }
--- a/core/talib.js
+++ b/core/talib.js
@@ -1,19 +1,25 @@
 var talib = require("talib");
+var semver = require("semver");
 var _ = require('lodash');
 
 var talibError = 'Gekko was unable to configure talib indicator:\n\t';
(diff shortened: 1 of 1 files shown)


metachris:
Not 100% sure you can merge it like this. 

* How did that work before // was this callback signature introduced in a newer version of talib? Then we should update package.json accordingly.
* Can result still contain the `error` property?

askmike:
Hey! The callback signature was changed as part in node-talib 1.0.3. If you manually install 1.0.2 this code will break (as well as everyone who has talib installed already and updates Gekko if this PR is in). I'll leave this PR open for a little bit until I've tested what happens when running this on talib 1.0.2.

metachris:
I see two ways of dealing with both the legacy talib and the new one:

1. Change this `execute` method to accept two generic arguments `arg1` and `arg2` and test them so it works with both the old and new version of talib
2. Have 2 different `execute` methods -- one for the old talib and one for the new?

What do you think?

askmike:
what about one proper execute method (the one in your PR), and one method that wraps the proper one and splits the error into it's own argument (in case of old talib verison)?

like so:

    const executeCallback = function(err, result) { /* version in your PR */. }
    const LegacyExecuteCallback = function(result) {
     var error = result.error;
     executeCallback.apply(this, [error, result]);
    }

And we can use the proper one by testing `talib.version` (in combination with semver for example)?
 What do you think?

metachris:
I think this is a good idea! I'll update the pull request.

askmike:
great thanks!

askmike:
I really like the style of composing (wrapping functions in functions), so that if we ever need to do some normalization of output (or input) or whatever, we only have to do that once.

Issue #759, opened 2017-05-27
RSI is wrong on tradebot


aneopsy:
calculated RSI properties for candle:
2017-05-27 21:43:06 (DEBUG):             rsi: 98.42998966176029

The RSI/TSI has a problem of range on tradebot is 98 to 100 in early.
EMA problem ?

Solved:
Create a TAlib version of RSI strat

darklow:
@aneopsy Could you please share your TAlib version of RSI strat?
Thanks.

talvasconcelos:
@darklow Talib doesn't install on windows, if it's your OS. Have the same issue. Don't know about the RSI problem though

askmike:
If anyone can let me know what is wrong with the RSI version I can fix that :)

andreialecu:
I have the same problem and there's nothing wrong with the RSI version, the problem is that there is no history being loaded when gekko first starts.

The history setting doesn't seem to work. For example with Bitstamp, it can load up to a day of history, but it doesn't actually make candles out of it. 

I've made a quick workaround for it in this commit: https://github.com/andreialecu/gekko/commit/04fa5758bde229009040b95538779e05ac8c0fb4, but I don't think it's the right fix. I need to get more familiar with the code before fixing it properly.

askmike:
> I have the same problem and there's nothing wrong with the RSI version, the problem is that there is no history being loaded when gekko first starts.

When the code hits [this path](https://github.com/andreialecu/gekko/commit/04fa5758bde229009040b95538779e05ac8c0fb4#diff-5f08f9bba973c819ee0352b5bf705834R171) it means that Gekko just called an exchange and it returned all data, which means that as soon as the pipeline starts the same data will be fetched and all data will be fed into the strat. If this did not work for you please open a new issue, because I think your commit will pump the same candles twice into the strat.

andreialecu:
While that may be true, without that particular hack the rsi always starts at 99+. At least on bitstamp. So the other code path is probably not working for some reason. I'll take another look at it tonight.

jjngpuri:
I have the same problem with RSI, any updates on this?

talvasconcelos:
rsi is calculated on live, at least is what I think.

Yroethiel:
[...]

Issue #911, opened 2017-08-02
v0.6.0 - Improve management and display of live gekkos


askmike:
Display to improve:

- Show current position (LONG or SHORT)
- Show full profit report table (same as a backtest)
- Show type of gekko (paper trader or real trader) (#1276)

Func to add:

- Stop button (option to also stop the market watcher) (#761)
- Auto restart on crash of market watcher
- "Auto Restart on Gekko restart" (#906)

Could haves:

- give live gekkos a name (#1033)

uknowsam:
Hi, Some isseus I will edit some more in the future if I see any.

Isseus:
-UI doesn't say API is wrong (CLI does)

tips:
- Showing if bot is working fine in UI or not
- Showing if bot is connected to exchange in UI or not

askmike:
@uknowsam good points, I will definitely make sure they make it into the next release.

stefangordon:
Eager to see the persistence of live gekkos (presumably what is meant by "auto restart").  I'm still new to this code base so I wouldn't be qualified to architect these features, but if there are any discrete tasks I can do to help with this issue send them my way.   (aka if you already know how you want something done but don't have time to do it.... :) )

krieghof:
Not only a name is usefull, we could also add field for notes. And the candle size, i always forget what time i set.

askmike:
Apologies for the delay, I was on a holiday.

@stefangordon

> Eager to see the persistence of live gekkos (presumably what is meant by "auto restart").

Yes that's what it means :) Everything you see in the app is a frontend app written in Vue, if you want to get your hands dirty see this document: https://gekko.wizb.it/docs/internals/gekko_ui.html

@krieghof 

> And the candle size, i always forget what time i set.

Yes this will definitely be there!

jlsalvador:
Hi!

Do you have a public branch with some WIP or POC? Maybe tomorrow I would spend some hours of development. A public branch can be awesome to prevent duplicate efforts.

Greetings!

askmike:
I have not started on this yet (just came back from a holiday), but before
we start coding I think it is was to start with a UI design of how we want
it to look!

jlsalvador:
Do you have some Design Lines or any idea (a POC is a good approach) around a first design? A hand drawing is perfectly viable for a first approach.
I'm not a good UX or UI developer, but I can give you some first steps as a Pull Request for simple ideas as a "Stop buttons".
[...]

PR #925, opened 2017-08-06
Add SMMA, replace RSI with original study

wuhkuh:
Part 1/2 of #759 

I'm going to split the pull request as this one is backwards compatible. The next part is not - it will replace the current RSI function with the one from the original study.

This means the patch version can be incremented for this specific PR, instead of a semver-minor.
Tests are included, however, it looked like the test suite is broken for this repository.
Anyway, I just added them in the same style as the tests for the other indicators.

The pre-calculated results are taken from my [own repository](https://github.com/wuhkuh/talib), which is documented and tested thoroughly. Sources used to determine the correct algorithm are documented, as well.

### DIFF
--- a/strategies/indicators/RSI.js
+++ b/strategies/indicators/RSI.js
@@ -1,23 +1,31 @@
 // required indicators
-var EMA = require('./EMA.js');
+var SMMA = require('./SMMA.js');
 
-var Indicator = function(settings) {
-  this.lastClose = 0;
+var Indicator = function (settings) {
+  this.lastClose = null;
   this.weight = settings.interval;
-  this.weightEma = 2 * this.weight - 1;
-  this.avgU = new EMA(this.weightEma);
-  this.avgD = new EMA(this.weightEma);
+  this.avgU = new SMMA(this.weight);
+  this.avgD = new SMMA(this.weight);
   this.u = 0;
   this.d = 0;
-  this.rs = 0;	
+  this.rs = 0;
   this.rsi = 0;
   this.age = 0;
 }
 
-Indicator.prototype.update = function(candle) {
+Indicator.prototype.update = function (candle) {
   var currentClose = candle.close;
 
-  if(currentClose > this.lastClose) {
+  if (this.lastClose === null) {
+    // Set initial price to prevent invalid change calculation
+    this.lastClose = currentClose;
+
+    // Do not calculate RSI for this reason - there's no change!
+    this.age++;
+    return;
+  }
+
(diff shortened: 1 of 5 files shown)


askmike:
Thanks a lot! Great stuff, let me know when you have both parts ready and I'll go over them.

wuhkuh:
Part 2/2 of #759

RSI has been replaced with the original study. Once again, tests are included and verified.
I haven't changed the interpretation of the RSI value, located under [strategies/RSI.js](https://github.com/askmike/gekko/tree/stable/strategies/RSI.js).

Keep in mind that this ~can~ will break the setup of the users.

I'd suggest doing a minor increment when implementing this, for this reason.

I assumed you'd want both parts together, so I put them together in this PR.
If not, I can split the pull request and send them in separately.

askmike:
This PR looks great, thanks a lot for the time you put in :)

> Keep in mind that this ~can~ will break the setup of the users.

What exactly will this break besides calculating different results?

wuhkuh:
> What exactly will this break besides calculating different results?

This is exactly it. I used the term break just to clarify that this change isn't a simple fix. I should've worded it differently, you're right.

Users of Gekko are probably used to working with the current results and I can imagine they wouldn't be too happy if we lower their revenue by implementing this change.

Assuming you're using Semver, if this version were to be 1.0.0 or higher, this change would classify as a semver-major - because we will change users results.
However, we aren't past 1.0.0 - so I'd suggest merging this whenever you're ready for 0.6.0.

askmike:
> so I'd suggest merging this whenever you're ready for 0.6.0.

Great, I will make it part of the release described in #911 and merge it into `develop` (the nightly branch now).

------

> Assuming you're using Semver, if this version were to be 1.0.0 or higher, this change would classify as a semver-major - because we will change users results.

Semver describes versioning according to API changes, but since Gekko is not a library that exposes a single API I found semver hard to follow:

- Gekko (CLI) can be configured using a config, the form of which has changed quite drastically.
- Gekko (UI) can also be configured (hosts, ports, etc), this config format has also changed.
- Gekko writes files to disk, very rarely I had to update this format (causing the need for users to migrate data)
- Gekko exposes a REST API (which is undergoing a lot of changes).
- The functionality available to strats (the "strat API") (which this PR does not change) changes sometimes.
- Functionality changes (new exchanges, new markets, new functionality).

For that reason I found it impossible to follow strict semver since all the above can be argued to be "APIs", and a lot of them have had breaking changes in the past. Also I am not comfortable labeling this v1+.

PR #926, opened 2017-08-06
Hide 80 port to resolve SSL issue

woogenius:
Problem
---
When I use gekko with SSL setup. `ERR_SSL_PROTOCOL_ERROR` has occurred. That's becuase SSL certificate cannot resolve `example.com:80`(only resolve `example.com`).

Solution
---
Hide port number when port is 80 on `api.js`


askmike:
Looks great, thanks!

PR #927, opened 2017-08-07
Fix #922 – GDAX improvements: round amounts to 8 decimals, have post_only default to true

### LINKED ISSUE #922: GDAX trade error: size too precise
I got this error as Gekko was trying to trade on GDAX with a size that has 17 decimals.

```
2017-08-05 18:07:41 (ERROR):    Error buying at GDAX: { message: 'size too precise (0.55431682711244915)' }
```
(numbers are changed for privacy reasons)

I made the following modifications
```
# exchanges/gdax.js
# [...]
    var buyParams = {
        'price': parseFloat(price).toPrecision(8),
        'size': parseFloat(amount).toPrecision(8),
        'product_id': this.pair,
        'post_only': this.post_only
    };

    var result = (err, data) => {
        if(err) {
            log.error('Error buying at GDAX:', err);
            return this.retry(this.buy, args);
        }
        callback(err, data.id);
    }

    this.gdax.buy(buyParams, this.normalizeResult(result));
}

Trader.prototype.sell = function(amount, price, callback) {
    var sellParams = {
        'price': parseFloat(price).toPrecision(8),
        'size': parseFloat(amount).toPrecision(8),
        'product_id': this.pair,
        'post_only': this.post_only
    };
    var result = function(err, data) {
        if(err) {
            log.error('Error selling at GDAX:', err);
            return this.retry(this.sell, args);
        }
        callback(err, data.id);
    }.bind(this);

    this.gdax.sell(sellParams, this.normalizeResult(result));
}
# [...]
```

Specificallly, in both price and size:

```
        'price': parseFloat(price).toPrecision(8),
        'size': parseFloat(amount).toPrecision(8),
```

Not sure if their API does allow 8 decimals, will report back when Gekko makes my first trade. May take a while, my strategy makes about one trade every 2 days.

pelletencate:
Amounts that had more than 8 decimals resulted in an API error `Size too precise`. They're now trimmed.

`post_only: false` was suggested elsewhere on the forum, but in my opinion it should really be true. If you set it to `false`, you should always deduct the taker fee from the price, or GDAX will reject any order. I believe GDAX rejects it if it's (partially) covered in the book as a taker, in which case Gekko should automatically recalculate the price and try to place the order again. I haven't tested this last situation, but IMO that's what should happen.

### DIFF
--- a/exchanges/gdax.js
+++ b/exchanges/gdax.js
@@ -9,7 +9,7 @@ var batchSize = 100;
 var Trader = function(config) {
     _.bindAll(this);
 
-    this.post_only = false;
+    this.post_only = true;
     this.use_sandbox = false;
     this.name = 'GDAX';
     this.import = false;
@@ -23,7 +23,7 @@ var Trader = function(config) {
         this.passphrase = config.passphrase;
 
         this.pair = [config.asset, config.currency].join('-').toUpperCase();
-        this.post_only = config.post_only ? config.post_only : false;
+        this.post_only = config.post_only ? config.post_only : true;
     }
 
     this.gdax_public = new Gdax.PublicClient(this.pair, this.use_sandbox ? 'https://api-public.sandbox.gdax.com' : undefined);
@@ -109,8 +109,8 @@ Trader.prototype.normalizeResult = callback => {
 Trader.prototype.buy = function(amount, price, callback) {
     var args = _.toArray(arguments);
     var buyParams = {
-        'price': price,
-        'size': amount,
+        'price': Math.floor(parseFloat(price) * 1e8) / 1e8,
+        'size': Math.floor(parseFloat(amount) * 1e8) / 1e8,
         'product_id': this.pair,
         'post_only': this.post_only
     };
@@ -129,8 +129,8 @@ Trader.prototype.buy = function(amount, price, callback) {
 Trader.prototype.sell = function(amount, price, callback) {
     var args = _.toArray(arguments);
     var sellParams = {
-        'price': price,
-        'size': amount,
+        'price': Math.floor(parseFloat(price) * 1e8) / 1e8,
+        'size': Math.floor(parseFloat(amount) * 1e8) / 1e8,
         'product_id': this.pair,
         'post_only': this.post_only
     };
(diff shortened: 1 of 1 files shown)


askmike:
> post_only: false was suggested elsewhere on the forum, but in my opinion it should really be true. 

Do note this situation should only happen when the top of the orderbook is updated between: `when Gekko fetches the orderbook --[ a few ms ]--> when Gekko places the order --[ a few seconds ]--> when the exchange executes the order`.

> I believe GDAX rejects it if it's (partially) covered in the book as a taker, in which case Gekko should automatically recalculate the price and try to place the order again.

Yes it should! Currently Gekko [assumes](https://github.com/askmike/gekko/blob/stable/plugins/trader/portfolioManager.js#L255) that the buy/sell order succeeds (because I only took fixable errors like connection issues into account: on connection error the exchange lib should just keep retrying). However retrying in this case doesn't work..

---

Though why don't we always assume the taker fee (for safety)? In most cases we will be maker but it doesn't matter that we don't trade the maximum available amount that much imo:

if you have 100 euro and Gekko is buying and selling that, if Gekko assumes a taker fee even though it's making the only downside is that Gekko is not swapping as much money as possible (if Gekko is buying you will still have some asset, like USD, in your account).

pelletencate:
> Though why don't we always assume the taker fee (for safety)? In most cases we will be maker but it doesn't matter that we don't trade the maximum available amount that much imo:

The reason, while not of the biggest relevance, is that you can't trade 100% of your balance, i.e. if I have € 100 that I trade on ETHEUR, the max order I can place is 99.70 EUR (ETH taker fees are .3%)

> Do note this situation should only happen when the top of the orderbook is updated between: when Gekko fetches the orderbook --[ a few ms ]--> when Gekko places the order --[ a few seconds ]--> when the exchange executes the order.

And that's why I don't think it matters too much, but I do agree with you that the retry should actually recalculate the entry price. Besides, this is the case, unless when it isn't. API could be slow, some people have slow or unreliable internet, who knows what. I know, as a trader you should take care of those, but still.) I haven't run into this situation yet, so I couldn't tell if it does or not, but if it doesn't – as in, if it just retries the exact same request – we may want to make it a bit smarter. 

Technology aside, as a trader I'd take a recalculated maker offer over an accidental taker offer due to latency any day of the week. And if it can propagate this particular error back to the portfolio manager saying "hey dude, refresh your thing, the order book has changed", I think that'd be ideal.

Pragmatically, of course sending in only 99.7% of the funds is the one-line immediate fix for this issue, but under the assumption that you can get a guaranteed maker trade, I'd consider that solution to introduce technical debt.

askmike:
> but I do agree with you that the retry should actually recalculate the entry price. Besides, this is the case, unless when it isn't. API could be slow, some people have slow or unreliable internet, who knows what. I know, as a trader you should take care of those, but still.) I haven't run into this situation yet, so I couldn't tell if it does or not, but if it doesn't – as in, if it just retries the exact same request – we may want to make it a bit smarter.

100% agreed. So just to be on the same page, there are two retries:

- When the trader created an order but it wasn't taken after 40 seconds (in full).
- When whilst adding an order it is refused (because we are suddenly taking because the OB shifted).

So to be sure: Gekko is already recalculating everything in the first case (including refetching balance & the order book). The second case is currently not handled at all, but yes: if we handle (by returning an `err` to the callback of the `buy` and `sell` methods) Gekko should indeed refetch balance & orderbook.

askmike:
> The reason, while not of the biggest relevance, is that you can't trade 100% of your balance, i.e. if I have € 100 that I trade on ETHEUR, the max order I can place is 99.70 EUR (ETH taker fees are .3%)

Yes, so in this case you cannot invest 30 cents but you lose the risk of having GDAX refuse the order since not enough balance (in case of a shifted OB).

PR #945, opened 2017-08-11
Fix GDAX fee for ETH and LTC

wingleung:
Fees for ETH and LTC should be 0.3% instead of 0.25% according to https://www.gdax.com/fees


wingleung:
Should probably merge #927 first

askmike:
awesome, thanks! They are both merged now.

Issue #946, opened 2017-08-11
Postgresql connection in Docker containers not working


cassvail:
I set up a docker-compose environment with gekko (original build image) and a postgresql-9.6 container.
The gekko container is linked to the postgresql container.
Edited the /web/vue/UIconfig.js to:
`adapter: 'postgresql'`
Edited the config.js file with the postgresql configuration:
```
config.adapter = 'postgresql';
...
config.postgresql = {
  path: 'plugins/postgresql',
  version: 0.1,
  connectionString: 'postgres://username:password@postgresql:5432', // if default port
  database: 'gek', // if set, we'll put all tables into a single database.
  schema: 'public',
  dependencies: [{
    module: 'pg',
    version: '6.1.0'
  }]
}
```
(Notice that @postgresql in docker is the linked host name)

The Web UI is running correctly but as soon as I interact with the database (i.e. clicking on "scan available data") I get this error:
```
Gekko encountered an error and can't continue

Error:

{ Error: connect ECONNREFUSED 127.0.0.1:5432
    at Object.exports._errnoException (util.js:1012:11)
    at exports._exceptionWithHostPort (util.js:1035:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1080:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5432 } '\n\n'

Meta debug info:

Gekko version: v0.5.2
Nodejs version: v6.3.1


npm info lifecycle gekko@0.5.2~start: Failed to exec start script
npm ERR! Linux 4.9.36-moby
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "start"
npm ERR! node v6.3.1
npm ERR! npm  v3.10.3
npm ERR! code ELIFECYCLE
npm ERR! gekko@0.5.2 start: `node ./gekko --config config.js --ui`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the gekko@0.5.2 start script 'node ./gekko --config config.js --ui'.
```

The connection host "127.0.0.1:5432" is obviously uncorrect, because it shoud use the Docker-compose link instead of localhost 127.0.0.1. 
Is this a bug or there's some other configuration needed in order to have Postgresql running?

cassvail:
Just added a PR #947 which adds postgresql to docker-compose config file. Could be useful to debug this issue and to have available in general

[...]

PR #947, opened 2017-08-11
docker-compose 'version 3' file with optional postgresql

cassvail:
Added optional configuration, compatible with latest docker versions, to run postgresql in docker-compose


askmike:
Will test it out tonight, will this also configure Gekko to actually use postgre as well?

cassvail:
This runs a postgresql database and links the gekko container to it. I opened an issue about connection configuration in the application #946

askmike:
great stuff! It's working smooth.

Thanks :)

PR #954, opened 2017-08-12
Feature/fix floating bug in gdax

### LINKED ISSUE #922: GDAX trade error: size too precise

wingleung:
Fixes #922


askmike:
Hey,

The function is now never passed the second argument (decimalLimit), but as soon as I do the output doesn't make sense anymore. Example:

     Trader.prototype.getMaxDecimalsNumber(132.123, 2); // "132.12299999"

if you need a dynamic amount of demicals, what is wrong with `toFixed()` (after a math.floor since to fix does a normal round)?

wingleung:
@askmike thanks for the review, indeed there was still a bug in other use cases. I fixed it in the following commit.

As for just using toFixed(), problem is setting toFixed(decimalLimit) will round up/down if the number has more decimals already. Math.floor() will only work if your multiplier is also dynamic, which I did do in the last commit. So it should work now...
`
 Trader.prototype.getMaxDecimalsNumber(132.123456789); // "132.12345678"
`
`
 Trader.prototype.getMaxDecimalsNumber(132.129, 2); // "132.12"
`
`
 Trader.prototype.getMaxDecimalsNumber(132.123, 2); // "132.12"
`

So to solve this... 
- we make the float number an integer by multiplying it (1e8 for 8 decimals, 1e2 for 2 decimals,...) so that the Math.floor will do its rounding down.
- divide by the same multiplier so it's a float number again
- toFixed(decimalLimit), to cut away all unnecessary decimals

Also fixed a bug where the config.post_only property would not set false when set to false.

askmike:
Looks great! Let's hope this gets rid of the error :)

askmike:
@damianleung it is still not working unfortunately, see #992.

PR #958, opened 2017-08-13
Fix typos in creating_a_strategy.md

discosultan:


askmike:
awesome stuff, thanks!

Also if that doc doesn't make a lot of sense let me know :)

PR #985, opened 2017-08-16
split fees taker & maker for paper trader

julesGoullee:

Many exchanges use different fees for make and take orders, it's useful in paper trader for backtesting to implement this  

* **Other information**:


askmike:
Thanks! I like the idea a lot.

But just to be clear: You are not always a taker when you buy or a maker when you sell. You are a maker when you add your order to the orderbook (which the real trader of Gekko always tries to do), and you are a taker when you take orders from the book. Making = limit orders, taking = market orders.

So maybe we could add an option in the paperTrader config that specifies whether the paperTrader should simulate using maker or taker fees?

julesGoullee:
Ha... Thank's to you for the clarifying, option has a good idea, I will work on it tomorrow, sleep now !

askmike:
Awesome stuff! Thanks

Wil test & merge soon.

Issue #891, opened 2017-07-28
Populate candleProps for non Talib indicators/strategies


thegamecat:
From playing around I've noticed that candleProps doesn't get set unless you add a Talib indicator to your strategy. From testing this is the line that does this contained in baseTradingMethod:

 if(this.asyncTick) {
    this.candleProps.open.push(candle.open);
    this.candleProps.high.push(candle.high);
    this.candleProps.low.push(candle.low);
    this.candleProps.close.push(candle.close);
    this.candleProps.volume.push(candle.volume);

    if(this.age > this.candlePropsCacheSize) {
      this.candleProps.open.shift();
      this.candleProps.high.shift();
      this.candleProps.low.shift();
      this.candleProps.close.shift();
      this.candleProps.volume.shift();
    }
}

If I rem out the  if(this.asyncTick) { then candleprops is set and usable to my non Talib indicators.

Can I be confident that the values set are correct and not jumbled due to async? (ie what is the purpose of the if statement?)

askmike:
> If I rem out the if(this.asyncTick) { then candleprops is set and usable to my non Talib indicators.

Except that non talib indicators get price updates directly from the candle. So `this.candleProps` is only used for talib indicators and for nothing else.

---

The baseTradingMethod is a bit of a mess at the moment, if you have any questions about how it works (or is supposed to anyway) feel free to ask here!

askmike:
The indicators get updates every candle here: https://github.com/askmike/gekko/blob/stable/plugins/tradingAdvisor/baseTradingMethod.js#L166-L173

thegamecat:
> Except that non talib indicators get price updates directly from the candle. So this.candleProps is only used for talib indicators and for nothing else.

Yes, but then I have to create and manage the arrays in the strategy - this way I always have candleProps to reference which is preferable to me as there is one version of the truth for all indicators then.

askmike:
These props were not meant to be exposed (and used in strategies), you can definitely uncomment them if you want though :)

askmike:
What about an option in the `tradingAdvisor` that populates these arrays regardless of ASYNC candles?

thegamecat:
Not sure what you mean - so long as the candles are populated I don't think it matters much.

askmike:
I know, but I don't want to always populate them. They are only needed for TAlib indicators or for you. They will be populated when there are TAlib indicators, so we could add a config switch so they will be populated for you.

thegamecat:
[...]

PR #993, opened 2017-08-17
add in candle vwap et nb trades

julesGoullee:
Everyone have their strategies, vwap and counter trades of candles are usefully !


askmike:
Hey! Nice that you are going through that (messy & undocumented code), but the `this.candleProps` is only used to properly populate TAlib indicators. And some of them are using volume (you just fixed a bug!), but none are using `vwp` or `trades`. So there is no point in populating and maintaining these arrays.

askmike:
Here you can find all TAlib indicator implementations: https://github.com/askmike/gekko/blob/1c0700d312e3f6bc6e4fb9f4d1fdcda5d3392736/core/talib.js

----

Would you mind changing this to just extend the volume array, but not the other ones?

julesGoullee:
Not very messy ;),I've already find /core/talib.js, In substance I added method for use async strategy (ipc) and I need this props. I think the messy code is in baseTradingMethod, in my opinion it was the simple way for add custom async strategy, if you have a better idea I take it !

askmike:
Fair enough, the candleProps arrays are only a 1000 elements anyway :)

julesGoullee:
Yep, I show this, but why is not the warmup props length, I'm not understand

askmike:
Good question actually! I have a *feeling* that people will try to get away with the lowest historySize (warmup period) as possible to not have to wait as long. But the overhead of storing a few more array items is pretty small. But the 1000 is quite randomly chosen by me, you are right..

thegamecat:
This is related to: https://github.com/askmike/gekko/issues/891

askmike:
@thegamecat fair point, I have responded there.

thegamecat:
Ok so as far as design decisions go it seems clear there are people who want access to the candle data. We can agree on that for sure.

Why don't you want people accessing the candleprops data? Or perhaps rather, what is your preferred method of giving access to that data?

askmike:
No: I am proposing a config flag that only exposes the candleProps to the
people who specifically want it, and not for everyone. As far as I know
there are 2 people who want this specific data and 4000 people who
installed gekko in the last month. No reason to decrease performance (speed
and memory usage) for all these people while we can make everyone
(including you) happy with a simple config switch.

thegamecat:
Oh I see. There might be another way then, because in the UI you're writing all this data again to build the graphs. Maybe that and candleprops can be unified?

Just for the record the speed and memory impact is absolutely tiny. I did however notice that writing the candles and roundtrip to files does slow things down, particularly the candles (as one would expect).

askmike:
>  because in the UI you're writing all this data again to build the graphs. Maybe that and candleprops can be unified?

The graphs only use the close of the candles no other properties. Also this is completely decoupled from a backtest for a few different reasons (including a bug you encountered where the backtest output was to big).

LooOOooM:
great guy, and where is finally the swtich -- I upgraded gekko after about 6 -8 month and finally my strat which used candleProps is not working anymore as it tells me undefinded.

- how can I enable candleProbs again?

medozero:
i need access to the previous candle(s) :) as well . so mike make them 4 ppl who will be happy including LooOOooM

IAMtheIAM:
I also need to access previous candles, where is the flag to enable `candleProps` or how can I expose it manually to my strategy?

UPDATE: I figured it out.
Here's how you do it, as long as you are using a TALib indicator:

`let candleProps = this.asyncIndicatorRunner.candleProps`

This is quoted from [an article](http://gekkowarez.com/tag/candleprops/) which helped me:

> If you want to use the candleprops array in a non talib using Strategy then do the following:
> 
> open plugins/baseTradingMethods.js and commend out the line below and the corresponding close bracket:
> 
> // if(this.asyncTick) {
> 
> Then go to your strategy and write console.log(this); in the update method and look for the candleprops arrays.
> 
> Note: baseTradingMethod is a core Gekko file so you should make this change only if you’re comfortable managing this change yourself.

askmike:
@IAMtheIAM the data behind those properties is only collected when you add certain indicators. The proper way is to store them yourself in your strategy.

PR #995, opened 2017-08-17
Develop

thegamecat:
Adds a timeout(0) to server.js

There is no timeout and multiple requests to the API causes an ECONNRESET.

Adding server.setTimeout(0) to line 79 which puts a timeout with no expiry. 

* **Other information**:
This bug is only really triggered by multiple requests hitting the api in parallel. The fix seems to be a common one when when using request / posts and receiving ECONNRESET errors randomly.


thegamecat:
This isn't what I wanted to do....sorry I don't know what I'm doing.

askmike:
The easiest way to add a PR like this is go browse to the file on github (on my repo - on the develop branch) and click on edit in the rop right :)

thegamecat:
haha ok...on it!

PR #1001, opened 2017-08-18
Create DMI.js

thegamecat:
Proposing this new DMI indicator. Needs to be tested by people who know more about how DMI works than I do. It should at the very least output ADX and MinusDI.

* **Other information**:

### DIFF
--- a/strategies/indicators/DMI.js
+++ b/strategies/indicators/DMI.js
@@ -1,0 +1,156 @@
+// required indicators
+var log = require('../../core/log');
+
+var EMA = require('./EMA.js');
+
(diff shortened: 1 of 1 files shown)


askmike:
Great stuff! As discussed on discord I can only accept this after unit tests that validate the calculations.

thegamecat:
Yep and hopefully someone will want to take a look at that.

askmike:
Closing this for inactivity. I unfortunately won't be adding indicators without tests as for all the support I will have to provide in the future (this has happened a few times in the past now).

If someone adds tests we can reopen this :)

Issue #992, opened 2017-08-17
GDAX - "Price too precise" on latest `stable`


GuyPaddock:
* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Left a MACD BTC/USD trading strategy running on GDAX throughout the day via CLI.

* **Expected result** (what you hoped would happen)
BUY and SELL orders would be processed without errors.

* **Actual result** (unexpected outcome)
BUY orders that are too precise fail, as follows:
```
2017-08-17 10:04:04 (DEBUG):    Processing 8 new trades. From 2017-08-17 14:03:50 UTC to 2017-08-17 14:04:01 UTC. (a few seconds)
2017-08-17 10:04:04 (DEBUG):    calculated MACD properties for candle:
2017-08-17 10:04:04 (DEBUG):             short: 4449.26238286
2017-08-17 10:04:04 (DEBUG):             long: 4431.91163462
2017-08-17 10:04:04 (DEBUG):             macd: 17.35074824
2017-08-17 10:04:04 (DEBUG):             signal: 17.08544991
2017-08-17 10:04:04 (DEBUG):             macdiff: 0.26529833
2017-08-17 10:04:04 (DEBUG):    In uptrend since 1 candle(s)

2017-08-17 10:04:04 (INFO):     We have new trading advice!
2017-08-17 10:04:04 (INFO):              Position: long
2017-08-17 10:04:04 (INFO):              Market price: 4467.99
2017-08-17 10:04:04 (INFO):              Based on market time: 2017-08-17 14:03:00

2017-08-17 10:04:04 (INFO):     Trader Received advice to go long. Buying  BTC
2017-08-17 10:04:04 (INFO):     Attempting to BUY 0.02090085414447822 BTC at GDAX price: 4467.97999999
2017-08-17 10:04:05 (ERROR):    Error buying at GDAX: { message: 'price too precise (4467.97999999)' }
2017-08-17 10:04:05 (DEBUG):    GDAX returned an error, retrying..
2017-08-17 10:04:16 (ERROR):    Error buying at GDAX: { message: 'price too precise (4467.97999999)' }
2017-08-17 10:04:16 (DEBUG):    GDAX returned an error, retrying..
2017-08-17 10:04:24 (DEBUG):    Requested BTC/USD trade data from GDAX ...
[...]

PR #1002, opened 2017-08-18
Force 2 decimals with any price / Fix deductFee function

wingleung:
Bugfix

- GDAX: price trim to 8 decimals
- GDAX: amount after deductFee() is more than 8 decimals

- GDAX: price always trims to 2 decimals
- GDAX: amount after deductFee() is always 8 decimals or less

* **Other information**:
Quick bugfix, should be tested with actual scenario. Currently running on ETH-EUR but I won't have time to be behind a computer the next couple of days.


exchanges/gdax.js
```diff
@@ -307,6 +307,22 @@ Trader.prototype.getMaxDecimalsNumber = function (number, decimalLimit = 8) {
   return decimalCount <= decimalLimit ? decimalNumber.toString() : (Math.floor(decimalNumber * decimalMultiplier) / decimalMultiplier).toFixed(decimalLimit);
 };
 
+Trader.prototype.deductFee = function (number) {
```
wingleung: @askmike not sure if it's necessary, but I think I had to deduct the Fee here to prevent the `price too precise` error. Do you know when the fee (should) be deducted? is it on the GDAX side? Or somewhere in Gekko?

exchanges/gdax.js
```diff
@@ -307,10 +307,20 @@ Trader.prototype.getMaxDecimalsNumber = function (number, decimalLimit = 8) {
   return decimalCount <= decimalLimit ? decimalNumber.toString() : (Math.floor(decimalNumber * decimalMultiplier) / decimalMultiplier).toFixed(decimalLimit);
 };
 
-Trader.prototype.deductFee = function (number, fee = 0.003) {
-    var resultNumber = parseFloat(number) * (1 - fee);
+Trader.prototype.deductFee = function (number) {
+    var _this = this;
```
askmike: nitpick: you don't need to do this if you are creating a function using the fat arrow, that wil automatically bind context.

askmike:
This `deductFee` function will return another function, so it would set the `size` prop to a function and not the result:

https://github.com/damianleung/gekko/blob/b7f1662a13f18835d03b648ea0e7af138bec3a5f/exchanges/gdax.js#L116

Though first: what bug is this supposed to fix? I am testing GDAX now and I think it's pretty stable.

wingleung:
Removed `deductFee()`, it should work without it

so undoing https://github.com/askmike/gekko/pull/1002#discussion_r134040005

askmike:
My experience with GDAX is limited but I thought:

- you can specify amount up to 8 decimals.
- you can specify price up to 2 decimals (or 5 for crypto markets).

We know the price can't be 8 from the error in #992 (8 decimals were specified but still an error). Looking at the orderbook ([here for USD](https://api.gdax.com/products/BTC-USD/book) and [here for LTC](https://api.gdax.com/products/LTC-BTC/book)) I see amounts specified in 5 and 2 decmials.

wingleung:
@askmike ok I see, I'll change it to 5 decimals then. I only see BTC as currency on this table https://www.gdax.com/fees/BTC-USD so the currency check is only on BTC.

askmike:
@damianleung thanks a lot for bearing with me! I'll merge this right now in the hope it fixes #992. I'll merge manually to remove some of the commits.

askmike:
Merged in these commits:

- https://github.com/askmike/gekko/commit/0e8f2b10115eb70fb56aca338f96a388504963ac
- https://github.com/askmike/gekko/commit/d2bf48eb947dcede4e1af1260a8ba1d2dfc48f2b

askmike:
Thanks :)

PR #1015, opened 2017-08-21
update kraken & poloniex markets

askmike:


askmike:
Great stuff, thanks @decoye

Issue #645, opened 2017-03-23
Backtest / API  tradingAdvisor issue


thegamecat:
In the dev branch I now need to pass through the tradingAdvisor config in a way that was not previously necessary (good).

The issue is if I pass it through as: 
tradingAdvisor:
{ enabled: true,
candleSize: 60,
historySize:30,
method: 'talib-dmi-vix'
}

I get an error on the server side saying; candleSize is not a number.

Now the value is clearly coming through to the candlebatcher but it's not a number. 

This a bug? If I remove the error catchers in candlebatcher then I get the same error on optintimeperiod....

Any idea?

askmike:
> the server side

Could you give more information like:

- how exactly are you starting the backtest?
- what is the exact input (in json)?

thegamecat:
async.mapSeries(tests_series, function (data, callback) {
        var outconfig = {
            tradingAdvisor: {
enabled: true,
method: 'talib-dmi-vix',
candleSize: 60,
historySize: 30,
adapter: 'sqlite',
talib: {
enabled: true,
version: '1.0.2'
}
},
            watch: {exchange: 'poloniex', currency: 'BTC', asset: 'BELA'},
            profitSimulator: {
                fee: 0.25,
                slippage: 0.05,
                simulationBalance: {asset: 1, currency: 100},
                enabled: true
            },
            'talib-dmi-vix': {
                parameters: {optInTimePeriod: data.optInTimePeriod},
                adxparameters: {optInTimePeriod: data.adxOptInTimePeriod},
                thresholds: {
                    adxthreshold: data.adxthreshold,
                    minusgap: data.minusgap,
                    plusgap: data.plusgap,
                    vixperiod: data.vixPeriod,
                    upperwvf: data.upperwvf,
                    lowerwvf: data.lowerwvf,
                    t1: data.t1,
                    t2: data.t2
                }
            },
            valid: true
        };
        request.post({
                url: 'http://localhost:3000/api/backtest',
                form: outconfig
            },
            function (err, httpResponse, body) {
                jsongap = JSON.parse(body);
                results.push(jsongap.report.profit);
                return callback(null, body);
            });
    }, function () {
        callback(results)
    });

thegamecat:
Basically if i leave out the trading advisor part then there is no error. 
But of course it doesn't actually run the backtest with the values / strategy.

askmike:
Would you mind putting that code in a gist / pastebin? This is unreadable.

[...]

PR #1078, opened 2017-09-04
Batch the backtest with variation in parameters, result in CSV

pprados:
This extension launch many times a back test, with different parameters.
All the results are added in a CSV file. It's easy to try different parameters values for the same data, or
to check a strategy in a sliding windows.

See `config.lab` in `sample-config.js`
To start, check the backtest
`node gekko --config <...> --backtest`
then use 
`node labgekko --config <...> --backtest`.
```
config.lab={
  enabled:true,
  file:"labs/"+config.tradingAdvisor.method+".csv",
  // Fields to add in CSV. Can be omitted or reorders
  fields:[
    "method", "exchange","currency","asset",
    "candleSize","historySize",
    "patchs",
    "startTime","endTime","timespan",
    "startPrice","endPrice","market","trades",
    "startBalance","balance","profit",
    "relativeProfit", "relativeResult"
  ],
  // Lists of modification to apply in config.
  patchs:[
    { 
      path: "MACD.short",
      loop: ["5", "<= 15", "+ 5"], // Normal:10
    },
    { 
      path: "MACD.long",
      loop: ["16", "<= 26", "+ 5"], // Normal:21
    },
    { 
      path: "MACD.signal",
      loop: ["8", "<= 10", "+ 1"], // Normal:9
    },
    {
      path:"backtest.daterange",
      exloop:[
        'config.{}={from:"2016-01-01", to: "2016-04-01"}',
        'moment(config.{}.to).isBefore("2017-12-15")',
        'slidingWindow(config.{},1, "month")',
        '// require nothing in global scope',
      ],
    },
  ],
}
```

* **Other information**:
It's possible to add a new mode and remove the specific `labgekko`.

### DIFF
--- a/labgekko.js
+++ b/labgekko.js
@@ -324,6 +324,16 @@ function start() {
 
 }
 
+// helper to store the evenutally detected
+// daterange.
+var setDateRange = function(from, to) {
+  config.backtest.daterange = {
+    from: moment.unix(from).utc().format(),
+    to: moment.unix(to).utc().format(),
+  };
+  util.setConfig(config);
+}
+
 if (config.backtest.daterange === 'scan') {
   scan((err, ranges) => {
     if(_.size(ranges) === 0)
--- a/labgekko.js
+++ b/labgekko.js
@@ -269,7 +269,8 @@ function newSimulation(cb) {
     function(report) {
       saveReport(report);
       var result=calulateResult(report);
-      log.info("Result="+result.toFixed(0)+" ("+((result>0) ? "Good deal !" :"Better than the market if positif")+")");
+      log.info();
+      log.info("Difference with marker ="+result.toFixed(0)+" ("+((result>0) ? "Good deal !" :"")+")");
       cb();
     });
 }
@@ -289,6 +290,7 @@ function printPatchedConfig() {
   for (var i=0;i<config.lab.values.length;++i) {
     log.info(config.lab.values[i].path+" = "+JSON.stringify(config.lab.values[i].value))
   }
+  log.info();
 }
 
 // -------------- Start the experience
--- a/labgekko.js
+++ b/labgekko.js
@@ -1,0 +1,352 @@
+console.log(`
+***********************
+****** LAB GEKKO ******
+***********************
+`);
+
+const _ = require('lodash');
+const fs = require('fs');
+const path = require('path');
+const util = require(__dirname + '/core/util');
+const moment = require('moment');
+const promisify = require('tiny-promisify');
+const pipelineRunner = promisify(require(__dirname + '/core/workers/pipeline/parent'));
+const log = require(__dirname + '/core/log');
+var scan = require(__dirname + '/core/tools/dateRangeScanner');
+
+const config = util.getConfig();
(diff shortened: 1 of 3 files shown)


askmike:
Good stuff, so the idea is that you have to specify all iterations you want to run in a config like so:

    patchs:[
      { 
        path: "MACD.short",
        loop: ["5", "<= 15", "+ 5"], // Normal:10
      },
      { 
        path: "MACD.long",
        loop: ["16", "<= 26", "+ 5"], // Normal:21
      },
      { 
        path: "MACD.signal",
        loop: ["8", "<= 10", "+ 1"], // Normal:9
      },
      {
        path:"backtest.daterange",
        exloop:[
          'config.{}={from:"2016-01-01", to: "2016-04-01"}',
          'moment(config.{}.to).isBefore("2017-12-15")',
          'slidingWindow(config.{},1, "month")',
          '// require nothing in global scope',
        ],
      },
    ],

Since 0.4 Gekko can act as a server with a REST api: you can start a the server (automatically started when you run the UI) and run backtests against it by passing a config, the server will return the result of the backtest.

What would you think of putting this functionality in a small standalone script that just calls the rest API of Gekko? :)

That way people how have different ideas about how to try out different parameters can do it differently, and we don't have to have big discussions about it in Gekko itself, see #767 for an example.

decoye:
@askmike would you mind to share information on how to use the api to pass information to it?

askmike:
@decoye #645 has an (outdated) example :)

pprados:
Hello,

I think it's a good idea. Now, the code juste start a fork. But, it's possible to invoke the API.
Where is the version 0.4 ? Which tag ?

imkane:
Neat stuff! This is what I had in mind awhile back but I wasn't smart enough to code it lol.

imkane:
@pprados - I'm trying to figure out the purpose of the relativeProfit and relativeResult cols (which are always identical). They don't seem to match any output from the UI using the same test params. Any advice?

For example, in the CSV file:

method | RSI.interval | RSI.thresholds.low | RSI.thresholds.high | RSI.thresholds.persistence | market | trades | relativeProfit | relativeResult
RSI | 2 | 25 | 55 | 1 | -0.0169 | 98 | 0.2632 | 0.2632

And in the UI:

screenshot - 9_4_2017 10_02_08 am

(Notice that the # of trades don't even match)

pprados:
Hello,
Check the candle size and history. Compare with the backTest.

The relativeProfit is just the result of the calculate report.
The relativeResult must be fixed to (balance-startBalance)-(endPrice-startPrice) to compare the market and the strategy. You must initialise the asset to zero.

imkane:
@Pascal66 Yes, I've checked candle size and history. I still can't get the lab results, even just for "market" to match what the UI outputs. Weird.

Are you able to get your lab results to match your UI?

imkane:
Also, how is it possible that "market" changes from test to test?

screenshot - 9_5_2017 11_35_30 am

pprados:
For the "market change", I propose to use my another pull-request with unit test of strategies.

Issue #1081, opened 2017-09-04
Kraken buy order precision


lbnt:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[ X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Ask gekko to trade on kraken

* **Expected result** (what you hoped would happen)
that gekko can buy!

* **Actual result** (unexpected outcome)
Error of kraken api call due to precision problem

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

This behaviour was explained by kraken a few days ago
https://blog.kraken.com/post/1261/announcement-reducing-price-precision-on-all-trading-pairs/

2017-09-05 00:00:47 (DEBUG):	kraken returned an error, retrying..
2017-09-05 00:00:52 (DEBUG):	BUY 0.359568 ETH @ 273.60199999 EUR
2017-09-05 00:00:53 (ERROR):	unable to buy Error: Order:Invalid price:XETHZEUR price can only be specified up to 3 decimals.
    at /home/XXX/bin/gekko/node_modules/kraken-api-es5/index.js:59:13
    at next (native)
    at step (/home/XXX/bin/gekko/node_modules/kraken-api-es5/index.js:1:253)
    at /home/XXX/bin/gekko/node_modules/kraken-api-es5/index.js:1:423
    at process._tickCallback (internal/process/next_tick.js:109:7)

askmike:
Ah yes, Kraken has changed a few days back. Will update!

julesGoullee:
I'm working on

askmike:
@lbnt can you try out the latest develop branch of Gekko? You can download it here:
https://github.com/askmike/gekko/tree/develop

It includes a fix for this issue by @julesGoullee!

julesGoullee:
I'm currently trying multiple pairs too

flux154:
I keep getting:

 TypeError: Cannot read property 'log' of undefined
    at pipelineRunner (C:\xxxx\xx\Downloads\gekko-develop\web\routes\startGekko
.js:86:14)
    at Object.message (C:\xxxx\xx\Downloads\gekko-develop\core\workers\pipeline
\messageHandlers\realtimeHandler.js:10:9)
[...]

PR #1083, opened 2017-09-05
kraken reduce precision

julesGoullee:
Apply specifique precision by pair, is not change (see https://blog.kraken.com/post/1278/announcement-reducing-price-precision-round-2/) conserve 8 digit by default 
[See issue](https://github.com/askmike/gekko/issues/1081)


askmike:
Great stuff, thanks a lot!

PR #1085, opened 2017-09-05
Inject config into his strategy

oktapodia:
Feature: To avoid injecting all the config file and then filter it by the strategy name, I did a little change to inject it directly through the Base strategy. This modification permits you to access to your `settings` configuration and also the `tradingAdvisor` configuration.

Today, we have to do 
```
var config = require('../core/util.js').getConfig();
var settings = config.UO;
```

With this PR, we could do 
```
method.init = function() {
  this.settings // Get the strategy settings
  this.tradingAdvisor // Get the trading advisor 
}
```

* **Other information**:
How do I do the test? I did not see any tests for the strategies, do you have some advice because I'm not totally confident with the library

### DIFF
--- a/docs/strategies/creating_a_strategy.md
+++ b/docs/strategies/creating_a_strategy.md
@@ -187,12 +187,10 @@ You can create configurable parameters for your method which allows you to adjus
 
 And in your method you can use them again (for example to pass to an indicator):
 
-    // in the init:
-    var config = require('../core/util.js').getConfig();
-    this.settings = config.custom;
-
     // anywhere in your code:
     this.settings.my_custom_setting; // is now 10
+
+___The name of your configuration must be the same as the name of the strategy___
 
 ### Tool libraries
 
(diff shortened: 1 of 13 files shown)


askmike:
:o very nice, great stuff!

Would you mind updating the documentation here: https://github.com/askmike/gekko/blob/develop/docs/strategies/creating_a_strategy.md#configurables

Thanks!

oktapodia:
@askmike Updated, and how about the testing?

askmike:
Unit tests? Not the full codebase is under unit test at the moment, mainly indicator calculations and core components. This logic you changed was not under unit test right now, you are free to write unit tests for them if you want.

I'll do some manual testing myself before I merge (ASAP).

Thanks a lot already!

oktapodia:
Hum ok I'll look into that after this PR :D

askmike:
Thanks a lot :)

PR #1092, opened 2017-09-07
color coded profit & loss in roundtrip table

bbjay:
better readability of P&L and Profit column

right-aligned, red for losses, green for profits, added % sign, limit to 2 fraction digits.
also removed the -(minus) sign for losses, since a negative loss would actually be a profit,
not sure if this appeals to everyone though.

here's how it looks now:
bildschirmfoto 2017-09-07 um 21 02 47

* **Other information**:


askmike:
Great stuff! Thanks :)

PR #1100, opened 2017-09-09
Fix a typo in the parameter list for the example UO strategy

stackedsax:
Bug fix

The example UO strategy doesn't really work because there is a typo.  "hight" should be "high".

I removed the extra "t" so that people would be able to play around with the example settings like normal.

* **Other information**:


askmike:
Nice find, thanks!

PR #1101, opened 2017-09-09
Fixed a bug in talib.js

cbwang2016:
in the ULTOSC method

* **Other information**:

### DIFF
--- a/core/talib.js
+++ b/core/talib.js
@@ -1273,7 +1273,7 @@ methods.typprice = {
 }
 
 methods.ultosc = {
-    requires: ['optInTimePeriod'],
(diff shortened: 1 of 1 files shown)


askmike:
nice find!

cbwang2016:
@askmike BTW, you also need to update the website: https://gekko.wizb.it/docs/strategies/talib_indicators.html#ultosc

askmike:
I'll try to get around that soon. Note that the website simply renders everything inside the `docs` folder, so if you want you can submit a PR to that specific doc as well: https://github.com/askmike/gekko/blob/develop/docs/strategies/talib_indicators.md

PR #1104, opened 2017-09-09
[Bugfix] talib macdfix option does not match the requires one

oktapodia:
Bugfix

Talib MACDFIX does not work because of the requires option (`SignalPeriod`) does not match the used option (`optInSignalPeriod`)

Fix the bug

* **Other information**:

### DIFF
--- a/core/talib.js
+++ b/core/talib.js
@@ -588,7 +588,7 @@ methods.macdext = {
 }
 
 methods.macdfix = {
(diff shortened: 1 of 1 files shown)


oktapodia:
https://github.com/askmike/gekko/issues/1105

askmike:
Cool stuff, thanks!

PR #1124, opened 2017-09-18
Supporting import and history from Kraken

cmroche:
* **What kind of change does this PR introduce?** (**Bug fix**, **feature**, **docs update**, ...)

The change brings the following:

(feature) Supporting import from Kraken, and update documentation to this effect
(bugfix/feature) Supporting history from the Kraken exchange
(bugfix) Adding BCH to the supported crypto asset list in the exchange module
(bugfix) Adding ETH to the backing currency list in the exchange module
(bugfix) Adding other special case currency identifiers to the "no prefix" list

Could not import from Kraken
Could not properly trade some alt-coins with Kraken (EOS, DASH or GNO)
Could not import history for ETH backed pairs (though needed importing to be a problem)

You can now import the historical data from kraken supported currency pairs for backtesting, and should have no issues with live trading any supported pair for BCH, EOS, DASH or GNO.

* **Other information**:

There are a lot of undocumented quirks with Kraken's API unfortunately, probably the reason why it was difficult to get historical trades in the past, I've added a couple of notes in the exchange code to this effect.

One such quirk is that Kraken appears to use a very precise timer for generating TIDs but does not expose that TID on historical trade data to the same degree of precision. For this reason requesting new timeframes of data can cause an overlap in the trades returned so a filter needs to be used to prevent re-inserting the same trade twice.

I've tested importing of 5-12 months of data for several pairs, including ETH-XBT, ETH-USD, XBT-BCH, XBT-EOS, USD-BCH, and live market watches for ETH-CAD and ETH-USD


askmike:
Awesome stuff, will test! As for:

> For this reason requesting new timeframes of data can cause an overlap in the trades returned so a filter needs to be used to prevent re-inserting the same trade twice.

This overlap should not be a problem for live gekkos (nor for importers), since the tradeBatcher is already filtering out trades that are double: https://github.com/askmike/gekko/blob/stable/core/budfox/tradeBatcher.js#L95

cmroche:
So after looking a bit more carefully at live trading, which I am unable to test right now, I'm pretty certain it is broken for BCH, EOS, GNO and DASH since they do not properly handle the prefix situation in commands around handling orders.

I also made a small error in the fix to getPortfolio. In the case of the non-prefixed currency pairs the API request sends the pair in the follow format `XBTBCH` but will receive balances in a format similar to this `{ [XXBT] => 1.0, [BCH] => 1.0 }` where the currency is still prefixed, but the asset is not.

Since calls to get trade and importing work, and this was the intent of this pull request I've undone the change to getPortfolio. I'll issue a new pull request later on today to address specifically the outstanding live trading issues on those asset pairs.

cmroche:
@askmike About the overlapping trade results. Normally my position on this is that someone calling the API with an incrementing trade id wouldn't expect the same results to be returned twice so I approach it from that perspective. If you think it is safe I will undo the filter, but do you think this could lead to unexpected side effects down the road?

cmroche:
I fixed a small error I made in the fix to getPortfolio. In the case of the non-prefixed currency pairs the API request sends the pair in the follow format `XBTBCH` but will receive balances in a format similar to this` { [XXBT] => 1.0, [BCH] => 1.0 }` where the currency is still prefixed, but the asset is not.

exchanges/kraken.js
```diff
@@ -423,31 +441,15 @@ Trader.getCapabilities = function () {
...
       { pair: ['LTC', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' }, precision: 6 },
-      { pair: ['XDG', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' } },
-      { pair: ['XLM', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' } },
-      { pair: ['XRP', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' } },
       { pair: ['CAD', 'XBT'], minimalOrder: { amount: 0.1, unit: 'asset' }, precision: 1 },
       { pair: ['EUR', 'XBT'], minimalOrder: { amount: 0.1, unit: 'asset' }, precision: 1 },
       { pair: ['GBP', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' } },
       { pair: ['JPY', 'XBT'], minimalOrder: { amount: 1, unit: 'asset' }, precision: 0 },
       { pair: ['USD', 'XBT'], minimalOrder: { amount: 0.1, unit: 'asset' }, precision: 1 },
-      { pair: ['DASH', 'XBT'], minimalOrder: { amount: 0.01, unit: 'asset' }, precision: 5 },
```
cmroche, line 434: These were incorrect entries 

`['DASH', 'XBT']` is backwards and duplicates `['XBT', 'DASH']` which is correct and already present. The same is true for the other lines and attempts to use one of these pairs would throw an error that the currency is unsupported.

askmike:
>  Normally my position on this is that someone calling the API with an incrementing trade id wouldn't expect the same results to be returned twice so I approach it from that perspective. If you think it is safe I will undo the filter, but do you think this could lead to unexpected side effects down the road?

The original idea was that calling `getTrades` is stateless. For example it is called twice on start (when running Gekko over the commandline, when the strategy needs a "warmup"): once for figuring out whether we can sync exchange data with data stored locally and once to actually seed the strategy. Gekko expects both to return (roughly) the same data, if you filter all the traders on the second call this assumption does not work anymore.

> If you think it is safe I will undo the filter, but do you think this could lead to unexpected side effects down the road?

It shouldn't especially if the TID is an ever increasing number, but with Kraken this is not the case. You can leave it for now as is and I'll give it a proper test ASAP.

askmike:
Apologies for the slow response, and fair play for the work you put in! Major thanks! I've tested it and it works great. I might ping you once issues come up.

Issue #918, opened 2017-08-04
BitFinex transaction history import for backtesting


pelletencate:
BitFinex allows import of history trades using the `v1/trades/SYMBOL` endpoint. Gekko should be able to use it to import trades in order to backtest on them.

askmike:
Good idea, I need to work on documentation on how to add importing support per exchange..

cmroche:
@pelletencate Changes were merged into develop this morning, if you grab the latest you should be able to confirm that bitfinex importing is working.

PR #1131, opened 2017-09-19
Support for Bitfinex Import

cmroche:

(feature) Support for importing trade data from Bitfinex

Issue #918 

Bitfinex import is functional though due to the high volume nature of this exchange it should be noted that those imports can be very slow and large.

The importer uses a 2 second delay between requests to prevent coming up against the API rate limit. I will attempt to lower this in subsequent tests. While the Bitfinex API documentation says that you can either do 60 or 90 requests per second (depending on the source in their website) in practice I'm finding it to be a little bit lower than that.

Since it is only possible to import historical data using the `v2` api, I've update the bitfinex-api-node library and made minor changes for compatibility to the initialization code. The regular exchange functions still use the `v1` API as `v2` is considered beta and a bit of a pain to change over.

* **Other information**:

Bitfinex's new API for trades uses a time range for the import, `start` and `end` must be specified for the query to work, and results are limited to a max of 1000 per query. However if the time range exceeds the record limit the API will return results from the `end` point first heading towards `start`, that is will return more recent trades first. Because of this approach it is necessary to do the import by iterating backwards. This does not appear to pose any problem for the trade batcher, however introduces an oddity in the UI where the progress bar counts backwards from 100% to 0%.

I would be willing to look at that in the future, but as of right now I am unfamiliar with vue and could not easily piece together how data flows from the import process to the front end.


cmroche:
Actually the trade batcher is broken, I'll need to fix it to work with reverse ordered trades.

cmroche:
Got complicated, I need to go back to the drawing board on this one. Closing the request until I sort out lingering issues.

PR #1151, opened 2017-09-22
Fix DASH with kraken

pprados:
- Fix a bug in Kraken API with DASH

- Impossible to trade with DASH and EUR. Invalid trade pair.

- Remove the prefix

* **Other information**:
Same problem [here](https://github.com/carlos8f/zenbot/pull/469), with other pair

### DIFF
--- a/exchanges/kraken.js
+++ b/exchanges/kraken.js
@@ -49,8 +49,10 @@ var addPrefix = function(value) {
 
   if(isFiat(value))
     return fiatPrefix + value;
-  else if(isCrypto(value))
+  else if(isCrypto(value)) {
+    if (value === "DASH") return value;
     return cryptoPrefix + value;
+  }
   else
     return value;
 }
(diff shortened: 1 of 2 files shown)


exchanges/kraken.js
```diff
@@ -49,8 +49,10 @@ var addPrefix = function(value) {
 
   if(isFiat(value))
     return fiatPrefix + value;
-  else if(isCrypto(value))
+  else if(isCrypto(value)) {
+    if (value === "DASH") return value;
```
askmike, line 53: you shouldn't need this anymore since you've added another ifcheck in L72!

askmike (changes requested):

askmike:
Also, what is the IML file?

cmroche:
This is already fixed in pull request #1124 which also fixes the prefixing in getPortfolio for BCH, DASH, EOS and GNO, whereas this change will only fix DASH in that case.

PR #1155, opened 2017-09-23
Constant time SMA indicator

bbjay:
performance improvement

current SMA is in O(n), no tests

SMA in O(1), tests added

* **Other information**:
Verified results in tests are from old implementation.


askmike:
Great stuff! Thanks :)

PR #1158, opened 2017-09-24
Change output for RSI

generalectric:
Changes the output of RSI.js to be this.result like the rest of the native indicators.

RSI indicator uses this.rsi instead of this.result. This is the only native indicator that doesn't output to this.result and can be confusing for many people. 

Fixes RSI output for upcoming commit to have indicator output exposed to the API and stays true to the standard for native indicator output being that all indicators except for RSI currently output as this.result

* **Other information**:


askmike:
:+1:  thanks!

Would you mind updating the RSI tests? They are failing since they are still expecting the `.rsi` prop.

askmike:
Awesome, thanks!

generalectric:
no problem, and thank you!

PR #1162, opened 2017-09-26
Update ppo.js for equality

krieghof:
Changed `var log = require('../core/log');` to `var log = require('../core/log.js');`
Because in all other strats it is with .js


askmike:
:+1:

PR #1166, opened 2017-09-30
execute end function after backtest complete

generalectric:

adds an end function to strats

strats arent capable of knowing when a backtest is completed

strats will be able to execute code after completing a backtest for example allowing writing to drop files or further processing of arrays

* **Other information**:


askmike:
Perfect, thanks! Would you mind removing it in the custom strat (might confuse people), since this is only really needed for advanced users (who just want to backtest).

generalectric:
Ok, I only added it to custom.js as an example of its use. Looking at gekko/docs/strategies/creating_a_strategy.md it could be documented here, maybe this is a better place for it?

askmike:
Thanks a lot!

PR #1172, opened 2017-10-01
change StochRSI.js to use correct indicator output

krieghof:


askmike:
RSI was updated recently, nice catch! Thanks :)

PR #1174, opened 2017-10-02
Fix „INVALID_API_KEY“ error, fix #1054

### LINKED ISSUE #1054: API key invalid - bittrex
keep getting 

2017-08-29 21:42:47 (DEBUG):    getTicker result { Bid: 0.02630017, Ask: 0.02677724, Last: 0.02630004 }
2017-08-29 21:42:47 (DEBUG):    getPortfolio called
2017-08-29 21:42:47 (ERROR):    getPortfolio Error { success: false, message: 'APIKEY_INVALID', result: null }
2017-08-29 21:42:47 (DEBUG):    Bittrex returned an error, retrying. [ [Function] ]
2017-08-29 21:42:57 (DEBUG):    getPortfolio called
2017-08-29 21:42:58 (ERROR):    getPortfolio Error { success: false, message: 'APIKEY_INVALID', result: null }
2017-08-29 21:42:58 (DEBUG):    Bittrex returned an error, retrying. [ [Function: bound] ]

each time I run. I checked for all misplaced commas, etc. 

here is the key section of my config file:

 config.trader = {
  enabled: true,
  key: '*******************************************',
  secret: '**********************************************',
  username: '', // your username, only required for specific exchanges.
  passphrase: '', // GDAX, requires a passphrase.
}

I've tried with several keys, and redownloading. Not sure what to do here

1024kilobyte:
Always init exchange with trader config if trader is enabled

Bug fix

If the trader is initialized after the trading advisor the api key and secret are not set correctly (at least for bittrex) #1054

Now the exchange is always initialized the same way if the trader is enabled

* **Other information**:

### DIFF
--- a/core/tools/dataStitcher.js
+++ b/core/tools/dataStitcher.js
@@ -177,10 +177,14 @@ Stitcher.prototype.checkExchangeTrades = function(since, next) {
   var provider = config.watch.exchange.toLowerCase();
   var DataProvider = require(util.dirs().gekko + 'exchanges/' + provider);
 
-  var exchangeChecker = require(util.dirs().core + 'exchangeChecker');
-  var exchangeSettings = exchangeChecker.settings(config.watch)
+  var exchangeConfig = config.watch;
+  
+  // include trader config if trading is enabled
+  if (_.isObject(config.trader) && config.trader.enabled) {
+    exchangeConfig = _.extend(config.watch, config.trader);
+  }
 
-  var watcher = new DataProvider(config.watch);
+  var watcher = new DataProvider(exchangeConfig);
 
   watcher.getTrades(since, function(e, d) {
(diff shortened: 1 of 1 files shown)


1024kilobyte:
Just a little explanation of the behaviour I found and fixed:
The first time the exchange is initialized from the trading advisor (data stitcher) it is called with the watch-config only. This way (at least for the bittrex.js) the properties key and secret are created on the watch object and are set to 'secretApiKey'. So when the exchange is initialized again with the object from _.extend(config.watch, config.trader) the key and secret properties from watch are not overriden with the correct values from config.trader. This way the trader trys to connect with apikey and secret set to 'secretApiKey' resulting in the described error. This PR should fix that problem.

askmike:
:+1:

PR #1177, opened 2017-10-03
Fix sample config broken link to command-line backtesting doc

discosultan:


askmike:
Ah nice find! Would you mind linking to here instead?
 https://gekko.wizb.it/docs/commandline/backtesting.html

discosultan:
Done!

askmike:
Awesome, thanks!

PR #1179, opened 2017-10-03
Fix bug with kraken getTrades

cmroche:
* **What kind of change does this PR introduce?** (**Bug fix**, feature, docs update, ...)

Bug fix for market watch and live trade within Kraken

Calling getTrades on Kraken with an empty "since" parameter will break by requesting data from the beginning of time. (Epoch TS 0)

Issue was due to erroneous use of !_.isNull(since) in a test, and really I just wanted to check if any value was defined. undefined, null or 0 should all prevent using the since argument to filter requests.

* **Other information**:


askmike:
Ah nice catch! Will probably do a new release to get this hotfix in asap.

PR #1185, opened 2017-10-06
add journalmode to sqlite plugin config

generalectric:

Adds journal mode to sqlite plugin config

Currently WAL is hardcoded

Users will be able to select between WAL or DEL based on their operating system or setup/needs

* **Other information**:


generalectric:
Possibly related to #1189

Also I was thinking maybe I should add a default of WAL so backwards compatibility is maintained to prevent braking old configs that do not specify a journal mode. Thoughts?

generalectric:
Added default to WAL to prevent braking legacy configs in the case there is no specified journalmode sqlite option or a user simply hasnt upgraded their config

askmike:
Apologies for the late reply :+1:

PR #1192, opened 2017-10-08
fix precision

ebrulato:

It fixes precision calcul
It adds a precision security on the price
It adds more detailed error messages when 'no data' is returned. 

the article https://blog.kraken.com/post/1278/announcement-reducing-price-precision-round-2/ talk about "price precision". 
sometimes we can have a bug on the price precision.

avoid a price precision error which causes an infinite loop of retry.

* **Other information**:


askmike:
great stuff, thanks!

Issue #708, opened 2017-04-21
Support tulip indicators (windows TA library)


askmike:
Gekko supports [TA-lib](http://ta-lib.org/index.html), a well known technical analysis library that supports a lot of different indicators. However it is very hard to compile this library on Windows (which Gekko supports).

[Tulip indicators](https://tulipindicators.org/) is a relative new library that supports roughly the same indicators. The node wrapper [just added precompiled windows binaries](https://github.com/TulipCharts/tulipnode/issues/2) which makes installing on windows a breeze.

Please +1 this if you want this implemented, will work on this regardless but might prioritise if a huge number of people want this.

generalectric:
Impressive looking at the benchmarks and well documented too... +1!

thegamecat:
So I agree with doing this but....from my testing on windows, just run gekko in bash on windows 10 then talib is fine.

If you use SQLite then go to the file gekko/plugins/sqlite and change the line 53 from:
db.run("PRAGMA journal_mode = WAL");

to 

db.run("PRAGMA journal_mode = DEL");

And if you're copying SQLite dbs from another install of gekko then make sure you change the db pragma journal mode to DEL in a tool like DB Browser for SQLite.

I would love to know the performance difference between DEL and WAL or even if they are used, but I have a 500meg btcusd dataset and it seems to make no difference.

All that said, the more indicators the merrier, I am personally unsure if it should be a priority, but I'd be happy with it integrated.

Edit; just reviewed the performance....this would be marvelous to implement! +1 from me.

Pedro942:
+1

vpatel9202:
+1

deke997:
+1 
I would love to have this implemented! In the mean time, I have installed gekko in bash on Windows 10.  Thank you @thegamecat for the instuctions to get that working... I was pulling my hair out trying to figure it out, but once I made the necessary change to handle.js, everything started working flawlessly (I think... at the very least, it stopped throwing errors)

Maybe we should add instructions to the docs to get gekko running on bash on Windows 10

BrentEaston:
+1

serzhiio:
I need awesome oscillator but Ta-Lib doesn't have one.
Tulip is in need...

askmike:
All thanks to @cmroche this is now available in the develop version of Gekko! (download here: https://github.com/askmike/gekko/tree/develop - using that link you can also browse the `docs` folder to find updated documentation on how to implement this in your strategies).

[...]

PR #1196, opened 2017-10-10
Add support for Tulip Indicators

cmroche:

This change adds a new feature by providing support for Tulip Indicators (via the tulind npm package), which provides a large library of efficient TA indicators that works on all operating systems.

It also provides a new simple strategy (tulip-macd) used to test the integration.

* **Other information**:

The indicator library is enabled by default, unlike TA-Lib since the library is available on all major OS's including windows.

Should resolve enhancement request #708


FLYBYME:
Code is very clean nice work.

askmike:
Awesome stuff! Thanks a lot for your hard work :)

I am not super happy with how sloppy I did the TAlib integration, but we can't wait forever before I have time to clean it up. So when I do I'll clean TAlib as well as tulip up.

Issue #1221, opened 2017-10-19
[bug] Wrong gekko version shown in the UI after updating to latest changes - Docker Image


dpsarrou:
**I'm submitting a ...**
[x ] bug report

**Action taken** (what you did)
- Cloned latest stable version v0.5.8
- Installed using Docker following the instructions on https://github.com/askmike/gekko/blob/stable/docs/installation/installing_gekko_using_docker.md

**Expected result**
UI should show Gekko version 0.5.8 at the footer

**Actual result**
UI shows Gekko version 0.5.2 at the footer

**Other information**
I've noticed that this is happening because:
- The UI is prebuilt for version 0.5.2 and committed to the repo as shown here:
https://github.com/askmike/gekko/tree/stable/web/vue/dist even though the latest stable is 0.5.8
- Docker build will not run the build process for the UI

To avoid rebuilding and commiting a new build.js file every time a new version of gekko is released, I can create a PR that does the following:
- Remove the web/vue/dist/build.js file from the repository since it can (and should) be rebuilt when installing from source
- Include a UI build step in the Dockerfile
- Update the documentation to notify people that they need to build the frontend as well

askmike:
This has since been fixed in v0.5.9. Let's keep the rest of the discussion going in the PR.

PR #1223, opened 2017-10-19
Hotfix/fix ui version

dpsarrou:
**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug Fix.
This PR addresses the bug with incorrect Gekko version shown at the UI, due to old artifact of the UI committed to the code.

**What is the current behavior?** (You can also link to an open issue here)
- UI is prebuilt for an older version (0.5.2) while current stable is 0.5.8
More information at #1221

**What is the new behavior (if this is a feature change)?**
- Fixes the bug by rebuilding the frontend
- After this update, users will need to additionally build the frontend (docs are updated)


docs/installation/installing_gekko.md
```diff
@@ -33,6 +34,20 @@ Once it is installed we need to install Gekko's dependencies, open your terminal
 
     npm install --only=production
 
+## Installing Gekko UI dependencies
```
askmike, line 37: The idea is that each new version (I push to Github) includes a compiled version of the frontend. End users are not meant to compile this manually. This is okay for docker users (since they are most likely advanced and not running a release but the "develop" branch, which does not always include the latest compiled version) to add this to the docker file. But I don't think this belongs in the default installation documentation.

askmike (changes requested):

docs/installation/installing_gekko.md
dpsarrou: Is there a specific reason that end users are not meant to compile the UI manually..? Since end users will always have to run `npm install` anyway to compile gekko, we could avoid any version mismatches such as the one described in this issue.

I can think of 2 options that could help improve the situation although I'm not sure for the feasibility of the first one:
- Define the UI compilation step as a task in the gekko `package.json` since users will always have to install that anyway. Unfortunately my experience with npm is minimal and I couldn't find an easy way to run npm with a different root directory (`web/vue`). Maybe this is not possible at all.
- Replace the installation process with a simple installation script, for example `install.sh`. There we could wrap the build process for gekko and gekko UI while keeping it easy to use and transparent for the users.

I would opt for the second approach as it is less of a hack. :) What do you think..?

askmike: > Since end users will always have to run npm install anyway to compile gekko

They run `npm install` inside the `gekko` folder, not inside the `gekko/web/vue` folder (which is required for compiling the UI). If we go for other of these options we should remove the currently compiled version out of git.

What is the point of compiling the frontend on install for stable releases (`stable` branch) that already have a compiled build?

> Is there a specific reason that end users are not meant to compile the UI manually..? 

Yes: compiling the frontend requires a lot of memory and takes a while, I have had this fail in a raspberry PI before. I want to move towards an easier installation process where less can go wrong (which means less installation issues in github), not towards a bigger one.

That said, I see a few different use cases:

1. non technical users who want to try out the tool: they download from github.com and having to install nodejs is already an extra step.
2. tech savy users who want to use docker/dev branch: they are building dockers anyway, it can definitely be part of the build process for them.

I don't want to burden the first one would a bigger installation process that takes longer and can break in more places. There are enough issues on github about people with installation problems as it is.

askmike:
I know see the current stable has an outdated UI build, I manually fixed this for now.

docs/installation/installing_gekko.md
hoshsadiq: I disagree. It doesn't quite make sense to potentially be running an older UI but the latest on the backend. If you push a new version, both backend and frontend need to be compiled and pushed. If you're simply running it by git clone, this shouldn't be the case, I should have to manually install and run both items.

billymcintosh: I see both sides and end users do love to get up and running ASAP. 

As a quick hack, I added a `setup` run-script to the `package.json` and it worked flawlessly.   I do wonder about Windows paths though.

```json
"scripts": {
  "setup": "npm install && cd web/vue && npm install && npm run build",
  "test": "./node_modules/.bin/mocha test/*.js --recursive test -u tdd --reporter spec",
  "start": "node ./gekko --config config.js --ui"
},
```
Taking it further, a `build.js` could still be bundled and `rimraf dist/build.js` could be initiated before the new build takes place.

I would consider `better-npm-run` if something like the above was taken further.

PR #1227, opened 2017-10-19
Feature/issue 1154 prettier roundtrips

taylorbett:
Small UI/Front end feature with added functionality for users and some small UI bug fixes/issue requests

Currently there is no ability to sort the roundtrip data at all in the web UI, nor were there any currency symbols in the entry/exit balance columns, plus some other small changes like the green text colour changing for both the Round Trip view and the Paper Trade summary view.

Adds front end sort functionality for the Round Trips view, can sort data by a specific column in ascending or descending order, including number and time data. Also adds currency symbols and some other small fixes from #1154

### DIFF
--- a/web/vue/src/components/backtester/result/result.vue
+++ b/web/vue/src/components/backtester/result/result.vue
@@ -7,7 +7,7 @@
     .hr.contain
     chart(:data='result', height='500')
     .hr.contain
-    roundtripTable(:roundtrips='result.roundtrips')
+    roundtripTable(:roundtrips='result.roundtrips', :report='result.report')
 </template>
 
 <script>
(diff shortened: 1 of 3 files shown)


askmike:
Apologies for the super late delay! I like this a lot, two small questions:

- can you remove this? https://github.com/askmike/gekko/pull/1227/files#diff-adf79107efe4cda153ebccb08a8cb1c0R34 
- can you instead of the controls make the headers clickable?

If not no worries and I'll merge this in and change that on top myself :)

susanconstant:
Yes, can't pull but can trust you.

susanconstant:
Please?

PR #1228, opened 2017-10-20
fix pg library > 0.7 errors

meyerd:

Bug Fix

Error with pg library > 0.7 when using Postgres for candle storage.

Quick fix errors for new pg api.

* **Other information**:


askmike:
:+1: thanks

PR #1239, opened 2017-10-22
Return action taken and candle data from advice() function

generalectric:

Allows return data including candle (id, open close etc....) as well as action taken returned from advice() function like so.... var tradeResult = advice(tradeDecision) where trade result will be an object accessible like so: tradeResult.action or tradeResult.candle['close']

strats have no way of knowing what action was taken or exactly what candle it happened on when using advice();

A more straightforward and accurate (less prone to error) way of knowing exactly when and where a trade actually happens vs trying to sync your strat to what is really going on when advice() is processed

* **Other information**:


askmike:
I don't fully understand the need for this: The advice will always be emitted if it was "short" or "long" (and if it wasn't the same as the previous advice). So the only case in which it gets ignored is when you advice `SHORT - SHORT` or `LONG - LONG`, if you really want to know when this happens, why don't you track it yourself in your strat?

generalectric:
thats exactly my point. we can eliminate the need to track it ourselves in our strats and allow us to keep our code simpler and cleaner. also saying that we could track it ourself in our strat is the same as saying we can have our strat decide if advice long should mean to advice long whether or not the last advice was long or short

askmike:
You can already access `this._prevAdvice` in your strategy which is set to whatever the previous advice was.

Though I really don't think this is the way to go: a strategy can always advice, but it's up to Gekko whether it does something what that advice (ignore if duplicate, not enough money to make a trade, etc). And your PR only feedbacks whether Gekko decided to emit the advice or not, nothing more (like response from the trader plugin, etc). So the `action` prop doesn't actually tell you what action Gekko took (if any at all).

generalectric:
And that's another thing I wanted to add to it was return values for amount of asset or currency along with candle ID and action taken. Honestly I took a look and was like well for now it looks like we're only gonna be able to retrieve the result of gekkos decision after blocking a double buy or sell and the candle that it happened at with ID price etc. I realize that I can do all this in strat. But I think it's a start and should actually be expanded on.

What actually led to this for me is that I need to process data in strat and have the results in csv.  And this would actually eliminate the problem altogether for me personally. If I could have processed data emitted as api output. Then I wouldn't be running the command line backtest I would just post to the api and get the result and it would have everything the buy sell candle id's prices indicator data... And the output I need from data processed in strat.

 If you don't want to accept the PR then ok I am fine with it. I agree it can be done in strat. But if you can point me in the right direction with how can I go about exporting my data for each candle through the api that would actually be even better and I would care less about returning data from advice function

thegamecat:
This is a good change to make.

askmike:
@generalectric I am pretty sure we want to end up with the same goal:

The strategy wants to know what happened to the advice: if a trade was executed and if so at what time/price. The good thing is that most of this is already present: after a trade an event is emitted, we only need to expose this to the strat and you should have what you want.

Because do you really care WHY no action was taken? because either duplicate action or not enough funds on account? I think you just want the result.

The thing is that the real trader needs a while to figure this out: on new signal it needs to check exchange balances, try to add a limit order, wait until it gets taken, if not re issue, etc. This takes time and thus means we cannot `return` the action from `this.advice` (because by the time we propogate the advice through Gekko all of the above things are not yet known).

Let me know if I correctly understand what you are trying to do, because if that's the case this PR is not the way to achieve it.

generalectric:
Well explained @askmike, makes perfect sense. And yes you understand correctly, and now I understand as well. I was only thinking about backtesting, I wasnt at all thinking about live trading. So I agree and like your idea better. However.... can we find a way to emit custom data from a strat to the API? Im still not familiar enough to figure out a way to do this.... but cant we just reserve an object designated for custom data... like customdata:{} that can be retrieved through the api?

askmike:
It is an important goal for Gekko to behave the same when backtesting as in
the live market (else what is the point of the backtesting).

It will take some time to hook it up, but the best way would be for
strategies to register a new method called 'processTrade' where they can
handle a trade event coming from the trader. All the logic is actually
already there in most of gekko (the code that calculates statistics already
implements this processTrade).

Issue #1041, opened 2017-08-25
Postgresql error: query.on is not a function


pushkarnagle:
I am trying to use Postgresql as DB adapter. I have installed latest version, created new user and database and have granted all privileges to the new user.
But while running Gekko, I am getting below error:
```
2017-08-25 13:34:22 (DEBUG):	Check database exists: gekko
/root/gekko/plugins/postgresql/handle.js:47
  query.on('row', function(row) {
        ^

TypeError: query.on is not a function
    at /root/gekko/plugins/postgresql/handle.js:47:9
    at Connection.<anonymous> (/root/gekko/node_modules/pg/lib/client.js:153:7)
    at Connection.g (events.js:292:16)
    at emitOne (events.js:101:20)
    at Connection.emit (events.js:188:7)
    at Socket.<anonymous> (/root/gekko/node_modules/pg/lib/connection.js:118:12)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)

```

pushkarnagle:
I installed old version of pg by `npm install pg@6.4.0` and the issue is solved!

askmike:
great :)

davidlondono:
Add it to the Dockerfile, is causing issues

pushkarnagle:
Well, installing an old version gave rise to another problem. I am now using version 6.4.0 of pg, but getting below error -
```
events.js:160
      throw er; // Unhandled 'error' event
      ^

error: syntax error at or near "ON"
    at Connection.parseE (/home/pi/gekko/node_modules/pg/lib/connection.js:567:11)
    at Connection.parseMessage (/home/pi/gekko/node_modules/pg/lib/connection.js:391:17)
    at Socket.<anonymous> (/home/pi/gekko/node_modules/pg/lib/connection.js:129:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:547:20)
```

EDIT: This might be helpful https://node-postgres.com/guides/upgrading
There are some changes in pg@7.0.0 onwards.

pushkarnagle:
@askmike Can you please fix this? `query.on` is not supported in pg 7.2.0

siedi:
[...]

PR #1245, opened 2017-10-25
Hotfix - pg version on docker

nourspace:
Bug fix

When installing using Docker `pg@7.x` is installed which has different API than `pg@6.x.`
Issue: #1041 

It now installs a working version of node-postgres when using Docker

* **Other information**:
I tried migrating to `pg@7.x` and later decided not to as the deprecations touch many points and to make sure everything is correctly migrated I need more time.


askmike:
Thanks for the PR, but why the linting? This is not how the rest of the codebase is styled ;)

nourspace:
I tried to match the style but seems it didn't work out well. I think something like eslintrc would be great or some kind of doc to refer to like https://github.com/airbnb/javascript


Closed in favor of #1257

askmike:
I agree about the outdated code style, though Gekko is an old project and the majority of the code was written a long time ago.

In newer projects I am using [prettier](https://github.com/prettier/prettier) which restyles all code to one common form, similar to the go fmt for golang. I might do that at some point for Gekko, but I am afraid I'll overwrite too many things that didn't change (messing up tools like git blame).

nourspace:
Makes sense. I just wasn't sure what to do with things that obviously needed reformatting.
I will just ignore these in future PRs.

Issue #1246, opened 2017-10-25
Posting to API: RequestError ESOCKETTIMEDOUT


generalectric:
My strat works fine and posts to the API for backtesting... but seldom it brakes and stops and more so when I test smaller candles. Im assuming its doing just as it says.... Timing out. I've given it a good look but I'm unable to figure out how to adjust and give it a longer timeout. Also would it make sense to have this as a config option in UIconfig.js?

    name: 'RequestError',
    message: 'Error: ESOCKETTIMEDOUT',
    cause:
     { Error: ESOCKETTIMEDOUT
         at ClientRequest.<anonymous> (/home/adm1n/gekko/node_modules/request/request.js:782:21)
         at Object.onceWrapper (events.js:293:19)
         at emitNone (events.js:86:13)
         at ClientRequest.emit (events.js:188:7)
         at Socket.emitTimeout (_http_client.js:679:10)
         at Object.onceWrapper (events.js:293:19)
         at emitNone (events.js:86:13)
         at Socket.emit (events.js:188:7)
         at Socket._onTimeout (net.js:352:8)
         at ontimeout (timers.js:386:14)
         at tryOnTimeout (timers.js:250:5)
         at Timer.listOnTimeout (timers.js:214:5) code: 'ESOCKETTIMEDOUT', connect: false },
    error:
     { Error: ESOCKETTIMEDOUT
         at ClientRequest.<anonymous> (/home/adm1n/gekko/node_modules/request/request.js:782:21)
         at Object.onceWrapper (events.js:293:19)
         at emitNone (events.js:86:13)
         at ClientRequest.emit (events.js:188:7)
         at Socket.emitTimeout (_http_client.js:679:10)
         at Object.onceWrapper (events.js:293:19)
         at emitNone (events.js:86:13)
         at Socket.emit (events.js:188:7)
         at Socket._onTimeout (net.js:352:8)
         at ontimeout (timers.js:386:14)
         at tryOnTimeout (timers.js:250:5)
         at Timer.listOnTimeout (timers.js:214:5) code: 'ESOCKETTIMEDOUT', connect: false },
    options:
     { url: 'http://localhost:3001/api/backtest',
       json: true,
       body: { gekkoConfig: [Object], data: [Object] },
       headers: { 'Content-Type': 'application/json' },
       timeout: 90000,
       method: 'POST',
       callback: [Function: RP$callback],
       transform: undefined,
       simple: true,
       resolveWithFullResponse: false,
[...]

PR #1249, opened 2017-10-26
add timeout option for api in uiconfig

generalectric:

nodejs server has default timeout of 2 minutes. This PR adds a timeout configuration option to uiconfig for api ability to run backtests with longer execution time

api backtest is capped at 2 minutes

users will be able to extend the timeout beyond the default time of 2 minutes to allow for longer backtest execution (eg:  multiple async runs using 1 year data with 1 minute candles... or just a slower system such as a raspberry pi using large data)

* **Other information**:

### DIFF
--- a/web/server.js
+++ b/web/server.js
@@ -77,7 +77,7 @@ app
   .use(router.routes())
   .use(router.allowedMethods());
 
-server.timeout = config.api.timeout;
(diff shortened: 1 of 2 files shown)


generalectric:
addresses issue #1246

askmike:
:+1:

PR #1257, opened 2017-10-28
Hotfix - pg version on docker

nourspace:
Bug fix

When installing using Docker `pg@7.x` is installed which has different API than `pg@6.x.`
Issue: #1041 

It now installs a working version of node-postgres when using Docker

* **Other information**:
I tried migrating to `pg@7.x` and later decided not to as the deprecations touch many points and to make sure everything is correctly migrated I need more time.


askmike:
:+1:

PR #1264, opened 2017-10-29
Fix UI service

cmroche:
Unable to connect to the web UI on port 3000, service is not binding on this port.

Change UIConfig.js API config back to port 3000

This was broken by a change in commit 92bc3da92929d32e6e3eaf34a5cea557758b703a which was adding a timeout option to the web server. The timeout still remains, but the change to port 3001 has been undone.

@generalectric


generalectric:
Oh no! I am not quite sure how I managed to do that but I do apologize and thank you for fixing!

askmike:
ahhh damn, nice catch!

PR #1265, opened 2017-10-30
Add new feature to specify humanized date ranges for import and backtest

cmroche:

Add support for humanized date ranges for import and backtest configs

Currently backtesting will only support an explicit to and from timestamp, or scan. Import only supports explicit 'to' or 'from'.

With this change for either backtesting or import you can specify the date range in a humanized format, using the 'juration' package string like this are parsed:

'30 days'
'1 month 10 days 6 seconds'

and the correct date range (from the current time) is computed.

This benefits automated script which may want automated updates/windows for testing by allowing for a simpler, non-changing configuration which reflects this behaviour.

* **Other information**:

'juration' 0.1.0 added to npm package list to support this feature

Tested import and backtest in various configurations (bitfinex and kraken).


askmike:
three things:

- so for now this would only be a CLI feature (not exposed to the UI)? We could later extend it to the UI as well of course. So not a problem, just an observation.
- If you set a duration it would always kind back from "right now", which works reliably for people who are running watchers/paper traders on the side (and thus have up to date data), but not for people who have previously imported I think.
- For stability reasons we need to force the start and endtime to precise minutes: some candle fetching/calculation logic expects all dates to be precise to the minute (we can simply round it using moment or so).

cmroche:
For now CLI only, but yes result in a simpler UI workflow since I imagine most people are only looking for "last so many days" when importing and backtesting anyway.

Yes, you are right that there are some issue with previous imports if there are missing candles at the end of the range. I'm going to look at this issue during the weekend. I think we should consider clamping the backtest to fit the available data in any case that precise dates and times are not given for the testing. However for automated scripts where backtests are happening immediately after import it is sufficient, or for the watchers/paper traders.

As for the last point I'll make this change tomorrow. Seems simple enough.

askmike:
> I think we should consider clamping the backtest to fit the available data in any case that precise dates and times are not given for the testing.

The current behaviour is a bit more extreme: if you set the backtest range to `scan` Gekko await the backtest until:

1. Gekko scanned local databases for possible datasets (that do not have gaps, eg. if you have JAN and JUN but not between those will be shown as 2 different datasets)
2. The user picked the desired dataset (in case of multiple).

That's the only "user friendly option". However, right now the only other option is the set a specific timerange (`from` and `to`). And gekko is not very forgiving if you didn't set it properly:

https://github.com/askmike/gekko/blob/095d2d4160951d4b94e16261951d5fab2a9d37b9/core/markets/backtest.js#L78-L80

Though that path would likely get hit of you use this feature but don't have really up to date data.

cmroche:
I've made changes to the way backtesting works, so that batchSize limits the number of records processed in a single tick, versus selecting the time range. This allows us to detect and better handle partial market data scenarios in a robust fashion.

With this new approach this is what can happen:

1) Range is specified and no market data exists within. Script will die and throw an error indicator that there was no market data (same behaviour).

2) Range is specific and all market data is available. Script will process normally.

3) Range is specified and market data is missing from the beginning or middle. A warning is printed to indicate that there is missing market data, but will process what is available. Previously missing data at the start would cause the script to abort if the missing time period was greater than batch size.

4) Range is specified and market data is missing from the end. A warning indicating that data is missing from the end if printer, and the rest is processed normally (you will see the performance analysis). Previously this would cause the script to abort if the missing time period was greater than batch size.

This should address any concerns about using "30 days" style of date entry into the time range. Since it will gracefully truncate the market data to the available ranges.

@askmike

askmike:
will try to look at this asap.

Issue #956, opened 2017-08-13
[TODO] Simpler configuration


cassvail:
It took some time to trying to understand the project configurations, which is poorly documented  (I understand this takes time)
I also had some troubles configuring the app in a "production" server, due to the many configurations all around the project

A few improvements could be implemented:
* I see many duplicated configurations inside the project (toml files, config-sample.js => config.js, UIconfig.js, baseConfig.js). I don't really understand the reason for this, but I think it would be a lot simpler for both the cli and the backend in general to have a unique centralized configuration in the config.js (with the config-sample.js committed) and to remove all the other files.
The config.js is where the user would look into immediatly and understand where to put his personalizations.

* baseConfig.js contains mostly duplicated configs from config.js (or viceversa), which could easily allow to merge in a unique config.js

* Also, in my opinion it, UIconfig.js  should not need to contain backend configs like "API url", usually the backend who serves the UI is it's own API backend so _relative_ urls would be enough. In case the backed is actually a different URL it could be the backend itself to serve the UI that information (moving again the configurationt to the backend config.js)
At this point the UIconfig.js is empty :)

* One single config.js file, it's simpler to configure but also simpler to document (resolving the first stated issue)

It's a bit of info, but I hope it helps!

askmike:
Hey!

Great points, and you are 100% right. Let me quickly explain how we got to the current mess:

- Gekko started out as a CLI tool, all configuration was stored in `gekko/config.js` (at some point users had to copy `gekko/sample-config.js` but the idea is the same).
- Everyone wanted a web UI, one of the goals was:
  - you can configure your gekkos in the UI, so you don't need to touch any config files anymore.
 - But then people wanted to configure their UI (ports, host, etc) so a new UI config file was created (`UIconfig.js`)
- The UI should make it easy to "configure" gekkos, for this the UI needs to know what kind of config options are available (and what their defaults are for UX reasons), but editing JSON in a webinterface is crap so a bunch of TOML files (for different parts of the config) were added. All of these are stored in the `gekko/config` directory.
- But there are some config settings that you rarely need to touch (where the DB is stored, what version of talib to use, etc), so a `baseConfig.js` exists for *advanced* users to adjust these settings *when using the UI*.
- But because a lot of users still want to run the CLI interface we can't just remove the legacy ` config.js` and `sample-config.js` system.

TLDR:

- some people want to run gekko over CLI, they need a config file, easiest way is to copy over `sample-config.js`. 
[...]

PR #1266, opened 2017-10-30
Added in code to use the config from commander

albertcbrown:
Bug fix - I wanted to use gekko with postgres and it was ignoring the -c config.js option.  These are the changes I made to have it use the config

It uses the defaults for PG (user:password) and it would fail due to no role for 'user'

-c config.js option on commandline is respected.

* **Other information**:


askmike:
Hey! Thanks a lot of this PR. Quick question: Does the bug description applies for the UI? As in did you start Gekko like `node gekko --ui -c myconfig.js`? In that case:

The story about the config mess is quite long one, TLDR: the idea is that you only need a `config.js` when manually running your gekkos (via the CLI), if you use the UI the config will be generated on the fly (depending on what you click).

To provide more flexibiliy regarding "advanced options" (like debug levels and database options) some options were added to the `UIConfig.js`, [see here](https://github.com/askmike/gekko/blob/stable/web/vue/UIconfig.js#L18) (the UI will always make sure the executed Gekko use the specified adapter).

A major overhaul of a simplified version of the config situation was discussed and agreed upon in #956 (though I am not sure if anyone started working on this yet).

---

if this is a bug regarding CLI usage let me know and ignore all the above :)

askmike:
Closing due to inactivity.

PR #1271, opened 2017-10-30
add schema to postgresql config for UI/API

generalectric:

    Adds schema to postgresql config for the UI and/or API


askmike:
Would you mind also adding it to sample-config? So this change will be the same for both CLI and UI users?

generalectric:
It's actually already in Sample-config in develop branch it was just missing in the ui base config

askmike:
And I am way behind on changes... You're right haha!

PR #1282, opened 2017-11-02
Fixing calculation issues with roundtrip table

mattgauzza:
Changing this.price to trade.price and this.current to trade.portfolio.
 stable


plugins/performanceAnalyzer/performanceAnalyzer.js
```diff
@@ -100,7 +100,7 @@ PerformanceAnalyzer.prototype.round = function(amount) {
 }
 
 PerformanceAnalyzer.prototype.handleRoundtrip = function() {
-  const roundtrip = {
+  var roundtrip = {
```
askmike, line 103: why?

askmike:
Damn race conditions, pretty sure you are right!

PR #1289, opened 2017-11-05
fix sqlite pragma to use config setting

generalectric:

    add config setting back to pragma for sqlite plugin

    sqlite plugin somehow got set back to just use hard coded WAL pragma

    sqlite will once again use setting from plugin config, or default to WAL if no config setting specified

* **Other information**:

### DIFF
--- a/plugins/sqlite/handle.js
+++ b/plugins/sqlite/handle.js
@@ -50,7 +50,7 @@ if(mode === 'realtime' || mode === 'importer') {
 }
 
 var db = new sqlite3.Database(fullPath);
-db.run("PRAGMA journal_mode = WAL");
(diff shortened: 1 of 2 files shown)


askmike:
Sorry if we talked about this before, but:  can we add this option to the UI since it isn't present in the base config: https://github.com/askmike/gekko/blob/develop/web/routes/baseConfig.js

because of your check this wouldn't break in the UI, but it will be hard for people to configure this (since the `config.sqlite.journalMode` isn't present and needs to be added).

generalectric:
ah yes, cant forget about the ui config! ok its done, thanks!

askmike:
:+1: thanks :)

Issue #1293, opened 2017-11-06
[Minor Improvement] Indicator loadout


Gab0:
[ ] bug report
[ ] feature request
[x] question about the decisions made in the repository

To create a new indicator, one should add it to the list `Indicators` at `tradingAdivisor/baseTradingMethod.js`
There is a specific reason for this? We could replace that with a search on indicators folder:
```
var indicatorFiles = fs.readdirSync(indicatorsPath);
var Indicators = {};

indicatorFiles.forEach(function(indicator){
    var indicatorName = indicator.split('.')[0];
    if (indicatorName[0] != '_') //ignore those starting with '_', so we have some control on which aren't loaded;
        Indicators[indicatorName] = require(indicatorsPath + indicator);
});
```
Then fix that line where `factory` attribute is required
```
 this.indicators[name] = new Indicators[type](parameters);
```
And set the input type (`candle` or `price`) on each indicator init function.

That would ease the process of creating and renaming indicators..

generalectric:
I dont hate your idea.... but I disagree with it. This makes writing indicators easier being that you dont have to mod basetradingmethod and copy paste then edit a couple words.... but this creates more work for end users.. who might not know if an indicator is price or candle data... and I think there are other things more important than making everyone edit all their existing strats to conform with this?

Gab0:
I see your point.. but man, it will be even easier to see that determined indicator takes one or other input type, because it will be declared on the indicator file itself:

>indicators/DEMA.js [config function]
```
var Indicator = function(config) {
  this.input = 'price'	
  this.result = false;
  this.short = new EMA(config.short);
  this.long = new EMA(config.long);
}
```
basetradingmethod.js and indicator files will be the only affected, no edit on strategies will be required. 
Thats for a very small improvement, for a bit of ease on indicator management. buy hey, its for free x]

askmike:
I think this is a great idea, and I think it's a nice start to some next steps, that are:

Allow for anyone to easily add indicators, for this we would need to apply your proposed fix and create some documentation on how indicators are supposed to behave and voila :)

askmike:
@generalectric 

> but this creates more work for end users.. who might not know if an indicator is price or candle data... and I think there are other things more important than making everyone edit all their existing strats to conform with this?

It is confusing that some indicators are always supposed to calculated over a specific value and others are not (like EMA). We should just make a nice API for this and explain it properly.

[...]

PR #1301, opened 2017-11-08
Indicators - Direct Load

Gab0:
#1293


askmike:
This looks great, one question: would you mind wrapping the require in a try/catch block like we do here:

https://github.com/askmike/gekko/blob/095d2d4160951d4b94e16261951d5fab2a9d37b9/core/pluginUtil.js#L27-L31

Else one buggy indicator will always crash Gekko.

askmike:
:+1: thanks :)

generalectric:
ok this is awesome! sorry i misunderstood gab0. well done sir!

PR #1314, opened 2017-11-11
Fix log.debug statements

andyklimczak:
- Revert partial code from commit 2c5306e12 that broke(?) `log.debug` statements inside strategies
- Fix small typo

bug fix

log.debug is not logging

fix log.debug

* **Other information**:
`log.debug` from OOTB strategies are not working on latest develop
I'm not sure if the `message.log` change was an intentional change, or an accidental addition


askmike:
I removed this (in a commit for something else - apologies) because it was slowing down backtests (since most strats log a lot and all this data was pushed from the backtester -> parent process -> stdout).

But I see this basically hides all logs from the backtesters (when using the UI or API).

I was thinking the following instead:

- in the UI add a "debug" tickbox to the backtest screen, this toggles the log level of the backtester (to either log or not).
- Apply this patch so that if the backtester logs, it should stream up.
- Add documentation on how to properly set up a "create strategy environment".

What do you think?

askmike:
I am actually applying this already, since it is currently actually a bug. Though the above question still holds.

andyklimczak:
For sure, some type of checkbox makes sense (and even control in the code with something like `this.debug = true`)

askmike:
How the config behaves in general (especially when using the UI) is grown a bit out of hand, if you are interested in the full discussion please check out #956.

Before the UI there was one config that did everything, now it is split into a few different categories (how you want everything to behave, like database stuff / how you want the UI to behave, like ports and hosts / how you want individual instances to behave, like this feature).

PR #1322, opened 2017-11-14
docs(supported_exchanges): use check and x symbols

levithomason:
**What kind of change does this PR introduce?**
Docs.

**What is the current behavior?**
_I think_ the `V` in the table is attempting to symbolize a check mark, meaning "supported", while the `X` is "not supported".  However, this is not clear.

**What is the new behavior (if this is a feature change)?**
I've replaced the `V` with HTML checkmark entities and the `X` with HTML multiplication entities.  These render nicely in markdown and in HTML.

**Other information**:
I originally started this PR to add the missing exchanges to the import list but I see that has already been done.  The doc site needs to be updated, though.


askmike:
great stuff, thanks. I will check how it renders on the documentation website ASAP.

askmike:
This looks great :)

askmike:
I tested this fine locally but it broke on production (=the live website):

screen shot 2017-11-19 at 13 57 24

EDIT: Nevermind, fixed by forcing UTF8 from the server.

PR #1330, opened 2017-11-16
Fix for Tulip ADX

mkondel:
Bug Fix. Fixed Tulip ADX indicator. Added tulip-adx sample strat.

Using Tulip ADX indicator will throw error with no messages.

Tulip ADX works, able to do backtesting at least.

* **Other information**:


askmike:
:+1:

Issue #1328, opened 2017-11-16
Kraken trading-pair EOSETH fails + possible solution


klemensh:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[x ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
started a trading-bot of EOS/ETH on kraken

* **Expected result** (what you hoped would happen)
bot starts trading

* **Actual result** (unexpected outcome)
bot does not recognise the current assetAmount

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
LOG:
2017-11-16 11:58:34 (ERROR):	Kraken did not return portfolio for EOS, assuming 0.

**Code (kraken.js - line 178/179)**
I found out that for EOS/ETH the prefix for **cryptocurrencies** is set in line 179:
```
var assetId = _.contains(assets_without_prefix, this.asset) ? this.asset : addPrefix(this.asset);
var assetAmount = parseFloat( data.result[addPrefix(assetId)] );
```

I think this should not be the case.
I assume assets **with** additional prefix would be prefixed twice so I **think** the following change should solve the problem for all pairs, not EOS/ETH only. Only tested for EOSETH though:

```
var assetId = _.contains(assets_without_prefix, this.asset) ? this.asset : addPrefix(this.asset);
var assetAmount = parseFloat( data.result[assetId] );
```

cmroche:
I agree, was an oversight probably on my behalf when trying to sort a bunch of this stuff out. I'll submit a PR in the next day or so, thanks.

Are you able to test the change with ETH/USD, BTC/USD or ETH/BTC though to validate it hasn't broken anything else?

cmroche:
Ok, I figured out what was happening. addPrefix would not add a prefix twice, it's a bit smarter than that which explain why no one else has reported issues with the common currencies yet. EOS is a non-prefixed assets (in the assets_without_prefix list) but the bug was causing it to be added anyway. Your change should work in all cases.

PR #1335, opened 2017-11-18
Fixing prefixless alt coins in getProtfolio (kraken)

cmroche:
This addresses an issue where getProtfolio was unable to check the result for pre-fixless currency pairs due to an accidental "addPrefix" call that was not removed when support for these alt coins was added.


cmroche:
This will fix issue #1328 and should be perfectly safe for prefixed common assets, please check the notes on that issue for more details.

cmroche:
Bonus bug fix, fees were incorrect

askmike:
@cmroche great stuff, as usual, thanks :)

exchanges/kraken.js
```diff
@@ -203,8 +203,12 @@ Trader.prototype.getPortfolio = function(callback) {
   this.kraken.api('Balance', {}, _.bind(setBalance, this));
 };
 
+// This assumes that only limit orders are being placed with standard assets pairs
+// It does not take into account volume discounts.
+// Base maker fee is 0.16%, taker fee is 0.26%.
```
askmike, line 208: Gekko should always be making (unless the price moves hard between a ticker fetch and order placement, which on Kraken I would not rule out).

PR #1344, opened 2017-11-19
Automated load of talib and tulind if available locally

cmroche:

Feature

Users were required to modify configuration files to enable or disable TALIB or TULIP INDICATORS as well as using NPM to manage their installation.

Removed the need to modify config files to indicate the presence of TALIB and TULIND, instead their presence is detected automatically and they are loaded when available. This bring module management down to NPM installation.

* **Other information**:

Was a bit of a thorn in switching between my active branches, and branches used for development, to ensure that all configuration files for test processes were using the correct settings.

Since there isn't any real cost to loading the indicator modules when unneeded, or a likely source of harm, and users can easily control their presence through NPM, this change seems like it will facilitate configuration for all users.

The core/talib.js and core/tulind.js scripts now detect when the libraries are missing by catching the exception thrown by require and export a null module, this is then used to validate if the indicators are available. If they are not an warning is printed to indicate that the indicators will not be available and execution resumes normally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 -Gordon Gekko
 
-Gekko is a Bitcoin TA trading and backtesting platform that connects to popular Bitcoin exchanges. It is written in javascript and runs on [nodejs](http://nodejs.org).
+Gekko is a Bitcoin TA trading and backtesting platform that connects to popular Bitcoin exchanges. It is written in JavaScript and runs on [Node.js](http://nodejs.org).
 
 *Use Gekko at your own risk.*
 
--- a/core/gekkoStream.js
+++ b/core/gekkoStream.js
@@ -3,6 +3,7 @@
 
 var Writable = require('stream').Writable;
 var _ = require('lodash');
+var async = require('async');
 
 var util = require('./util');
 var env = util.gekkoEnv();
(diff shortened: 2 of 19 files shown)


cmroche:
I've tested various combinations of having talib and tulind loaded, and trying to run indicators from those modules.

askmike:
Loving the idea, though I think stuff like this will break / not work correctly anymore:

https://github.com/askmike/gekko/blob/c86ce48d539a5f45cab2f2b14b59743ccccb8719/plugins/tradingAdvisor/baseTradingMethod.js#L10-L14

askmike:
If you also think we need to catch these situations before applying to PR let me know and I can do it as well :)

plugins/tradingAdvisor/baseTradingMethod.js
```diff
@@ -9,44 +9,17 @@ var ENV = util.gekkoEnv();
 var mode = util.gekkoMode();
 var startTime = util.getStartTime();
 
-if(config.tradingAdvisor.talib.enabled) {
```
cmroche, line 12: @askmike I believe this is the code you are referring to? I've already remove the old check and wrapped them around with the new approach. Is that what you mean?

cmroche (commented):
Didn't realize I also had to click submit review

askmike:
Apologies, you are very right! Thanks a ton :)

PR #1346, opened 2017-11-19
Sample strat that combines tulip indicators

mkondel:
Example of a strat that uses 2 tulip indicators: ADX and MACD.
Can backtest with 'tulip-multi-strat'.

### DIFF
--- a/config/strategies/tulip-adx.toml
+++ b/config/strategies/tulip-adx.toml
@@ -1,0 +1,7 @@
+historySize = 80
+optInTimePeriod = 15
+candleSize = 10
+
+[thresholds]
+up = 30
+down = 20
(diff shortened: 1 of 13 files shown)


askmike:
Great stuff, thanks!

If you feel like doing more you can add it to the docs here: https://gekko.wizb.it/docs/strategies/example_strategies.html

mkondel:
I will write an entry, sure. Do I just change this file, and the website documentation will update automatically: https://github.com/askmike/gekko/blob/develop/docs/strategies/example_strategies.md? 

My strat.js file does not follow the gekko convention of [parameters] and [thresholds]. I did it to make it work with gekkoga, another project you probably know. All strat parameters are just flat under this.settings, I also pass the same this.settings to all inner strats. Do you see any reason to have the multiple levels in that settings object?

PR #1347, opened 2017-11-19
Fix journaling mode problems on Windows BASH

cmroche:

Bugfix
Feature kind of

Journal mode is configured as 'WAL', which while faster, is not supported across all operating systems in Gekko due to file locking.

SQLITE recognizes DELETE as their default journal mode, additionally this is the most portable across operating systems, so if no journal mode is specified in the config then the default is now DELETE in handle.js -- we should strive for portability in that case.

However baseConfig.js specifies the journal mode out of the box, so I've added a script to detect if we are using windows, either natively or through bash on Windows 10 and in that case it sets the journal mode to DELETE, in all other cases it sets the journal mode to WAL.

This should mean that SQLITE works out of the box for all users now.

* **Other information**:


askmike:
@thegamecat curious as to your opinion on this :)

*(I don't have access to any windows client, but I do want Gekko to work reliably on windows)*

cmroche:
@askmike @thegamecat I tested this last night on 3 setups:

1) OSX -> It chose WAL
2) Windows Native -> It chose DELETE
3) Bash on Windows 10 -> It chose DELETE

Even though bash on Windows 10 acts like a linux system in most ways, it is still constrained to Window's locking behaviour on files and needs to be treated like a windows variant in this way. Without it, I can write, but cannot scan or reuse market data in the DB file and an SQLITE_IOERROR is printed to the output.

askmike:
Merging this, though still would love to more from windows people. @generalectric maybe?

@cmroche before this hits stable we need to update this: https://gekko.wizb.it/docs/installation/installing_gekko_on_windows_with_bash_on_windows_10.html#Edit-the-handle-js-file-taken-from-a-href-https-github-com-askmike-gekko-issues-708-issuecomment-296329300-target-_blank-here-a-

Since this PR should take care of properly setting the correct journaling mode. But let's merge this is now first :)

askmike:
TODO:

- > before this hits stable we need to update this: [DOCS]
- make sure the same config flag also gets added to `sample-config.js`.

cmroche:
I'll push those changes tonight.

PR #1348, opened 2017-11-19
Adding prettier default config file

cmroche:
Now everyone using this tool will have consistent settings automatically.

These are based on your defaults @askmike


askmike:
:+1:

PR #1349, opened 2017-11-19
Support for Post Only orders in the portfolio manager (and trade config)

cmroche:

feature

GDAX was the only exchange that had implemented support for this, it was an exchange only feature and read an option directly from the config.

Post only is not a feature of the "trader" and implemented in the portfolioManager and pushed as an option to buy and sell commands on the exchange.

The reason for going with this approach is that support for the feature is now checked on the exchange capabilities, and if the feature is requested by not supported or implemented in the exchange we can detect and throw a standard warning indicating the the feature will not be used (even though it was requested).

This functionality can then also be extended to retry posting the order in future engine ticks.

I've added implementations for GDAX, Kraken and Bitfinex. I looked at Poloniex but the node API module we use will need to be updated to add support there.

* **Other information**:

Post only can be an important option for strategies which seek to work on smaller margins, ie: PSAR by avoiding more costly taker fees when a limit request would otherwise cross the books and complete an order. Currently it aborts the order and does not attempt to repost, however in the future we can look to extended that functionality to possibly update price and attempt reposting the order if there is an interest.


askmike:
The current execution strategy is always making. Details are here: https://gekko.wizb.it/docs/introduction/scope.html#Execution-strategy (@cmroche I know you are aware of them as you implemented a lot of exchanges, putting it here for broader discussion's sake).

I want to move to a more flexible setup (where users can define speed over best price). However I don't see the value of having this configurable? Is there any situation (in the current execution model) where people would want to swtich this between other ON or OFF? Why not always force `post_only` on the exchange integration level (inside the `exchange/gdax.js` file) for now?

cmroche:
@askmike There is no reason, that I can think of, that would cause people to enable or disable this option while the bot is running, at least not right now and not without something like order book strategies (a feature I am considering for the near future). Rather my reason is about other users understanding how and if they can use it.

In general as we implement support for these types of features (also think hidden, iceburg, oco purchases) I think we should notify users when they try to use a feature that either doesn't work or is not implemented in the exchange interface.

Arguably the notification doesn't provide that much value, if you would rather go without it, I'll remove the framework and re-implement as an exchange only feature.

cmroche:
@askmike I just had an idea.. I can get the best of both worlds by adding an init phase to the exchange. This is something I was going to do anyway so I can pre-load product information from exchanges that export that information through their API, we could use it to hand an exchange config and validate the features request against the exchange. It will be much cleaner overall, so I am closing this PR.

askmike:
> This is something I was going to do anyway so I can pre-load product information from exchanges that export that information through their API, we could use it to hand an exchange config and validate the features request against the exchange.

Is this information about what is possible specific to the current user? We can also make the `getCapabilities` method async to allow for some fetching and dynamic determination of what is possible.

cmroche:
Specific to the user isn't what I had in mind, but that is easily a possibility (ie: margin trading available), I was instead thinking about using this to populate currency pair info from exchanges when supported by the exchange API so we didn't have to manually enter and update the records.

askmike:
Ah yes! It would be great if we can do that programatically. Though I doubt we can ever do that reliably on kraken given their current API.

Issue #1350, opened 2017-11-19
[request] Adding Kraken Tradepair USD/USDT


haccg:
Hello,

for different reasons I'd like to "trade" USD against USDT, so I can provide automatic conversions from and to USD/USDT.

Is it possible to manually implement the API-support for that trading pair?

Greetings

julesGoullee:
See pairs listing in /exchanges/kraken js

haccg:
Already have seen it.
Tried to add it... didn't work, no matter in what order I added them.

Doesn't retrieve any data, neither in Market Watch nor at importer

askmike:
If anyone feels like creating a PR, this is an easy one :)

as for:

> didn't work

It would be good to explain exactly what you tried and what you mean with "didn't work".

cmroche:
@haccg In addition to adding the pair to the list (order is USDT, USD), you will also need to add "USDT" to the crypto_currencies array (top of the file) and the assets list in the exchange capabilities (not the currency list, on Kraken USDT is an asset not a currency).

This should get trading working, if when live trading you see an error like "portfolio could not find asset USDT" then you need to also add an entry to the assets_without_prefix array, but only add it here if you see that error.

haccg:
First of all, thank you for your assistance!

Already tried to add as described:

Added var at crypto_currencies "USDT"

Assets area looks like that:

`assets: ['XBT', 'LTC', 'GNO', 'ICN', 'MLN', 'REP', 'XDG', 'XLM', 'XMR', 'XRP', 'ZEC', 'ETH', 'BCH', 'DASH', 'EOS', 'ETC', 'USDT'],`

And at the market area:

`	  //Tradeable against USDT
      { pair: ['USD', 'USDT'], minimalOrder: { amount: 0.01, unit: 'asset' }, precision: 2 },`

Starting Paper Trader leads to Error:

`Error: Gekko only support the currencies [CAD, EUR, GBP, JPY, USD, XBT, ETH] at Kraken (not USDT)`

Turning the order at market area around leads to nothing. It starts, but doesn't work with no error.
![watcher](https://user-images.githubusercontent.com/33813829/33100567-2caba122-cf15-11e7-9a99-e64e501ee7c8.JPG)

Any other suggestions?

cmroche:
OK, I'll take a crack at it.

cmroche:
[...]

PR #1355, opened 2017-11-22
Making kraken work with USDT

cmroche:
Yet another undocumented "feature" of kraken, USDT is halfway between the prefixless approach of alts, and prefixed approach of popular coins.


cmroche:
Resolves issue #1350

Issue #1324, opened 2017-11-15
Backtesting API gives incorrect results when handling many concurrent requests


cabloo:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Generated a lot of concurrent requests to the Gekko Backtesting API with the same query data.

* **Expected result** (what you hoped would happen)
Consistently answer the same query with the same results.

* **Actual result** (unexpected outcome)
Results are wildly different and graphs look completely changed, a lot of data appears to be missing from the graph.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
I suspect race conditions with processing so many queries at once?

askmike:
There should not be: each backtest runs in it's own child process.

Could you tell us more about what strategy you are running (using which indicators, and whether they are native or TAlib/Tulip).

cabloo:
Hi Mike,

The strategy was MACD. I'm not home so don't have the exact query I ran but
will send you the full query later tonight.

Thanks,
Zane

askmike:
Strange, if you could provide some more details like: the MACD settings, the backtest timerange and candleSize as well as the market I can try to reproduce.

cabloo:
```json
{
      "gekkoConfig": {
        "watch": {
          "exchange": "kraken",
          "currency": "USD",
          "asset": "BCH"
        },
        "paperTrader": {
          "feeMaker": 0.16,
          "feeTaker": 0.25,
          "feeUsing": "maker",
          "slippage": 0.05,
          "simulationBalance": {
            "asset": 1,
            "currency": 1500
          },
          "reportRoundtrips": true,
          "enabled": true
        },
        "tradingAdvisor": {
          "enabled": true,
          "method": "MACD",
          "candleSize": 63,
          "historySize": 10,
        },
        "MACD":  {
    "short": 3,
    "long": 5,
    "signal": 2,
    "thresholds": {
      "down": -0.0933,
      "up": 0.0379,
      "persistence": 6
    }
  },
        "backtest": {
          "daterange": {
            "from": "2017-11-07T00:41:00Z",
            "to": "2017-11-13T01:31:00Z"
          }
        },
        "performanceAnalyzer": {
          "riskFreeReturn": 2,
          "enabled": true
        },
        "valid": true
[...]

Issue #1285, opened 2017-11-03
GUI, Backtesting does not complete, Random Duration


just-ak:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[X ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Backtesting 1 Minute for 3 Months Data.

* **Expected result** (what you hoped would happen)
3Months of data processed

* **Actual result** (unexpected outcome)
25,23,13 Days.  Random Timespan covered.

![wrong duration1](https://user-images.githubusercontent.com/3976536/32357172-cf3b2690-c06c-11e7-8d04-6582d449951b.png)

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

I have worked through the problem, and created a simplified example of the  problem https://github.com/akentigernfox/node-child-process-fork .

I have worked through most of a fix and am testing it now.

askmike:
Can you tell more about your OS, your computer (type of disk) and your strategy?

And could you explain what your solution changes? I only see 3 similar scripts that ping messages between a child/parent.

just-ak:
Mike,  sorry I'm a bit slow. getting the code into it's own commit block and re-testing it.   I ended up with loads of comments everywhere and two other changes in the same commit so i'm splitting them up.

I'm running your gekko, on Digital Ocean Ubuntu 16 Droplet.  I have 3 users with 3 installs, trading 2 poloniex accounts.  BTC/VRC  BTC/VTC , and a Test Account.

Additional code pieces are for   
      1) UIConfig  put into home/.gecko/ so settings are different per user, and pull's don't affect it.
      2) Assigning the --inspect/--debug port for the child_process fork's.

askmike:
> Assigning the --inspect/--debug port for the child_process fork's.

I don't use these tools haha.

Anyway the comments around the scripts you posted indicate windows, does this mean you have the same problem on both ubuntu and windows? And where were you able to implement a fix (also both platforms)?

just-ak:
OK, reviewing on both Ubuntu & Windows.  

[...]

PR #1363, opened 2017-11-24
Fix bug with incorrect backtest results

cmroche:

BUG FIX

Often seen with concurrent API request, but can also be accomplished with a single API request for backtesting on slower PC, or PCs with high CPU usage in other processes. The gekko stream would call terminate on the child process before process.send() was able to clear it's buffer and finish sending backtest reports to the parent. This would result in varying and incorrect results, since the API generally relies on the last report to give final stats you would see incorrect information displayed about the length of the profitability period, and be looking at a snapshot from some point within the backtest process.

To prevent pre-mature shutdown of the child process a signalling scheme was setup where the child would notify the parent it was done by an IPC message, ensuring that the IPC buffer would clear before the parent will respond to acknowledge and request that the child terminate.

And additional handler was added to detect the parent process exiting prematurely and abort the child process, this helps prevent zombies in some rare cases.

* **Other information**:

This was a fun bug.

Solves isses #1324 and #1285


askmike:
Great fix!

generalectric:
very nice! before patching my gekko running on e2c strat would error out on anything over 10 at a time (socket hang ups)... after patching runs flawless 80 at a time

PR #1366, opened 2017-11-25
Fix binace import for days with no records

cmroche:
Binance is a new exchange, if you try to import records before binance was trading it would crash, now it will happily pass over those tradeless days.


ms121:
Very nice! Thanks!

PR #1369, opened 2017-11-26
Fix various issues with SQLITE 3 and import task shutdown

cmroche:

**BUG FIX**

I was able to spend some time dedicated to test the DB journaling changes and importing this weekend, I had a bug with low volume markets (particularly with Binance) where importing the default date range of BTC/POWR would:

1) Result in a lot of fragmented date ranges when scanning available datasets
2) Occasionally not show any results at all after import
3) Insert records into the database really slowly when using DELETE journalling

While these problems were largely only noticed in DELETE journaling due to the overhead, occasionally I could produce it with WAL if my CPU was kept busy.

* ** What is the new behaviour **

After doing a bit of research in the sqlite3 documentation, there were a couple of problems present and addressed

1) We did not call shutdown on the DB - Technically this isn't a huge problem as row operations are atomic in SQLITE 3, but since the nodejs module is async we have to use db.close to wait for a callback indicating the DB operations are completed and the DB has flushed.

2) We did not wait for the shutdown of the DB to complete. Import was calling process.exit directly. To fix this I've added a callback function to finalize which waits on the signal that the DB is flushed. Once finalization is complete gekkostream will now shutdown the process by sending the IPC request to exit like with backtesting. This will give is a singular and consistent approach to handling shutdown and ensure that the DB is ready.

3) We were not properly using transactions for inserts. The code did not use transactions at all, in this case the default behaviour for SQLITE is to automatically treat every insert as atomic which adds a lot of overhead in all journaling modes, though in WAL journalling was still fairly performant. I added calls to db.serialize and BEGIN/COMMIT for every candle batch, this made staggering improvements to insert speed (15 minutes in one test, to 5 seconds).

4) The use of _.defer(writeCandles) was potentially causing the last candle write to attempt to happen after the DB was closed. I removed the call here and instead use a counter to trigger a batch write in batches of 1000, as well as an explicit call in finalize to flush pending candles. With transactions the write time is negligible.

5) While DELETE is the default journal mode in SQLITE, I changed the mode on windows to PERSIST. The causes the journal files to stick around, but our DBs are very small and the journal files even smaller (measured in KB), this added to insert performance.

6) SQLITE generally recommends setting FULL synch mode for non-WAL journaling, this helps ensure the integrity of the database should the import process crash, WAL is by nature more robust and recommended to leave synch as NORMAL when it is being used. This is now respected.

* **Other information**:

I've done a fairly extensive amount of testing on these changes over the last 48 hours, on both OSX and Windows, using both journal modes on each system, importing and backtesting from different exchanges, and validating DB integrity by direct integrity checks through SQLITE DB BROWSER.

I have not added shutdown procedure for Mongo or Postgres. They may need to be checked in the future, however since they are not embedded solutions they are  likely to be more persistent and not be prone to the above problems anyway.

### DIFF
--- a/core/gekkoStream.js
+++ b/core/gekkoStream.js
@@ -40,7 +40,6 @@ Gekko.prototype.finalize = function() {
 }
 
 Gekko.prototype.shutdown = function() {
-  console.log("Finalizing gekko stream");
   async.eachSeries(
     this.candleConsumers,
     function(c, callback) {
--- a/core/gekkoStream.js
+++ b/core/gekkoStream.js
@@ -40,10 +40,12 @@ Gekko.prototype.finalize = function() {
 }
 
 Gekko.prototype.shutdown = function() {
+  console.log("Finalizing gekko stream");
   async.eachSeries(
     this.candleConsumers,
     function(c, callback) {
       if (c.finalize) c.finalize(callback);
+      else callback();
     },
     function() {
       // If we are a child process, we signal to the parent to kill the child once it is done
--- a/core/gekkoStream.js
+++ b/core/gekkoStream.js
@@ -3,6 +3,7 @@
 
 var Writable = require('stream').Writable;
 var _ = require('lodash');
+var async = require('async');
 
 var util = require('./util');
 var env = util.gekkoEnv();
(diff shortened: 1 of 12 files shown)


askmike:
will test and merge asap!

cmroche:
Should fix #1375, #1373, #1393

askmike:
:+1:

PR #1372, opened 2017-11-26
Correctly catch Promise exception when trying to open browser

rpoddighe:
Bug fix

UnhandledPromiseRejectionWarning when automatically opening UI in browser fails

Promise rejction is handled and a an error message is printed.

* **Other information**:
For some reason the opn module fails to open my browser, and try/catch doesn't catch the error since opn() returns a Promise. Just wanted to point that out, feel free to change the console statement to your taste :)


askmike:
Ah very nice one, thanks!

PR #1381, opened 2017-11-29
Fixing use of USDT with kraken

### LINKED ISSUE #1350: [request] Adding Kraken Tradepair USD/USDT

cmroche:

Bugfix

getTicker doesn't respect the proper USDT pair type. Also it turn out that kraken accepts "USDTZUSD" and "USDTUSD" as the pair for the API query strings (always seems to return USDTZUSD though). Still an odd duck, but simplified the bug fix by letting me move it to a more global scope.

Fixed that stuff

* **Other information**:

We'll still need to validate the return format of getPortfolio since the wallet pairs were again formatted differently from the input pairs, another fix may yet be required for that after testing in live.

Should fix #1350


askmike:
Merging this because #1411 is currently pending.

PR #1396, opened 2017-12-02
Improvements to kraken retry handler

cmroche:
Fixed the 520 code detection (was using the wrong result string for testing previously)
Added 504 and 502 as these also happen with Kraken's recent issues (and are retryable)

Should help with #1350 also this will help with the import process suddenly stopping for people trying to import from Kraken.


askmike:
:+1: rest of discussion regarding retries is in #1411.

PR #1400, opened 2017-12-02
Make the Dockerfile leaner

cmroche:

Feature

Docker builds result in a final image of about 860MB, layer size of 298MB

Docker builds result in a final image of about 760MB, layer size of 198MB

* **Other information**:

Accomplished by a few minor tweaks:

1) Add --production flag to all installs
2) Combine RUN commands to reduce layer count
3) Add npm cache clean at the end of every install (within the same layer)

Clean test of build (by deleting all images first), run, import and paper trade in the image.


askmike:
:+1:

Issue #1106, opened 2017-09-10
Poloniex importing not working


roneetshaw:
I'm relatively new to gekko trading bot, I just installed the bot yesterday after reading the documentation and seeing the installation video. I followed the exact step to get three month historical data from Poloniex so that I backtest the strategies. 
When I start the import from Poloniex, I instantly get import done as soon as I start the import and I don't get any data, I know it's not able to pull any result from Poloniex. However, when I repeat the same steps with Gdax, it works fine.
I also see the db files in `\history\` where poloniex_0.1.db does get created but it's only of 4KB. I'm using the stable branch of gekko. 
Do I have to config anything to get Poloniex start importing data?

oktapodia:
Which pair did you want to import?

roneetshaw:
USDT/BTC (FROM : 2017-06-10 10:59 TO: 2017-09-10 10:59)

roneetshaw:
![image](https://user-images.githubusercontent.com/19503905/30248443-9cc35882-9645-11e7-86c6-705032c02b10.png)
This I get everytime. Poloneix one doesn't complete ever

matthew-restaff:
similar issue here. import from poloniex does not work. bittrex is not available as an option.

logicappsource:
yes same issue here

askmike:
For anyone who has this issue: I am pretty sure poloniex has banned your IP address. See #1055 for defails. If that is not the case please reopen and post the output of following the instructions in that issue.

roneetshaw:
https://pastebin.com/cppNVFeK is broken

askmike:
What happens when you go to this URL? https://poloniex.com/public?command=returnTradeHistory&currencyPair=BTC_NXT&start=1410158341&end=1410499372 (Do you see a captcha or not?)

roneetshaw:
Yes,I get a "I'm not a robot window".
Trying to fix. Thanks for the help!

battlezone100:
I'm having the same issue where I can't reliably import data, regardless of currency pair.
I just get the message "Import done" yet it only imported like 5 minutes to a week or anything in between. It's a nightmare to get a few months of data imported.

I do not see any captcha when visiting the above page.

ghost:
Im having this issue on binance

Issue #869, opened 2017-07-21
Problem with Gekko data importing


vlad3609:
Good afternoon everyone,

First of all big thanks for such a great tool!

Now to my problem, if anyone has the solution, I'd really appreciate your help. 
It's my second computer, where I'm installing gekko. The system is Macintosh, MacOs Sierra, MacBook Pro. Though it doesn't seem to be any problems at the first one, there's at the second. 
In order to install gekko, firstly I preinstalled nodejs and then followed certain commands for the app installation (like in your youtube video). 
Then I'm trying to import data from polonium and here is the result. Actually it's the same with all the options.

<img width="1041" alt="1" src="https://user-images.githubusercontent.com/30337216/28453894-ca7d8854-6e01-11e7-8192-48acd7588d5f.png">
<img width="492" alt="2" src="https://user-images.githubusercontent.com/30337216/28453895-caa759d6-6e01-11e7-80c2-c6a0a2698c1a.png">
<img width="1060" alt="3" src="https://user-images.githubusercontent.com/30337216/28453896-caa874ba-6e01-11e7-9092-c3045980bf88.png">

My Firewall is turned off.

I'd really appreciate your help,
Have a good day, 
Cheers!

JohannesFerner:
What does the Node console print out?

vlad3609:
<img width="572" alt="screen shot 2017-07-21 at 17 11 22" src="https://user-images.githubusercontent.com/30337216/28467220-d2ca642e-6e37-11e7-8cde-91dc67d3dff6.png">

vlad3609:
Any ideas?

askmike:
Very strange..

Can you open this file: https://github.com/askmike/gekko/blob/stable/web/routes/baseConfig.js

and set line 9 and 10 like so:

    config.silent = false;
    config.debug = true;

and rerun and post output?

vlad3609:
Sure, here it is:
<img width="547" alt="screen shot 2017-07-22 at 20 42 10" src="https://user-images.githubusercontent.com/30337216/28493363-4979616a-6f1e-11e7-8766-685ce22cab32.png">

Well, it doesn't look very different.
I tried different browsers, no difference.

raffus:
Are you running this native or through a VM machine?

Can you open this URL from your browser/console (or VM)?
https://poloniex.com/public?command=returnTicker

vlad3609:
No problems in accessing it (other than Google captcha). 
[...]

PR #1411, opened 2017-12-04
[KRAKEN] Normalizing the approach for retry handling

cmroche:

Feature, will probably fix some bugs though

Currently retry handling is implemented by the individual exchanges, and so isn't very consistent. Depending on the implementation, some exchanges have irrecoverable errors to solve issues like API errors causing orders cancels to never end, others impement them in different functions. In Kraken, for instance, there was no retry handling on getTicker, but some other exchanges do implement this. 

Also recent work (6 months ago) @askmike started to add retry handling directly in the portfolioManager. This is implemented fairly easily using the "retry" package which comes with feature not otherwise handled by our per exchange implementation, such as limiting the number of retries and exponential backoff of the retry delay.

In this PR I extended that functionality to all exchange functions, and was able to drop retry handling from kraken completely.

This approach is also fully compatible with exchange that currently have their own retry handling, so there is no need to update all the exchange implementations until someone is ready to test them.

This should help simplify the implementation and debugging of exchanges going forward.

* **Other information**:

Kraken's service is in such a state is disrepair right now that it was a convenient place to test the retry handlers worked correctly. Through various actions, importing, paper trading the recoverable errors are caught and dispatched through the retry handler as expected.

### DIFF
--- a/plugins/trader/portfolioManager.js
+++ b/plugins/trader/portfolioManager.js
@@ -104,7 +104,7 @@ Manager.prototype.setPortfolio = function(callback) {
 
   }.bind(this);
 
-  util.retry(this.exchange.getPortfolio, set);
+  this.exchange.getPortfolio(set);
 };
 
 Manager.prototype.setFee = function(callback) {
@@ -117,7 +117,7 @@ Manager.prototype.setFee = function(callback) {
     if(_.isFunction(callback))
       callback();
   }.bind(this);
-  util.retry(this.exchange.getFee, set);
+  this.exchange.getFee(set);
 };
 
 Manager.prototype.setTicker = function(callback) {
@@ -130,7 +130,7 @@ Manager.prototype.setTicker = function(callback) {
     if(_.isFunction(callback))
       callback();
   }.bind(this);
-  util.retry(this.exchange.getTicker, set);
+  this.exchange.getTicker(set);
 };
 
 // return the [fund] based on the data we have in memory
--- a/core/error.js
+++ b/core/error.js
@@ -1,0 +1,25 @@
+const _ = require('lodash');
+
+let RetryError = function(message) {
+    _.bindAll(this);
+
+    this.name = "RetryError";
+    this.message = message;
+}
+
+RetryError.prototype = new Error();
+
+let AbortError = function(message) {
+    _.bindAll(this);
+
+    this.name = "AbortError";
+    this.message = message;
+}
+
+AbortError.prototype = new Error();
+
+module.exports = {
+    'RetryError': RetryError,
+    'AbortError': AbortError
+};
+
--- a/core/util.js
+++ b/core/util.js
@@ -5,6 +5,7 @@ var fs = require('fs');
 var semver = require('semver');
 var program = require('commander');
 var retry = require('retry');
+var Errors = require('./error');
 
 var startTime = moment();
 
@@ -15,6 +16,24 @@ var _gekkoMode = false;
 var _gekkoEnv = false;
 
 var _args = false;
+
+var retryHelper = function(fn, options, callback) {
+  var operation = retry.operation(options);
+  operation.attempt(function(currentAttempt) {
+    fn(function(err, result) {
+      if (!(err instanceof Errors.AbortError) && operation.retry(err)) {
+        return;
+      }
+
+      if (err) {
+        console.log('About to throw an error along ' + operation.mainError());
+        console.log(err);
+      }
+
+      callback(err ? operation.mainError() : null, result);
+    });
+  });
+}
 
 // helper functions
 var util = {
@@ -163,23 +182,18 @@ var util = {
     return startTime;
   },
   retry: function(fn, callback) {
-    var operation = retry.operation({
+    var operation = {
       retries: 5,
       factor: 1.2,
       minTimeout: 1 * 1000,
       maxTimeout: 3 * 1000
-    });
+    };
  
-    operation.attempt(function(currentAttempt) {
-      fn(function(err, result) {
-        if (operation.retry(err)) {
-          return;
-        }
-
-        callback(err ? operation.mainError() : null, result);
-      });
-    });
-  }
+    retryHelper(fn, options, callback);
+  },
+  retryCustom: function(options, fn, callback) {
+    retryHelper(fn, options, callback);
+  },
 }
 
 // NOTE: those options are only used
--- a/core/util.js
+++ b/core/util.js
@@ -25,12 +25,7 @@ var retryHelper = function(fn, options, callback) {
         return;
       }
 
-      if (err) {
-        console.log('About to throw an error along ' + operation.mainError());
-        console.log(err);
-      }
-
-      callback(err ? operation.mainError() : null, result);
+      callback(err ? err.message : null, result);
     });
   });
 }
--- a/exchanges/kraken.js
+++ b/exchanges/kraken.js
@@ -106,7 +106,7 @@ var retryCritical = {
 var retryForever = {
   forever: true,
   factor: 1.2,
-  minTimeour: 10,
+  minTimeout: 10,
   maxTimeout: 30
 };
 
@@ -125,13 +125,15 @@ Trader.prototype.processError = function(funcName, error, alwaysRecoverable) {
 };
 
 Trader.prototype.handleResponse = function(funcName, error, body, callback) {
-  if(_.isEmpty(body) || _.isEmpty(body.result))
-    err = new Error('NO DATA WAS RETURNED');
+  if(!error) {
+    if(_.isEmpty(body) || !body.result)
+      error = new Error('NO DATA WAS RETURNED');
 
-  else if(!_.isEmpty(body.error))
-    err = new Error(body.error);
+    else if(!_.isEmpty(body.error))
+      error = new Error(body.error);
+  }
 
-  return callback(this.processError(funcName, err, true), body);
+  return callback(this.processError(funcName, error, true), body);
 };
 
 Trader.prototype.getTrades = function(since, callback, descending) {
--- a/exchanges/kraken.js
+++ b/exchanges/kraken.js
@@ -116,11 +116,11 @@ Trader.prototype.processError = function(funcName, error, alwaysRecoverable) {
   if (!error) return undefined;
 
   if (!alwaysRecoverable && !error.message.match(recoverableErrors)) {
-    log.error(`[kraken.js] ${funcName} returned an irrecoverable error: ${error.message}`);
+    log.error(`[kraken.js] (${funcName}) returned an irrecoverable error: ${error.message}`);
     return new Errors.AbortError(error.message);
   }
 
-  log.info(`[kraken.js] ${funcName} returned an error, retrying: ${error.message}`);
+  log.info(`[kraken.js] (${funcName}) returned an error, retrying: ${error.message}`);
   return new Errors.RetryError(error.message);
 };
 
@@ -138,6 +138,8 @@ Trader.prototype.getTrades = function(since, callback, descending) {
   var startTs = since ? moment(since).valueOf() : null;
 
   var process = function(err, trades) {
+    if (err) return callback(err);
+
     var parsedTrades = [];
     _.each(trades.result[this.pair], function(trade) {
       // Even when you supply 'since' you can still get more trades than you asked for, it needs to be filtered
@@ -157,7 +159,7 @@ Trader.prototype.getTrades = function(since, callback, descending) {
       callback(undefined, parsedTrades);
   };
 
-  var reqData = {
+  let reqData = {
     pair: this.pair
   };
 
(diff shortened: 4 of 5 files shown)


askmike:
Great stuff, generally:

- I want exchange libraries to be responsible for retrying everything

askmike:
Great stuff!

Apologies for the text below which is about the context of the PR, not the contnt. The following is not a direct response to this PR, but about the direction I want Gekko to go in:

- I want exchange libraries to be responsible for retrying everything that can be retried, because whether there was a `recoverable` or a `non recoverable` should be something decided at this level instead: if the error is recoverable: the lib should retry*, if it is not it should upstream the error (things like: invalid API key, not enough balance). Throughout the exchange API this is sometimes done, but not consistently.
- I love the idea of exposing helper functions to actually implement this on the exchange level, the exchange wrapper can have the unique checks in place per exchange to determine whether something is recoverable or not. I love how you are upstreaming either a recoverable or non recoverable error as well.

*After X retries (100 or a 1000 or so) it should stop retrying and upstream the error, not sure how to handle this when exchanges are temporary down. But not sure if trying to `buy` based on an advice from a day ago is still a smart thing to do.

Most things in this PR are already in the same line as above.

core/budfox/marketFetcher.js
```diff
@@ -72,7 +72,7 @@ Fetcher.prototype._fetch = function(since) {
   if(++this.tries >= this.limit)
     return;
 
-  this.watcher.getTrades(since, this.processTrades, false);
+  util.retryForever((callback) => this.watcher.getTrades(since, callback, false), this.processTrades);
```
askmike: I'd rather keep this on the exchange level, if you are retrying inside kraken.js already, what is the need for this?

core/tools/dataStitcher.js
```diff
@@ -186,18 +186,20 @@ Stitcher.prototype.checkExchangeTrades = function(since, next) {
 
   var watcher = new DataProvider(exchangeConfig);
 
-  watcher.getTrades(since, function(e, d) {
-    if(_.isEmpty(d))
-      return util.die(
-        `Gekko tried to retrieve data since ${since.format('YYYY-MM-DD HH:mm:ss')}, however
-        ${provider} did not return any trades.`
-      );
+  util.retryForever((callback) => {
```
askmike: Same as above.

plugins/trader/portfolioManager.js
```diff
@@ -214,7 +214,8 @@ Manager.prototype.buy = function(amount, price) {
     'price:',
     price
   );
-  this.exchange.buy(amount, price, this.noteOrder);
+
+  util.retryForever((callback) => this.exchange.buy(amount, price, callback), this.noteOrder);
```
askmike: see above.

plugins/trader/portfolioManager.js
```diff
@@ -245,7 +246,8 @@ Manager.prototype.sell = function(amount, price) {
     'price:',
     price
   );
-  this.exchange.sell(amount, price, this.noteOrder);
+
+  util.retryForever((callback) => this.exchange.sell(amount, price, callback), this.noteOrder);
```
askmike: see above.

plugins/trader/portfolioManager.js
```diff
@@ -261,13 +263,15 @@ Manager.prototype.noteOrder = function(err, order) {
...
 
 Manager.prototype.cancelLastOrder = function(done) {
-  this.exchange.cancelOrder(_.last(this.orders), alreadyFilled => {
-    if(alreadyFilled)
-      return this.relayOrder(done);
-
-    this.orders = [];
-    done();
-  });
+  util.retryForever((callback) => {
```
askmike: see above.

plugins/trader/portfolioManager.js
```diff
@@ -363,7 +367,7 @@ Manager.prototype.relayOrder = function(done) {
 
   var getOrders = _.map(
     this.orders,
-    order => next => this.exchange.getOrder(order, next)
+    order => next => util.retryForever((callback) => this.exchange.getOrder(order, callback), next)
```
askmike: see above.

askmike (commented):
See nits in combination with the story above.

askmike:
Maybe there is a difference in retrying at exchange level and at the portfolio level which I am not seeing, in that case please let me know why you did :)

cmroche:
@askmike As you may have noticed over the weekend, there were some issue with the implementation of error and retry handling within kraken.js, when I think about the nature of the issue (ie: When should we retry, who decides we should retry) these are the criteria that come to mind:

1) The exchange implementation should be easy to keep consistent through all implementations, and reduce the amount of copy and paste code. This will reduce bugs in other feature implemented in the engine as it is generally difficult to test multiple exchanges particularly in a live trade scenario.

2) The exchange knows when an error is not recoverable, and we should abort the retry (ie: data errors like an order doesn't exist, or bad parameters for addOrder) retrying would be problematic in this case.

3) The upstream calls determine if it want's a request to be retried. For example there is a cancel order that is possibly called every minute to adjust the entry, we don't need to keep retrying if this fails, wait a minute and let it try again. Implementing a retry without blocking the time could result in a run-away retry problem spamming the API.

The goal of this system was to meet those 3 criteria. The exchange tells us when to retry, the engine provides the implementation, reduces copy-paste code in the exchange function and reducing error. It opens the door the the engine deciding that a retry isn't appropriate (it: get ticker updating every frame could be implemented without the retry since it is already on a timer).

That was my thinking :)

askmike:
I agree with the first 2 points, but I don't understand the third one:

> The upstream calls determine if it want's a request to be retried.

That makes sense, but right now the upstream code just retries infinitely:

- here: https://github.com/askmike/gekko/pull/1411/files#diff-4dd24913ae3c25be2a5914d44f799c4aR75
- and here: https://github.com/askmike/gekko/pull/1411/files#diff-5f08f9bba973c819ee0352b5bf705834R189

In case getTrades fails (which on Kraken has a big change of happening but definitely not on exchanges like bitfinex/poloniex), there is most likely something else wrong. See #869 and #1106 as examples to this.

cmroche:
@askmike One of the cancel orders in portfolioManager doesn't have a retry at all, I'm pretty sure it would survive. But you are right, getTrades has other issues. It's a longer term goal that it could be addressed, if this was in place.

The main argument IMHO is really about reducing bugs. I personally feel that exchange implementations will be more reliable long term with this approach since it simplifies the implementation, reduces copy and paste errors, etc. If you look at the current exchanges they don't all do retry in the same way, or even on the same functions. No one person is likely to refactor all exchange files if it can be avoided because there is risk in the amount of testing necessary to be reliable. This approach makes all improvements, present and future, available reliably to all exchange implementations.

If you feel that this is not a priority, then it would probably make sense to move the util.retry calls into the exchange.

askmike:
I am 100% with you on decreasing bugs as the main priority! Though I think changing the behaviour of code that implements getTrades (by adding infinite retries) might possibly add new bugs on exchanges where getTrades not working means something being really wrong (IP being banned, as soon as they change their API, etc).

> No one person is likely to refactor all exchange files if it can be avoided because there is risk in the amount of testing necessary to be reliable. This approach makes all improvements, present and future, available reliably to all exchange implementations.

I agree that no one is going to go over all exchanges all at once and fix everything, but just infinitely retrying all exchanges (even the ones that are considered very stable like Poloniex, Bitstamp and Bitfinex, a few of the most used exchanges) is not a great solution to this imo.

> One of the cancel orders in portfolioManager doesn't have a retry at all

This is the part where I think we should think about how we want `cancelOrder` to work in the future, and slowly build towards that (without having to fix all exchange implementations). Specifically looking at cancelOrder:

> ## cancelOrder
>
> ```this.exchange.cancelOrder(order, callback);```
>
> The order will be something that the manager previously received via the sell or buy methods. The callback should have the parameter err.

Right now the docs (=spec) does not define what the err should be, but we can change this right now into: the exchange implementation should retry unless the error is not recoverable (order was filled, order does not exist, etc). In which case the `err` hitting portfolioManager indicates a problem where retrying does not even make sense.

As soon as there is any issue with the `cancelOrder` implementation in another exchange we can patch it up. I think that would already be an improvement from the current situation. What do you think?

cmroche:
@askmike Ok, I think I misunderstood your previous points. The issue is about the infinite retry, yes you are likely right. I did it this way to preserve what I saw in the exchange implementations, but fully agree that they should perhaps be something less intense. 3 retries seems a bit short, but perhaps 10 retries with a max of 30 seconds. This falloff I think is very safe from being banned.

Additionally can use the helper still with the error types to give better control, and move these functions into the exchange itself, but that sort of implies doing the same thing for changes you made in the portfolioManager a few months back, we want all the retries on exchange functions in the exchange in that case right?

This is definitely still an improvement I'll make the changes once I have your feedback.

askmike:
> but that sort of implies doing the same thing for changes you made in the portfolioManager a few months back

Ah! This is probably where the confusion came from: I shouldn't have add retries here, since it goes against the design (we both agree on). I'll try to correct this soon.

> we want all the retries on exchange functions in the exchange in that case right?

Yep!

> This is definitely still an improvement I'll make the changes once I have your feedback.

A complete PR would be great, but I am already very happy about all of these changes (just not infi retry in getTrades consumers). But I'll gladly wait for more :) Your efforts are extremely appreciated!

cmroche:
FYI, I am working on this, but it may take a day or two to complete and test what I had in mind.

cmroche:
@askmike 

So this is what it looks like if we move the retry handling directly into the exchange. In this case we still use the helper function for handle retry, which gives us a nicer implementation of retry logic, but allows us to specify the retry configuration at exchange level since ideals are bound to vary from one implementation to another, and also preserves the upstreaming of irrecoverable and recoverable errors.

If you are happy with these changes the next step will be for me to update the documentation to clarify the expectation of retry handling on the exchange implementation, and reference this implementation as a template for how to handle it.j

I'd likely also take this myself into the Bitfnex and Binance exchanges since I use those regularly. But only after we've tested a bit in our stress test case (Kraken) :)

As a bonus I also cleaned up a lot of the copy and paste code into a handler function, and standardized the formatting of the error messages. It look much nicer on the output when errors do occur.

askmike:
> So this is what it looks like if we move the retry handling directly into the exchange. In this case we still use the helper function for handle retry, which gives us a nicer implementation of retry logic, but allows us to specify the retry configuration at exchange level since ideals are bound to vary from one implementation to another, and also preserves the upstreaming of irrecoverable and recoverable errors.

:+1::+1::+1::+1:

> If you are happy with these changes the next step will be for me to update the documentation to clarify the expectation of retry handling on the exchange implementation, and reference this implementation as a template for how to handle it.

I am very happy with the overall changes, however there were a few small issues I found while testing. I have fixed them here: https://github.com/cmroche/gekko/pull/1

If you agree & merge them they *should* show up here.

cmroche:
Changes look great thanks @askmike, thanks.

I will work on updating the documentation tonight and send a second PR for those changes probably over the weekend.

wupeng1211:
@askmike  gekko/core/util.js:187
    retryHelper(fn, options, callback);
                    ^

ReferenceError: options is not defined
    at Object.retry (/home/ka/gekko/core/util.js:187:21)
    at Mailer.mail (/home/ka/gekko/plugins/mailer.js:84:8)
    at Mailer.bound [as mail] (/home/ka/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Mailer.setupMail (/home/ka/gekko/plugins/mailer.js:42:12)
    at Mailer.setup (/home/ka/gekko/plugins/mailer.js:79:15)
    at Mailer.bound (/home/ka/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at new Mailer (/home/ka/gekko/plugins/mailer.js:15:8)
    at load (/home/ka/gekko/core/pluginUtil.js:95:22)
    at /home/ka/gekko/node_modules/async/dist/async.js:1156:9
    at replenish (/home/ka/gekko/node_modules/async/dist/async.js:1030:17)
    at iterateeCallback (/home/ka/gekko/node_modules/async/dist/async.js:1015:17)
    at /home/ka/gekko/node_modules/async/dist/async.js:988:16
    at /home/ka/gekko/node_modules/async/dist/async.js:1158:13
    at load (/home/ka/gekko/core/pluginUtil.js:68:14)
    at /home/ka/gekko/node_modules/async/dist/async.js:1156:9
    at replenish (/home/ka/gekko/node_modules/async/dist/async.js:1030:17)
    at iterateeCallback (/home/ka/gekko/node_modules/async/dist/async.js:1015:17)
    at /home/ka/gekko/node_modules/async/dist/async.js:988:16
    at /home/ka/gekko/node_modules/async/dist/async.js:1158:13
    at load (/home/ka/gekko/core/pluginUtil.js:68:14)
    at /home/ka/gekko/node_modules/async/dist/async.js:1156:9
    at replenish (/home/ka/gekko/node_modules/async/dist/async.js:1030:17)

tdsticks:
See my solution in #1465, hope that works for you.

PR #1421, opened 2017-12-05
Upgrade moment & semver dependencies

askmike:
This fixes:

screen shot 2017-12-05 at 14 24 28

Which in Gekkos case are definitely not critical, but good to update nonetheless.


askmike:
I am very confident these updates do not have breaking compatibility issues.

PR #1425, opened 2017-12-05
Fix retry logic in kraken

### LINKED ISSUE #1424: Recovering from errors
Right now Kraken suffers timeout issues and Gekko suffers trying to get the history data and call the API after that and throw an "irrecoverable error" and shuts the script.

Could it retry multiple times instead by waiting for a longer interval each time ?

cmroche:
Small error where I inverted the wrong compound of a conditional in a previous cleanup of the retry handler. This will fix issues where we report all errors as irrecoverable.

Fixes #1424 and #1350


cmroche:
No, this is wrong ....

PR #1426, opened 2017-12-05
Fixing kraken timeouts, getTicker must always succeed

### LINKED ISSUE #1350: [request] Adding Kraken Tradepair USD/USDT

cmroche:
If you look at portfolioManager, the call it makes to getTicker must always succeed, if it were to fail on any request then the this.ticker would be null and the engine will crash on access.

Should resolve #1350 and probably #1424


askmike:
I am closing this since #1411 contained a safer retry mechanism for the ticker call, making this redundant. Please reopen if that's not the case!

PR #1427, opened 2017-12-06
Binance, fix a case where small purchases could have small rounding errors

cmroche:
This change addresses a case where an attempt to trade the minimal amount may result in small rounding errors which cause the API to reject the order. Ensures that orders are always for the minimum amount.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 -Gordon Gekko
 
-Gekko is a Bitcoin TA trading and backtesting platform that connects to popular Bitcoin exchanges. It is written in javascript and runs on [nodejs](http://nodejs.org).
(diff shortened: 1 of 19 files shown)


ms121:
Ok, it is now throwing the following;

2017-12-06 02:52:27 (DEBUG):	[binance.js] entering "setOrder" callback after api call, err: { code: -1105, msg: 'Parameter \'quantity\' was was empty.' }  data: { code: -1105, msg: 'Parameter \'quantity\' was was empty.' }
2017-12-06 02:52:27 (ERROR):	[binance.js] unable to buy { code: -1105, msg: 'Parameter \'quantity\' was was empty.' }

Thanks again!

cmroche:
Typo is fixed

ms121:
@cmroche Seems to be fixed.
Thanks again for the updates!

askmike:
:+1:

PR #1428, opened 2017-12-06
Fix sqlite buffer when using the market leecher

cmroche:

Bugfix

When launching a paper trader from the UI, a market leech is used in conjunction with a watch/candlewriter to populate trade data to the papertrader. Buffering in the SQLITE adaptor changed to being batched (as opposed to time) which menas 1000 candles must arrive before the data is written, and processable by the paper trader.

This does not appear to impact a papertrader started from the command line.

The SQLITE adaptor will now set a flag to know if it should buffer based on the presence of the market leech. If a market leech is in the config, it will stop buffering the candle writes. They are being received 1 per minute in this case and so buffering isn't required.


cmroche:
Nope... I got this backwards, which means it isn't working

cmroche:
Should fix #1429 , #1430

askmike:
> This does not appear to impact a papertrader started from the command line.

It doesn't, the commandline version pipes a budfox (exchange poller) straight into your strat :)

The UI does it like does so you can setup 10 paper traders on bitstamp USD/BTC without hammering their API with 10 different pollers.

askmike:
@cmroche it seems this only limits buffering to imports, please let me know if I am wrong :)

cmroche:
@askmike You are right. Looking at the scenario, imports are the special case, everything else should probably write without the buffering since the data stream will be much slower.

Detecting the leech case was problematic, because the candleWriter wasn't on the same process as the leech.

askmike:
> Detecting the leech case was problematic, because the candleWriter wasn't on the same process as the leech.

@cmroche sorry you had to figure this out the hard way. These new architecture changes I have been doing to smooth the UI experience (the UI doesn't map 1-to-1 to a single Gekko process that does everything) are not really documented anywhere.

Nonetheless I like the approach you finally took: any gekko (that is not importing) should write candles as soon as possible. Whether there is a "leecher" polling the database or not. This way a "market watcher" can be used for people who want to host a charting website or similar.

Issue #1439, opened 2017-12-08
Import Data Problem Error


jp054:
Hi!

When I import Data I have this error :

![capture](https://user-images.githubusercontent.com/34319814/33776251-10a42b96-dc41-11e7-861a-cb7b507c8aed.PNG)

![capture2](https://user-images.githubusercontent.com/34319814/33776297-30fdf692-dc41-11e7-87cd-fdf56f1edc37.PNG)

Thanks for your answers

cmroche:
Kraken's API is not reliable right now. Basically it failed after multiple subsequent queries and gave up. Try again when the service is a bit more responsive.

wakandan:
I can't even see the import data running like before. It just says that the import is finished

wakandan:
seems like only data from bitfinex can be imported at the moment, other exchanges importer just hangs

jvandenaardweg:
Got a similar error when trying to import GDAX data. Got that error immediately when i started the import.

```
Import 530332521897539 started
  --> POST /api/import 200 8ms 145b
/usr/src/app/exchanges/gdax.js:248
                if (moment.utc(last.time) < moment.utc(since)) {
                                   ^
2017-12-09T10:41:36.285089800Z 
TypeError: Cannot read property 'time' of undefined
    at .<anonymous> (/usr/src/app/exchanges/gdax.js:248:36)
    at Request._callback (/usr/src/app/node_modules/gdax/lib/clients/public.js:45:7)
    at Request.self.callback (/usr/src/app/node_modules/gdax/node_modules/request/request.js:187:22)
    at emitTwo (events.js:106:13)
    at Request.emit (events.js:191:7)
    at Request.<anonymous> (/usr/src/app/node_modules/gdax/node_modules/request/request.js:1044:10)
    at emitOne (events.js:96:13)
    at Request.emit (events.js:188:7)
    at IncomingMessage.<anonymous> (/usr/src/app/node_modules/gdax/node_modules/request/request.js:965:12)
    at emitNone (events.js:91:20)
RECEIVED ERROR IN IMPORT 530332521897539
Child process has died.
```

Bitfinex is working as stated above.

cmroche:
@wakandan @jvandenaardweg There is a fix for GDAX issues in the develop branch, please sync and try there.

PR #1440, opened 2017-12-09
[Kraken] Adding back response code 525

cmroche:
This was accidentally dropped in all the kraken work and refactors.

Should address #1439


wakandan:
does this mean that gekko could not import data from the exchange because the exchange returns error?

cmroche:
@wakandan Yes, hitting this error would have bailed out the import.

wakandan:
does this apply to only kraken or poloniex as well?

cmroche:
Just Kraken

askmike:
:+1:

PR #1365, opened 2017-11-24
Wrapper for ccxt exchanges

Yoyae:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature

* **What is the new behavior (if this is a feature change)?**
Added ccxt wrapper that add all exchanges supported by ccxt (https://github.com/ccxt/ccxt)

* **Other information**:
I've not been able to add the wrapper without modifying some core files but I kept the number of changes as low as possible.

As discussed in #1171 , ccxt is working in a different way. This a wrapper to make it work wit Gekko.

PR #1441, opened 2017-12-09
Don't die if the order fails, just print an error [Discussion]

cmroche:

Bugfix

If an order fails due to insufficient funds, the trade bot dies.

The tradebot drops the request and prints an error instead noting why the order could not be completed

* **Other information**:

Refer to the latest comments in #1350 . In this case on Kraken, it isn't possible the query the available balance, only the balance available. So any other order pending could lock funds and may cause a situation where the order would not have enough funds to clear.

It is not abnormal for manual order entries on a bot that is running, especially is keepAsset is specified. We would expect it to recover once the order is cleared and funds are again available, or in the next action when the balance is updated.

In the case of #1350 the error was only recoverable by user intervention, but this isn't representative of all cases.

Looking at the code we seem more than capable of recovering from this condition, though we will miss the current action.

On the other hand addOrder would not stop execution of the engine if the API keys were invalid. But this would be caught in getPortfolio, where irrecoverable errors actual do require execution to stop. Are there conditions where getPortfolio would suceed and addOrder would fail where we should end execution?

Thoughts?


erasmuswill:
I totally agree with this logic. This idea could really help with Luno implementation with #1365 because it constatntly fails with either an insufficient balance or an orderNotFound error. I've been needing to manually restart every single time until I get teh time to fix it.

EDIT: Why did you delete it? My interface reloaded after original comment and showed it as deleted.

PR #1446, opened 2017-12-10
Git-Ignore users config files

Ali1:
I store my config files as e.g. config-btc.js so this commit helps avoid conflict.


askmike:
:+1:

greenbigfrog:
Shouldn't the wildcard indicator rather be `config*.js`, since some people might prefer something like `config_btc-usd.js`?

askmike:
Fair point. I am really up for either, if you make a PR for extending the wildcard like that I have no problems accepting that :)

Issue #937, opened 2017-08-08
[DISCUSSION] trade with partial balance


metachris:
I would like to experiment with strategies that issue multiple buy advices, each only buying for a part of the current balance (eg. in percent). Perhaps a nice implementation would be like this: `this.advice('long', 10)` (10 for 10% of balance).

See also issue #438, PRs #448, #449, #450

Any thoughts on this topic or any reasons why this feature might not be desirable?

askmike:
This has been a major point of discussion, so here are the hard parts:

- The profit report (and all risk metric calculations, etc) assume that you trade in full swings. Implementing this not properly will mess up all profit reporting.
- If you allow a parameter like "max order 100 USD" Gekko can still drain a 1000 USD account. Unless a lot of trade data is persisted and there are 100% no bugs in any of that code (the code that writes trades to disk, reads from disk, in all cases incl. race conditions, etc).
- It is not recommended to trade on an account Gekko is already trading, it will mess up all profit calculations (which is based portfolio polling). So if users should not trade on the account, why would they keep funds they don't want to trade in the first place?

metachris:
Thanks for your thoughts.

I was primarily thinking of strategy development; to play around with strategies issuing multiple buys and later selling (or vice versa). Not sure if this has potential or not though.

thegamecat:
What risk metric calculations exist when you make a bid/ask decision?

askmike:
> Not sure if this has potential or not though.

To build a strat that can buy in small parts over time? I think it definitely has, this way you can have different levels of risk exposure your strat is willing to accept at certain points. Though the next step would be to have this logic close to the orderbook, to allow for specific types of orders in specific cases (which Gekko does not expose to strategies at all).

-----

> What risk metric calculations exist when you make a bid/ask decision?

The profit metric as well as all roundtrips (and thus risk metrics like sharpe ratio) are based on how much your portfolio increased because of the trades. If you could configure Gekko to only do 50 USD trades whilst having 100,000 USD on your exchange account your profit and loss (as calculated by gekko) would never be low or high, since if you make 50 euro your balance would only go to 100,050 USD.

thegamecat:
Ah I see. I thought you meant there was a predictive element based on a potential trade....I was going to say that's a hell of a miss by me!

metachris:
@askmike do you think this would be a good addition to gekko // do you want to have that?

askmike:
Yes definitely!  And not just me if you look at all the issues around this.

[...]

PR #1461, opened 2017-12-12
keepCurrency functionality

yuvalsegev:
Similar to the "keepAsset" functionality. 

Every time there's a "buy" action, gekko will use all the funds available.

Now, it will keep a certain percentage from the total funds (0 = keep nothing , 1= keep everything).

* **Other information**:


Ali1:
This doesn't really help if there are 2 completing assets that are being day trading with the same currency. What do you think of my suggestion #937

greenbigfrog (commented):
This looks "ok", but you should for sure add info to the example config.

Issue #1462, opened 2017-12-12
Bug: Death during buy attempt


Ali1:
Command line Gekko has been so good at not failing recently however failed me today!

```
2017-12-11 09:57:20 (INFO):	Trader Received advice to go long. Buying  XRP
2017-12-11 09:57:20 (INFO):	Pushbullet Message Sent
2017-12-11 09:57:20 (INFO):	Attempting to BUY 944.64226014714 XRP at Bitstamp price: 0.19866
2017-12-11 09:57:22 (INFO):	Send advice via email.
2017-12-11 09:58:21 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 09:58:21 (INFO):	Attempting to BUY 944.3586831440438 XRP at Bitstamp price: 0.19850
2017-12-11 09:59:21 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 09:59:22 (INFO):	Attempting to BUY 947.6777125069036 XRP at Bitstamp price: 0.19857
2017-12-11 10:00:22 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:00:23 (INFO):	Attempting to BUY 948.1537650072839 XRP at Bitstamp price: 0.19895
2017-12-11 10:01:23 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:01:24 (INFO):	Attempting to BUY 943.8392339233925 XRP at Bitstamp price: 0.19917
2017-12-11 10:02:24 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:02:25 (INFO):	Attempting to BUY 947.2021378029809 XRP at Bitstamp price: 0.19900
2017-12-11 10:03:25 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:03:26 (INFO):	Attempting to BUY 947.5349899598394 XRP at Bitstamp price: 0.19903
2017-12-11 10:04:26 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:04:26 (INFO):	Attempting to BUY 943.8392339233925 XRP at Bitstamp price: 0.19943
2017-12-11 10:05:27 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-11 10:05:27 (INFO):	Attempting to BUY 945.3046025942807 XRP at Bitstamp price: 0.19896
[...]

PR #1463, opened 2017-12-13
Fix for #1462 dying bitstamp

Ali1:
Bugfix #1462 

Can die

Prevents death in edge cases

* **Other information**:


exchanges/bitstamp.js
```diff
@@ -89,6 +89,9 @@ Trader.prototype.getFee = function(callback) {
 Trader.prototype.buy = function(amount, price, callback) {
   var args = _.toArray(arguments);
   var set = function(err, result) {
+    if (typeof result === 'undefined' || !result) {
```
askmike: if this if check gets hit the result object with not have an ID property, it might not crash anymore but it still causes Gekko to not be able to monitor the order (since this will pass `undefined` to Gekko).

askmike (changes requested):

askmike:
See comment (and please let me know if you think I am wrong).

A better way would be changing the ifcheck:

https://github.com/askmike/gekko/blob/1766fb032ac2751130fe17472e9f944280167703/exchanges/bitstamp.js#L91-L95

to something like `if(err || _.isUndefined(result) || result.status === "error") {`

exchanges/bitstamp.js
Ali1: I'm not actually sure where the result object has come from and can you clarify "gets hit the result object with not have an ID property"

Ali1:
If you're suggesting the whole change should be just ```if(err || _.isUndefined(result) || result.status === "error") {``` instead of my changes, then your still going to get an error on the next line ```log.error('unable to sell:', err, result.reason, 'retrying...');``` because result.reason is not defined.

Ali1:
Sorry, organised my branches and has messed up this pull request. Will make another one after your reply @askmike

PR #1464, opened 2017-12-13
Fix typos in tradebot doc

alexleventer:

Docs Update

* **Other information**:


docs/commandline/tradebot.md
```diff
@@ -6,11 +6,11 @@ You can set Gekko up as a tradebot, this will instruct Gekko to:
 - Run a strategy (in semi-realtime) over live market data.
 - Automatically create buy/sell orders based on signals coming from the strategy.
 
-*As with everything in Gekko, the tradebot will make decisions based on the by YOU selected/configured/created strategy. If you end up losing money you have no one to blame but yourself.*
+*As with everything in Gekko, the tradebot will make decisions based on the selected/configured/created strategy. If you end up losing money, you have no one to blame but yourself.*
```
askmike: I think the `by YOU` is an emphasis that we shouldn't remove. If the sentence structure is currently wrong, can we fix it instead of removing this?

askmike (changes requested):

docs/commandline/tradebot.md
greenbigfrog: IMO the emphasis on the `YOU` should stay. Maybe sth like `As with everything in Gekko, the tradebot will make decisions based on strategies, **YOU** select/configure/create.`?

alexleventer:
Fixed @askmike

greenbigfrog (approved):
This looks good now

askmike:
Awesome, thanks :)

PR #1478, opened 2017-12-14
Downgrade from postgres 9.5 dependency by not using 'on conflict do nothing'

kevinmershon:
- [x] Postgres versions earlier than 9.5 can now be used (good for older VMs/OSs)


askmike:
looks great, thanks :)

PR #1484, opened 2017-12-15
Fix loading JPG file on non standard install

JReming85:
This will load the image file reguardless of the path set in web/vue/UIconfig.js as it will always look for assets in the same directory as you browsed to

Bug fix

Currently the jpg on the main page is set to staticly pull from /asset/gekko.jpg

This will dynamically set the the root folder to match the set path

* **Other information**:


askmike:
You've only pushed an update to the compiled build file. This would get overwritten whenever we update something in the frontend app and recompile. Could you pushed the change to whatever source files you changed (inside this dir `gekko/web/vue/src`)?

greenbigfrog:
@askmike mind diversifying your labels a bit? In this case you aren't really waiting on an answer, but on a update.

JReming85:
@askmike Hey sorry about that, submitting it now

JReming85:
#1498

askmike:
#1498 has been merged!

PR #1486, opened 2017-12-15
Adds error type: MIN_TRADE_REQUIREMENT_NOT_MET

austinlangdon:
This add's the "minimum funds not available" error type from the bittrex exchange so that when this error is received, gekko doesn't continue looping/attempting to retry();. If the minimum funds aren't available, gekko shouldn't continue to retry.

Adds 'MIN_TRADE_REQUIREMENT_NOT_MET' error type for the bittrex exchange.

When this error is received by gekko, gekko continues to retry() ordering.

If this error is recieved by gekko, gekko doesn't continue to retry() ordering and instead only logs it as an error. If minimum trade requirement isn't met, it doesn't make sense to continue retrying();. 

* **Other information**:

### DIFF
--- a/exchanges/bittrex.js
+++ b/exchanges/bittrex.js
@@ -184,6 +184,7 @@ Trader.prototype.buy = function(amount, price, callback) {
         callback(null, 'dummyOrderId');
         return;
       } else if (err && err.message === 'MIN_TRADE_REQUIREMENT_NOT_MET') {
+        callback(null, 'dummyOrderId');
         log.error('Error buy ' , 'MIN_TRADE_REQUIREMENT_NOT_MET', err );
         return;
       }
--- a/exchanges/bittrex.js
+++ b/exchanges/bittrex.js
@@ -184,8 +184,8 @@ Trader.prototype.buy = function(amount, price, callback) {
         callback(null, 'dummyOrderId');
         return;
       } else if (err && err.message === 'MIN_TRADE_REQUIREMENT_NOT_MET') {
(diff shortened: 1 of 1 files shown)


exchanges/bittrex.js
```diff
@@ -183,6 +183,9 @@ Trader.prototype.buy = function(amount, price, callback) {
       } else if (err && err.message === 'DUST_TRADE_DISALLOWED_MIN_VALUE_50K_SAT') {
         callback(null, 'dummyOrderId');
         return;
+      } else if (err && err.message === 'MIN_TRADE_REQUIREMENT_NOT_MET') {
+        log.error('Error buy ' , 'MIN_TRADE_REQUIREMENT_NOT_MET', err );
+        return;
```
askmike, line 189: would mind calling the callback function with the error here? Else Gekko never completes tracking the order.

askmike (changes requested):
see nit.

austinlangdon:
@askmike  added in the callback

greenbigfrog (approved):

austinlangdon:
@askmike would you like this commit to be merged into develop instead of stable?

greenbigfrog:
Development normally happens against develop @austinlangdon

austinlangdon:
@greenbigfrog @askmike updated base to develop

austinlangdon:
@greenbigfrog @askmike 

is this good to be pulled?

greenbigfrog:
Mike is the one to ping in cases like this :)

PR #1519, opened 2017-12-18
Fix sqlite writing candles with low vol markets in realtime

cmroche:

Bugfix

When leeching a market and writing the candles, a reporting period with no candle still dispatched the candle to the writer which pretty much does nothing. However this can cause SQLITE_CORRUPTED errors to be thrown by other processes trying to read data at the same time.

Even though this error occurs I've been able to open, and run `pragma integrity_check` on the DB with no issues, restarting doesn't seem to be a problem either. None the less, we should fix this.

Simple check that there is candle data to write before trying, if not don't open the transaction.

* **Other information**:

### DIFF
--- a/plugins/sqlite/writer.js
+++ b/plugins/sqlite/writer.js
@@ -53,6 +53,9 @@ Store.prototype.writeCandles = function() {
     return;
 
   var transaction = function() {
+    if (!this.cache.length)
+      return;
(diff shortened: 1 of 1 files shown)


cmroche:
Actually this didn't resolve the issue, still investigating.

Issue #1473, opened 2017-12-14
GDAX not importing data?


outpoints:
* **I'm submitting a ...**
[x] bug report

* **Action taken** (what you did)
Nothing happens when I try to import gdax data?

* **Expected result** (what you hoped would happen)
Import like any other exchange?

* **Actual result** (unexpected outcome)
![image](https://user-images.githubusercontent.com/12377381/33968854-09305b00-e01f-11e7-8f38-dfc9e04fb6cd.png)
![image](https://user-images.githubusercontent.com/12377381/33968857-12b4dd9a-e01f-11e7-9cd7-8861d7d227e7.png)
![image](https://user-images.githubusercontent.com/12377381/33968958-ae41f6a8-e01f-11e7-942f-7b6c483c0346.png)
![image](https://user-images.githubusercontent.com/12377381/33968998-d5d0ec42-e01f-11e7-8301-50b4e2766166.png)

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
No idea

mew351:
![image](https://user-images.githubusercontent.com/12377381/33968901-5293e500-e01f-11e7-839d-3b86cbdc0865.png)
Apparently this happens

holeyness:
Here is the cause, GDAX has a rate limit of 3 - 6 requests a second. The code in gdax.js, in the method getTrades ignores that and blasts threw that limit. When the api returns 429 (rate limit exceeded), the code retries the whole thing, thus it will never complete.

outpoints:
@holeyness do you have a fix?

holeyness:
@mew351 im working on one, ill see if i can make a pull req when it works

outpoints:
Let me know if I can help with anything!

tdsticks:
I simply added a sleep at the start of the process method around line 228.  This is just a workaround, I'm sure there's a better solution.

`    var process = function(err, response, data) {`
`        sleep.sleep(2)`

You'll need to install the sleep npm
`npm install sleep --save`

outpoints:
So something like this?

`    var process = function(err, response, data) {
    sleep.sleep(2)
    var process = function(err, response, data) {
        if (data && data.message)
            err = new Error(data.message);       `

tdsticks:
This is what I have:

[...]

PR #1534, opened 2017-12-21
Throttle requests to GDAX

mmeisel:

Bug fix

See issue #1473 

* **Other information**:

To throttle requests, I simply introduced a wrapper for getProductTrades that only sends the requests every 334 ms (3 times per second). If GDAX sends back a 429, anyway, increase this delay slightly. This should account for cases where, e.g., more than one import is running from the same IP.

### DIFF
--- a/exchanges/gdax.js
+++ b/exchanges/gdax.js
@@ -224,13 +224,41 @@ Trader.prototype.cancelOrder = function(order, callback) {
 Trader.prototype.getTrades = function(since, callback, descending) {
     var args = _.toArray(arguments);
     var lastScan = 0;
+    // GDAX rate limit of 3 per second
+    var minDelay = 334;
+    var lastAttempt = 0;
+
+    var getProductTrades = (options, gdaxCallback) => {
+        var now = moment.utc();
+
+        if (now - lastAttempt < minDelay) {
+            setTimeout(() => getProductTrades(options, gdaxCallback), minDelay - (now - lastAttempt));
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


holeyness:
Hey @mmeisel, so GDAX will sometimes return 429 even when you don't think it makes sense. The code currently will retry the entire import if it ever goes wrong (see this.retry). My solution was to keep the "after" variable in the request on hand to retry just that specific request if it went wrong, and keep retrying until it worked.

mmeisel:
Hey @holeyness, my code should handle that case just fine, it doesn't call the callback when it gets a 429, it just tries again with a longer delay (see line 242).

gregorybesson:
Hello @mmeisel Thx for your code. It solves the gdax import problem !

cmroche:
@mmeisel Thanks for this contribution! As it turns out I've spent the last couple of days rewriting the importer (and GDAX implementation) to be more robust overall. I also added handling for rate-limiting, and catching the 429 response as a retryable error in that effort. It could use some testing though if you are live trading. #1543

mmeisel:
Hi @cmroche, that's great, sure, I'd be happy to test it out. I'll comment on your PR afterwards.

zipzapp:
@mmeisel I applied your PR and the gdax importer is working for me now. Good idea with the throttle support. It'll be good to see this solution folded into a more generalizable solution though!

mmeisel:
@athaax I believe that's @cmroche's intention in #1543, this was just meant as a quick hack to get things working again.

PR #1543, opened 2017-12-24
GDAX - Move to new retry/error handling approach and re-write importer

cmroche:

Feature

Retry and error handling is sparse, errors cause non-ending retries in some cases, and abort in others. Normalizing errors is not consistent between function calls.

While importing, the backscan process prevent the importer from showing a progress update and causes an in-memory buffer of the full import to be stored.

Moving GDAX to using the new pattern for retry and error handling, similar to that of Kraken. This is not tested yet in live trading and should be validated with someone who actively uses live trading on this exchange.

Re-wrote the GDAX importer and moved to a custom getTrades function in the importer which scans for the starting TID in history, then batches and subiterates over 1000 trades at a time. This means that the progress bar updates correct while the import is happening.

Also added rate limiting based on GDAX documentation, and handling for 429 (rate-limit exceeded messages) should they occur.

Added error handling for non 200 HTTP response codes, and marked the typical ones for retry. I found that GDAX actually does get 504 Gateway Timeouts rarely from Cloudflare.


mmeisel:
Hi @cmroche, the indentation is all messed up here, and there are two close braces missing at the end of the process function. Maybe you had a bad merge? FWIW, when I change the indentation in a file, I generally like to have a separate commit that only fixes the indentation with no other changes, it makes merges and diff reading way easier.

cmroche:
@mmeisel Oy, that was one messed up merged.

ddey94:
Hey @cmroche , thanks for writing this! I copied over the two files to test out the PR. Do you know why it takes a few minutes for the download to start? Only after a few minutes did the view update and progress indicator start.

cmroche:
@ddey94 That is normal, this is due to the fact that GDAX paginates trade history by trade ID and not date, and then requires you to scan backwards through time. When an import is started the new importer does a quick scan through history to map out the first trade it finds with a date before your requested start it then starts to batch forward until it get to the first trade. This process causes a 2 - 5 minute delay at the start of an import, but is necessary to make the progress bar work during the import.

ecomkings:
Deprecation warning: moment().add(period, number) is deprecated. Please use moment().add(number, period). See http://momentjs.com/guides/#/warnings/add-inverted-param/ for more info.

cmroche:
@ecomkings Fixed in another PR.

PR #1545, opened 2017-12-24
Multistage Docker Alpine Build - 130MB

pepijnblom:
Major tweakage of the Docker file

The whale is obese and takes up too much space

Whale now fits designer clothes and gets all the chicks

* **Other information**:

A few comments on my changes:
- You do not need to mkdir when setting a workdir, this is done for you. I see a lot of people doing this wrong :)
- I decided to build in a separate step because you don't really need python and the whole build-base/build-essential in your container once it's been built
- Because everything is built in separate stages you do not need to clean up, those stages are discarded automatically
- I added [node-prune](https://github.com/tj/node-prune) which cleans up the node_modules dir of useless files. This would not build in the node build stage so had to give it's own stage.
- I added tini so zombies are handled properly, tried to use pm2 first but that took way too long to get working. Tini is easy, simple and works just beautifully.
- I changed the workdir to /app, just because OCD
- Because it's now multistage you will need Docker 17.05 or higher
- The docker-compose.json could maybe use a bit more tweaking

I hope someone can test this a bit more. I actually got carried away and did all this without even having used the application. I was about to try it out for the first time and then noticed how bloated the image was, so I couldn't help myself. I just love Dockerizing, I think I'm addicted 😭 

All this took the docker image size down *from 769MB to 130MB*


Dockerfile
```diff
@@ -1,28 +1,32 @@
...
-# Install app dependencies
-COPY package.json /usr/src/app
-RUN npm install --production && \
+COPY package*.json ./
+COPY --from=golang /go/bin/node-prune /usr/local/bin/
+RUN set -x && \
+    apk add --no-cache python build-base && \
+    npm install -g --production node-gyp && \
+    npm install --production && \
     npm install --production redis@0.10.0 talib@1.0.2 tulind@0.8.7 pg && \
```
hoshsadiq: What's the reason for this not being in package.json?

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
-
-CMD [ "npm", "start" ]
\ No newline at end of file
+# Prepare the application
+WORKDIR ${WORKDIR}
+COPY . .
+COPY --from=build /usr/src/app/node_modules/ /app/node_modules/
+RUN set -x && \
+    apk add --no-cache tini &&  \
+    ./configureUI.sh
```
hoshsadiq: Probably better to have this mounted.

Dockerfile
cmroche: node-gyp has to be installed globally to build properly, talib and tulind are optional modules so they are not part of the package.json

cmroche (approved):

hoshsadiq:
One more thing! Need to build latest UI. see #1223. Can't seem to get it built at the moment (issue with cross-env, see #1551), but the following should be enough:
```
COPY web/vue ./web/vue

RUN set -ex; \
    cd ./web/vue; \
    apk add --no-cache git; \
    npm install -g --production cross-env@3.X; \
    npm install; \
    npm run build; \
    du -h -d 1; \
    node-prune; \
    du -h -d 1
```

And in final container:
```
COPY --from=build /usr/src/app/web/vue/dist/ ${WORKDIR}/web/vue/dist/
```

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
-ENV HOST localhost
-ENV PORT 3000
-
-# Create app directory
-RUN mkdir -p /usr/src/app
+FROM node:8-alpine as build
 WORKDIR /usr/src/app
-
-# Install GYP dependencies globally, will be used to code build other dependencies
-RUN npm install -g --production node-gyp && \
```
hoshsadiq, line 11: Needs `-ex` instead. Otherwise if any of these commands fail, the docker image continues building.

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
+# Set environment vars
+ENV HOST localhost
+ENV PORT 3000
+ENV WORKDIR /app
 EXPOSE 3000
-RUN chmod +x /usr/src/app/docker-entrypoint.sh
-ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
-
-
-CMD [ "npm", "start" ]
```
hoshsadiq: Needs `-ex`

Dockerfile
```diff
@@ -1,28 +1,32 @@
...
+    du -h
+    
+FROM node:8-alpine
+# Set environment vars
+ENV HOST localhost
+ENV PORT 3000
+ENV WORKDIR /app
 EXPOSE 3000
-RUN chmod +x /usr/src/app/docker-entrypoint.sh
-ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
```
hoshsadiq, line 25: Need to remove this file

Dockerfile
pepijnblom: You mean the configureUI.sh script? That's actually not a bad idea..

Dockerfile
pepijnblom: Good point, I'm actually not too familiar with npm, my core strength lays with php & infrastructure. I was actually surprised things just continued even if there was an error.

pepijnblom:
I'm going to have to take a look at the new UI some other time, now it gets confusing :)

Dockerfile
hoshsadiq: Sorry! I was mistaken, you are using `&&`, I generally use `;` to separate them. This confused me. They should behave the same way though :)

Dockerfile
hoshsadiq: No, I mean the whole file that configureUI is editing should just be mounted. `configureUI.sh` shouldn't be needed.

pepijnblom:
If https://github.com/askmike/gekko/pull/1223 could be merged first then I could just merge from upstream and fix the Dockerfile. I think that would be the best course of action?

askmike:
Hey @pepijnblom, what do you want out of #1223? I don't want to merge because I don't agree with the changes to the documentation. We can cherrypick other stuff.

pepijnblom:
Sorry, forgot about this mr. Of you are not merging #1223 my mr should be good to go. Just be sure to test the whole application because I haven't.. I'm not sure if python is needed during runtime or just during build.

greenbigfrog:
@pepijnblom you sure you didn't just push some stuff you didn't plan on pushing?

pepijnblom:
Oops, you're right.. I'll make a new MR at a later date, I've been having some problems with tulind/talib so I want to get those fixed first. I currently removed the 2nd build step just to be sure that's not causing the problems.

PR #1546, opened 2017-12-24
Add Kodi notifications

billymcintosh:
Allow sending of start-up and trade advice to Kodi and derivatives

Feature that enables sending messages to Kodi.

No such functionality exists.

This addition will send optional startup message and advice to an active Kodi installation if the jsonrpc service is enabled.

* **Other information**:


greenbigfrog:
This is cool! Thanks!

billymcintosh:
Refactored a little bit as the message string was too long and kodi was ignoring it.    Removed the unnecessary `node-fetch` and utilised the already installed `request` module & a little tidy up.

PR #1554, opened 2017-12-25
Update bitcoin-co-id

nevrending:
- Update bitcoin-co-id currency pairs and minimal orders
- Use === to compare against null
- indentation and empty lines
- missing semicolons


greenbigfrog:
WIthout having tested this, this looks good!

PR #1559, opened 2017-12-26
Better slack messages

tg44:
PIMP the slack messages for a more readable format.

screen shot 2017-12-26 at 11 24 26

screen shot 2017-12-26 at 11 24 45

* **Other information**:
Just see the pictures :D

### DIFF
--- a/plugins/slack.js
+++ b/plugins/slack.js
@@ -17,20 +17,11 @@ const Slack = function(done) {
 
 Slack.prototype.setup = function(done) {
     this.slack = new WebClient(slackConfig.token);
-    
+
     const setupSlack = function(error, result) {
         if(slackConfig.sendMessageOnStart){
-            const exchange = config.watch.exchange;
-            const currency = config.watch.currency;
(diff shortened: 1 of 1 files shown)


greenbigfrog (approved):

herkyl:
Usage of template strings would make this code much easier to read. Nice PR though, so much more readable!

tg44:
@herkyl like this?

herkyl:
I was referring to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals . This change really good as well.

tg44:
I would let that go for now. I'm not qualified enough to make this change without braking the code. Plus I think that wouldn't make it more readable... I could use that in 3 places. L42 is problematic bcs of the backticks in the string. L83 has more code than string. L88 is an url  (`/asd/${asd}.png`). I wouldn't use string literals in these cases in scala either :D

herkyl (approved):

askmike:
looks great, thanks!

PR #1568, opened 2017-12-28
Binance API changes now limit queries to 1 hour, no longer 1 day.

cmroche:

Bugfix

Binance changed their API very recently, to limit historical queries to 1 hour (from 24)

Implemented the query change

* **Other information**:


importers/exchanges/binance.js
```diff
@@ -24,13 +24,13 @@ var fetch = () => {
 var handleFetch = (unk, trades) => {
   if (trades.length > 0) {
     var last = moment.unix(_.last(trades).date).utc();
-    var next = last.clone();
+    var next = from.clone().add(1, 'h').subtract(1, 's');
   } else {
```
thecodeboss: I'm curious why you need to subtract 1 second here. Could you comment about that?

thecodeboss (commented):
Looks good to me, just one comment about the 1 second thing.
For the record I've been using this branch since yesterday and imports seem to work as expected.

importers/exchanges/binance.js
cmroche: Binance has millisecond resolution in their queries, Gekko only deals with seconds in the imported trades. If you did not subtract 1 second then you potentially lose 999 milliseconds of records on every iteration.

Also note that I no longer use the last imported record's time, but instead the start time of the last iteration. Binance's queries work by specifying a start and a stop, so we iterate more efficiently if we fix every iteration relative to the last, not relative to the last imported query. This also stop us from getting stuck if any given 1 hour period has no results.

thecodeboss: Ah I see. Binance limits queries to 3600 seconds (1 hour). But if you just specify start and end ranges in the naive way, you're actually asking for 3601 seconds, and Binance will truncate it? Hence you lose 999 ms of data.
Makes sense to me. Ship it!

thecodeboss (approved):
Seems solid to me.

importers/exchanges/binance.js
cmroche: @thecodeboss In re-reading this... it's an error. Since I moved away from using trade times, and to fixing the iterator, the minus 1 second is no longer necessary and won't loose any data. I'll remove it :)

SICKFREDO:
I made the suggested changes and this is the error i'm getting, how did you bypass that??

img_20171229_164931745

cmroche:
@SICKFREDO This is not a limitation of the binance exchange implementation or importer, this must be a limitation specifically set in Gekko somewhere. However Binance is pretty fast at importing, why not just import a full day?

SICKFREDO:
@cmroche im confused, i did that fix and that states they only allow 1hr queries now,
When i do a full day i get this error:
image
  <-- POST /api/import
Import 4141849383593814 started
  --> POST /api/import 200 156ms 150b
C:\Users\Zion\gekko-stable\exchanges\binance.js:35
  if (!error || !error.message.match(recoverableErrors)) {
                               ^

TypeError: Cannot read property 'match' of undefined
    at Trader.retry (C:\Users\Zion\gekko-stable\exchanges\binance.js:35:32)
    at Trader.bound [as retry] (C:\Users\Zion\gekko-stable\node_modules\lodash\d
ist\lodash.js:729:21)
    at Trader.process (C:\Users\Zion\gekko-stable\exchanges\binance.js:70:19)
    at bound (C:\Users\Zion\gekko-stable\node_modules\lodash\dist\lodash.js:729:
21)
    at Request.request [as _callback] (C:\Users\Zion\gekko-stable\node_modules\b
inance\lib\rest.js:52:21)
    at Request.self.callback (C:\Users\Zion\gekko-stable\node_modules\request\re
quest.js:188:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
    at Request.<anonymous> (C:\Users\Zion\gekko-stable\node_modules\request\requ
est.js:1171:10)
    at emitOne (events.js:116:13)
RECEIVED ERROR IN IMPORT 4141849383593814
Child process has died.

thecodeboss:
@SICKFREDO Try rebasing on top of #1508. The error handling is broken for Binance, but that PR fixes it.

Medardas:
Sorry, but I think i have better fix #1637

cmroche:
@Medardas Respectfully I disagree, there are a couple of bugs in your PR. The only notable change was going from 1h to 3600000ms, could you explain why that is better? I am more than happy to incorporate it into my change if it alleviates some issue.

Additionally you can send a PR to the branch used for my PR and I can merge that in to be included with the change.

Medardas:
1h sometimes would cause error that candles are longer than 1h still to be displayed, no idea why, maybe processor thing.
I did create pull request to your branch, if you will solve conflicts fine, I will revoke my PR here

cmroche:
@Medardas That is indeed interesting, actually it may make a lot of sense. Moment converts the time representation into a unix epoch timestamp, which is require to consider leap-time in the conversion for a correct mapping. It is possible that leap seconds are being introduced and that Binance is rejecting the query based on this, HOWEVER adding 3600000 ms should produce the same problem.

There is a case here for a 1 second subtraction in any case, I'll make that change manually.

Perhaps you could validate then that subtracting 1 second rectifies your problem, or if not provide details about what pair and time range generates the error.

Thanks for the help.

Edit: As a note, I was using the -1 second approach last week when I did a test and imported the full history of all asset pairs on Binance. So I do expect that this should work.

cmroche:
@askmike Appveyor appears to be broken. I don't think it is an issue with this change.

eskimojoecrypto:
Hey guys, I'm also getting a similar error:

C:\Users\lukem\Downloads\gekko-stable\gekko-stable\exchanges\binance.js:35
  if (!error || !error.message.match(recoverableErrors)) {
                               ^

TypeError: Cannot read property 'match' of undefined
    at Trader.retry (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\exchanges\binance.js:35:32)
    at Trader.bound [as retry] (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\node_modules\lodash\dist\lodash.js:729:21)
    at Trader.process (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\exchanges\binance.js:70:19)
    at bound (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\node_modules\lodash\dist\lodash.js:729:21)
    at Request.request [as _callback] (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\node_modules\binance\lib\rest.js:66:21)
    at Request.self.callback (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\node_modules\request\request.js:188:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
    at Request.<anonymous> (C:\Users\lukem\Downloads\gekko-stable\gekko-stable\node_modules\request\request.js:1171:10)
    at emitOne (events.js:116:13)
RECEIVED ERROR IN GEKKO 508328913393901
Child process has died.

How do I fix this please?

PR #1569, opened 2017-12-28
Minor fixes to CCI strategy

werkkrew:
Fix

- Debug output is missing data/formatted wrong
- History parameter in strategy pulls from the wrong setting

- Cleaner output
- History parameter in strategy pulls from settings instead of trading advisor, although I believe the actual indicator already does this so it does not actually have a change in effect.

* **Other information**:


askmike:
:+1:

PR #1571, opened 2017-12-28
Update binance.js irrecoverableErrors handling

antonrifco:
Current code caused issue on my server, and I fixed it this way

Bug Fix for handling binance irrecoverableErrors

Current code is not clear and caused issue during the runtime

To detect error and check if it's outside of irrecoverable error list

* **Other information**:


cmroche:
@antonrifco Thanks for your work, I've already made this change in one of my pending PRs.

Edit: PR #1508 refactors the error and retry handling within Binance to be much more robust and incorporates the check you added here.

cmroche:
@askmike @antonrifco  This PR is no longer valid, the issue has been resolved and merge from another PR. Please consider closing this PR.

For posterity, this fix is also incorrect. If error.message is undefined we want to handle as irrecoverable. Retrying an error should only ever happen in clearly defined cases that are known to be retryable.

Issue #1306, opened 2017-11-09
Unable to stop a market watcher or Strat runner


djh82uk:
Is there any way?

StevenHop:
Lurking for a while now: this is a question i can answer easily :)
The documentation states:

> Stopping Gekko :  In the command line hold ctrl + c.

It is on the bottom of this page: [https://gekko.wizb.it/docs/installation/installing_gekko_on_windows.html](url)

djh82uk:
Yeah I can stop the script, but I just want to stop the Strat from running

StevenHop:
Ah. I recognize the problem. If you have two Strats running you have to kill both. Somewhere i read this is on the roadmap. I have solved this for now by starting each gekko in a separate directory with their own port. C:\Gekko0=3000, C:\gekko1=3001, enz..  I sometimes copy the database between them or to a third directory to test strategys..

tomand285:
Another way of doing this is saving the [PID](https://nodejs.org/api/process.html#process_process_pid) of each process into the database and then [KILL](https://nodejs.org/api/process.html#process_process_kill_pid_signal) the process you want to remove. You could make a html button to remove each processes. I recommend saving them in the DB because if the main program crashes, the child processes can still be controlled upon restart of the UI program. Of course you would have to check to see if the processes are still running every now and then and remove the ones that are not from the DB but that should be simple.

askmike:
The UI is very basic for now. This will be properly supported in the next version. See #911 for details :)

hilkoc:
I finally got this fixed. See my last comment on #2059.

PR #1577, opened 2017-12-29
Adds Stop runner / watcher button

ricardojsmarques:
Adds a stop button below gekko title.
Internally calls already defined API method `killGekko`.

This change adds a "Stop runner" button in the strat runner and market watcher individual page.

There isn't a way to stop a runner from the UI as shown in #1306 

The added button stops the current live runner (market watcher or strat runner)

* **Other information**:
Currently the button effectively kills the live runner, but I couldn't find a way to update the list view after  the runner was killed. refreshing the page effectively updates the runner lists


osztro:
@ddrjm ,  I  use 0.5.12 and called the killGekko endpoint from Postman. That may uses the api, but the child process runs after that. So the program keeps running the (watcher|papertrader|tradebot).
Are you aware of this? However you are right, One piece of this puzzle is to make a button available on UI.

hilkoc:
@ddrjm  I've build Gekko with this pull request merged in. It does remove the runners from the UI. However it does not kill the process! Like @osztro says, Gekko keeps executing trades/fetching data after clicking the stop button.

@askmike How do we correctly kill the child process?

osztro:
@hilkoc node package https://www.npmjs.com/package/tree-kill
use that to stop the gekko process
store the PID (process id ) of the gekko in the gekkoManager (the cache) and listManager. currently a randomly generated id is the key in the cache (gekkoManager)
use the existing killGekko.js as an entry point to perform the stop of the process.

askmike:
@osztro the child process currently doesn't handle graceful shutdown very well, there is some more work involved than simply killing the process to make sure we clean up DB handlers and the such.

This (most desired) functionality is planned for in the next UI release, I already have a rewritten all code that manages running gekkos (current code was alpha & spaghetti code) will push this to #1850 asap. Better to hook proper lifecycle management in there.

Apologies to @ddrjm for putting the time in to write this, but I think the best thing would be to close this for now.

hilkoc:
Thanks @ddrjm , with the help of this button in the UI I managed to make a working stop button. See my last comment on #2059.

PR #1583, opened 2017-12-30
Change the order of the subtract function params in the moment.js

cloggy45:

Bug fix

It currently spams `Deprecation warning: moment().subtract(period, number) is deprecated. Please use moment().subtract(number, period).` when running in real time mode. 

It removes this specific message.

* **Other information**:

I know there is an exact same issue with the add function that I will look into also.


askmike:
👍

Issue #1497, opened 2017-12-17
Error: SQLITE_MISUSE: Database is closed


wupeng1211:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[*] bug report
[ ] feature request
[ ] question about the decisions made in the repository

`2017-12-17 12:16:58 (INFO):             Usable local data available, trying to match with exchange data..
2017-12-17 12:16:58 (INFO):             Full history locally available. Seeding the trading method with all required historical candles.
2017-12-17 12:17:18 (INFO):     Setting up:
2017-12-17 12:17:18 (INFO):              Mailer
2017-12-17 12:17:18 (INFO):              Sends you an email everytime Gekko has new advice.
2017-12-17 12:17:18 (INFO):

2017-12-17 12:17:20 (INFO):     Send advice via email.
2017-12-17 12:17:20 (INFO):     Setting up:
2017-12-17 12:17:20 (INFO):              Advice logger
2017-12-17 12:17:20 (INFO):
2017-12-17 12:17:20 (INFO):     Setting up:
2017-12-17 12:17:20 (INFO):              Paper Trader
2017-12-17 12:17:20 (INFO):              Paper trader that simulates fake trades.
2017-12-17 12:17:20 (INFO):

2017-12-17 12:17:20 (INFO):     Setting up:
2017-12-17 12:17:20 (INFO):              Performance Analyzer
2017-12-17 12:17:20 (INFO):              Analyzes performances of trades
2017-12-17 12:17:20 (INFO):

2017-12-17 12:17:20 (INFO):     Starting to watch the market: Poloniex BTC/USDT
Deprecation warning: moment().add(period, number) is deprecated. Please use moment().add(number, period). See http://momentjs.com/guides/#/warnings/add-inverted-param/ for more info.
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: SQLITE_MISUSE: Database is closed
    at Error (native)
    at Database.run (/home/ka/gekko/node_modules/sqlite3/lib/sqlite3.js:20:25)
    at Store.transaction (/home/ka/gekko/plugins/sqlite/writer.js:56:13)
    at Database.bound (/home/ka/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Store.writeCandles (/home/ka/gekko/plugins/sqlite/writer.js:88:11)
[...]

PR #1591, opened 2017-12-30
FIx SQLITE_MISUSE error (accidental DB handle closure)

cmroche:

Bugfix

Occassional "SQLITE_MISUSE" error when importing due to improperly handled DB closure. This happens because the DB handled is cached in multiple require('handle') calls but was being treated as always instanced.

handle.js returns a function which will handle initializing the db handle as an instance instead of as a globally cached handle. This means that individual consumers of the DB can close when they are done without risk to other consumers.

* **Other information**:

SQLITE is very robust at handling multiple connections both in the same process and outside of the same process. There is no perf or safety gain by using a globally cached handle.

Fixes issue #1497


askmike:
:+1:

PR #1612, opened 2018-01-02
bugfix - bitfinex minimal order size IOTA/ETH

Dries-Vandenneucker:
Bug fix

The bot tries to place an buy order while the minimum order size is not met.

The bot uses the new minimum order size, as is given in the specific error message from bitfinex.

* **Other information**:
While I can’t find the exact documentation for the correct minimal orders of all currencies on bitfinex, I do get an error while running the bot in command line: “unable to buy Error: Invalid order: minimum size for IOTA/ETH is 6.0”


askmike:
thanks :)

PR #1629, opened 2018-01-04
Handle rejected promises

billymcintosh:
Bugfix to handle rejected promises in case of spelling errors or APIs returning bad data. 

When the marketFetcher received a warning and tries to log to the console, it finds a misspelt variable and throws rejection

Fixes the spelling mistake and adds a listener to `gekko.js` to catch any other unhandled promises.   This listener only logs to the console for now.

* **Other information**:
Happy for the unhandledRejection process listener to be dumped but it will be worthwhile going forward when most people catch up with the latest node versions.  In future node versions, the process will exit on unhandled rejections.

### DIFF
--- a/gekko.js
+++ b/gekko.js
@@ -32,6 +32,7 @@ console.log(`
    $$$$$$/  $$$$$$$$/ $$/   $$/ $$/   $$/  $$$$$$/
 `);
 
+const log = require(__dirname + '/core/log');
 const util = require(__dirname + '/core/util');
 
 console.log('\tGekko v' + util.getVersion());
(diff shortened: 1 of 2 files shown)


gekko.js
```diff
@@ -59,3 +59,7 @@ pipeline({
   mode: mode
 });
 
+process.on('unhandledRejection', (reason, p) => {
+  console.log(`Unhandled Rejection at: ${p}, Reason: ${reason}`);
```
askmike: can we use the the logger that comes with Gekko, and use the `error` method to log to stdout?

That logger can be found here: https://github.com/askmike/gekko/blob/develop/core/log.js

askmike (changes requested):

gekko.js
billymcintosh: Yep, makes sense.

billymcintosh: Updated and pushed changes.

PR #1636, opened 2018-01-05
Web UI basic auth

yohanesgultom:

* Web UI basic auth using old version of `koa-basic-auth`. Username and password are set using env variables BASIC_USER and BASIC_PASSWORD


askmike:
I don't think adding basic auth on this level is a good idea: basic auth sends passwords in plain text, which means it's extremely bad on public networks if your traffic is not encrypted. For this reason it's better to put something else in front that can add something like SSL as well is proven to quite secure. 
 
In the docs we recommend NGINX for this.

yohanesgultom:
I see. Nevermind, then. I deploy gekko in public vps & I don't want anyone playing around with it especially after I setup the exchange api keys. So I personally use this to prevent unwanted access.

PR #1637, opened 2018-01-05
Binance importer fix. Over #1568

Medardas:

Bug fix for Binanace importer

Daily price history requests

Hourly price history requests and additional check for incremental TID

* **Other information**:

Fix over #1568 - check for incremental TID because sometimes Binance returns several transaction within second and result list can cointain only one item. That results in a loop for history requests.

### DIFF
--- a/exchanges/binance.js
+++ b/exchanges/binance.js
@@ -134,14 +134,14 @@ Trader.prototype.getPortfolio = function(callback) {
 
     if (!_.isNumber(assetAmount) || _.isNaN(assetAmount)) {
       log.error(
-        'Binance did not return portfolio for ${this.asset}, assuming 0.'
+        `Binance did not return portfolio for ${this.asset}, assuming 0.`
       );
       assetAmount = 0;
     }
 
     if (!_.isNumber(currencyAmount) || _.isNaN(currencyAmount)) {
       log.error(
-        'Binance did not return portfolio for ${this.currency}, assuming 0.'
+        `Binance did not return portfolio for ${this.currency}, assuming 0.`
       );
       currencyAmount = 0;
     }
--- a/exchanges/binance.js
+++ b/exchanges/binance.js
@@ -22,7 +22,7 @@ var Trader = function(config) {
     key: this.key,
     secret: this.secret,
     timeout: 15000,
-    recvWindow: 60000, // suggested by binance
+    recvWindow: 60000, // suggested by binance, defaults to 5000 millisecond. With recvWindow, you can specify that the request must be processed within a certain number of milliseconds or be rejected by the server.
     disableBeautification: false, // better field names
   });
 };
@@ -32,8 +32,8 @@ var recoverableErrors = new RegExp(
 );
 
 Trader.prototype.retry = function(method, args, error) {
-  if (!error || !error.message.match(recoverableErrors)) {
-    log.error('[binance.js] ', this.name, 'returned an irrecoverable error');
+  if (!error || (error && !error.msg.match(recoverableErrors))) {
+    log.error('[binance.js] ', this.name, 'returned an irrecoverable error or there is no error');
     return;
   }
 
@@ -94,7 +94,7 @@ Trader.prototype.getTrades = function(since, callback, descending) {
 
   if (since) {
     var endTs = moment(since)
-      .add(1, 'd')
+      .add(3600000, 'ms')
       .valueOf();
     var nowTs = moment().valueOf();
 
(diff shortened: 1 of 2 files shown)


helmuteke:
Great ,

ATM start import from binance . Have to look further now. Thanks

exchanges/binance.js
```diff
@@ -32,8 +32,8 @@ var recoverableErrors = new RegExp(
 );
 
 Trader.prototype.retry = function(method, args, error) {
-  if (!error || !error.message.match(recoverableErrors)) {
-    log.error('[binance.js] ', this.name, 'returned an irrecoverable error');
+  if (!error || (error && !error.msg.match(recoverableErrors))) {
```
billymcintosh: If `error` isn't a falsy then it's a truthy and there's no requirement to re-check. 
A check for `error.msg` existence before the regex would be beneficial though.

```js
if (!error || (error.msg && !error.msg.match(recoverableErrors))) {
```

Medardas: actually retry() seems not always to be called with defined error, thats why there was "type error undefined for .match" before

exchanges/binance.js
```diff
@@ -134,14 +134,14 @@ Trader.prototype.getPortfolio = function(callback) {
 
     if (!_.isNumber(assetAmount) || _.isNaN(assetAmount)) {
       log.error(
-        `Binance did not return portfolio for ${this.asset}, assuming 0.`
+        'Binance did not return portfolio for ${this.asset}, assuming 0.'
```
cmroche: Won't print out correctly, you need backticks for the inlined variables.

exchanges/binance.js
```diff
@@ -134,14 +134,14 @@ Trader.prototype.getPortfolio = function(callback) {
...
       );
       assetAmount = 0;
     }
 
     if (!_.isNumber(currencyAmount) || _.isNaN(currencyAmount)) {
       log.error(
-        `Binance did not return portfolio for ${this.currency}, assuming 0.`
+        'Binance did not return portfolio for ${this.currency}, assuming 0.'
```
cmroche: Same here

importers/exchanges/binance.js
```diff
@@ -22,15 +22,15 @@ var fetch = () => {
 };
 
 var handleFetch = (unk, trades) => {
-  if (trades.length > 0) {
-    var last = moment.unix(_.last(trades).date).utc();
-    var next = last.clone();
+  if (trades.length > 0 && lastTid < _.first(trades).tid) {
+    var lastTid = _.last(trades).tid;
+    var next = moment.unix(_.last(trades).date).utc();
```
cmroche, line 27: Don't use the timestamp of the last trade, if there are no trades within a 1 hour period (which does in fact happen) then it will get stuck here. Please see my PR with this fix, it addresses this issue.
  
(Note this fix was made on #1568 which was the change I submitted to support 1h queries, when using 24h queries this wasn't important since that case didn't happen).

cmroche: Actually I see that this bug is also still present in my PR... I made this change somewhere, and have lost it. I'll reapply to my branch momentarily

Medardas: right, it could return empty...in that case we need to go to another hour, I'll push a fix

Medardas: if there are no trades it will just go to another hour, as expected =]
Hence first check:
`trades.length > 0`

exchanges/binance.js
Medardas: well that was silly of me

mboremski:
Could yo have look here @cmroche?
Maybe this is kind of outdated since #1568 has been merged.
TY

cmroche:
@mboremski @Medardas Yes, this PR shouldn't be applicable any longer as PR #1568 has been merged and should handle the leap-second issue when requesting trade data time ranges.

askmike:
@Medardas thanks for your work, do you agree with @cmroche? 

If you do: Apologies for the double work, I'm way behind on PRs which is the likely reason.
if you don't: please re-open and tell us your thoughts :)

Medardas:
@cmroche and @askmike 
per my testing regarding responses for transactions lists Binance sometimes returns single transaction with same ID. To avoid that check for transaction sequence is needed, otherwise BudFox locks into the transaction retrieval loop.
This is the line I'm talking about:
`if (trades.length > 0 && lastTid < _.first(trades).tid) {`

But I've already opened pull request on @cmroche repo so I'm sure he can deliver.

Issue #1632, opened 2018-01-04
Saving `time` in mongoDB consumes lots of disk space


cgav:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[ ] bug report
[ ] feature request
[x] question about the decisions made in the repository

* **Action taken** (what you did)
Started import with mongodb adapter.

* **Expected result** (what you hoped would happen)
`time` field in `EXCHANGE_candles` collection is of type `number`.

* **Actual result** (unexpected outcome)
`time` field in `EXCHANGE_candles` collection is JSON version of `moment().utc()`.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
Serializing a `moment` object for each time is very heavy on memory, where a simple number timestamp would also suffice. Why was it decided to persist the `moment` object?

Here a snippet of how one trade document looks like in mongo:
```
{
  "_id": ObjectId("5a4e8771774fdd004e21f82b"),
  "time": {
    "_isAMomentObject": true,
    "_isUTC": true,
    "_pf": {
      "empty": false,
      "unusedTokens": [ ],
      "unusedInput": [ ],
      "overflow": -2,
      "charsLeftOver": 0,
      "nullInput": false,
      "invalidMonth": null,
      "invalidFormat": false,
      "userInvalidated": false,
      "iso": false,
      "parsedDateParts": [ ],
      "meridiem": null,
      "rfc2822": false,
      "weekdayMismatch": false
    },
    "_locale": {
      "_calendar": {
        "sameDay": "[Today at] LT",
        "nextDay": "[Tomorrow at] LT",
        "nextWeek": "dddd [at] LT",
        "lastDay": "[Yesterday at] LT",
        "lastWeek": "[Last] dddd [at] LT",
        "sameElse": "L"
      },
      "_longDateFormat": {
        "LTS": "h:mm:ss A",
        "LT": "h:mm A",
        "L": "MM/DD/YYYY",
        "LL": "MMMM D, YYYY",
        "LLL": "MMMM D, YYYY h:mm A",
        "LLLL": "dddd, MMMM D, YYYY h:mm A"
      },
      "_invalidDate": "Invalid date",
      "_dayOfMonthOrdinalParse": /\d{1,2}(th|st|nd|rd)/,
      "_relativeTime": {
        "future": "in %s",
        "past": "%s ago",
[...]

PR #1639, opened 2018-01-06
Fix moment utc format

billymcintosh:
Store moment utc as a string rather than a serialized moment object in mongodb.

Bug fix: moment is being serialized and stored in mongo

`moment().utc()` is being serialized and stored in mongodb

Changed to `moment().utc().format()` so we get a string of the datetime.

* **Other information**:
Reported here: https://github.com/askmike/gekko/issues/1632


cmroche (approved):

askmike:
great stuff, thanks!

PR #1655, opened 2018-01-08
Initial Websocket Support - Portfolio Management Enhancements

werkkrew:
Feature

All aspects of the ticker, order management, etc. are handled using exchange REST API's.  This change introduces preliminary support in portfolioManager as well as the GDAX exchange to leverage the websocket feed for certain aspects of trading and portfolio management.

There are 3 new settings.  "websocket", "tradeSpeed", and "tradeTimeout".  If the exchange in use supports websockets and websockets is true, the following behaviorial changes to portfolioManager occur (all of which gracefully fall back to standard REST API methods when WS is unavailable).

1. The ticker is kept up to date from the websocket feed.  Any attempt to trade will use the most up to date websocket orderbook data as opposed to an API call.
2. When a trade is initiated, the order will be managed every "tradeSpeed" ms to ensure it stays at the top of the orderbook.  Once the time in seconds set by "tradeTimeout" has expired, the order will be cancelled and portfolioManager will wait for new advice to start a new trade.

* **Other information**:

To Do:
1. Use websocket stream to continuously monitor order filled status instead of periodically issuing "check order"
2. Additional exchange support

Potential Future Enhancements:
1. Real-time stop-loss, trailing stop-loss
2. Use websocket feed to keep trade / order records for trend analysis between candles
3. Margin trading / shorting


werkkrew:
Another noted change in this PR is that it supports GDAX node library v0.5 which introduces a few changes and enhancements.

okonon:
@werkkrew  this is awesome! What other exchanges offer ws support that you know?

werkkrew:
@okonon I believe most of them do at this point, at least most of the mainstream ones.  There are a number of enhancements that could be added to Gekko by leveraging websockets but many of them are likely outside my current skill level.

akhadzhy:
Hi @werkkrew,
Thank you a lot for perfect job. It seems very exciting! 
I am not so skilled to understand logic from code and how to run it properly. 
But  does I understood right, that now we can add to Live Gekkos (node gekko --ui) with this correction few trade-bots on different pairs on one exchange? So it can run e.g. CCI on binance for ETH/NEO, ETH/TRX at the same time ?

billymcintosh:
@Tabakixan - no.  This simply makes data stream from an exchange to Gekko using WebSockets instead of pulling small bits of data at time.

When I say data, I mean the current trades.  It'll make things a lot faster.

werkkrew:
The last commit doesn't break anything but the newly added feature attempt is non-functional.  

Right now I am working on two things that I am not entirely sure how to solve and I would love to have someone else take a look.

1. When starting an instance of the bot which uses the market watcher or trader plugins: After history is fully synced up, use the websocket feed for all trade data going forward.  

The main barrier for me right now is that the Trader/portfolioManager plugins create an instance of the Exchange module which is separate from the marketFetcher instance of the exchange which the Market Watcher uses.  This potentially produces more than 1 instance of the WS feed and the data structures associated with those feeds and they do not share data.  For memory/resource/performance reasons it would be ideal to share the websocket object between multiple instances of the exchange module OR have the core of Gekko only create one instance and reference it in other modules vs creating new instances.

2. The core/heart (literally core/heart.js) of gekko revolves around scheduled "ticks", which can be configured to as low as 1 second.  Most events related to anything (trades, analysis, checking orders, etc.) happen on the tick interval.  It would be nice to have websocket stream data directly into the database or other data structure in use instead of updating it only on the "tick" interval.  

An example of things that would ideally happen any time as potential interrupts, asynchronously from scheduled ticks:
   - When I have an open order which gets filled, gekko can know nearly instantly via the WS feed vs. waiting until the next tick.
   - Have a stop loss or trailing stop loss than could cancel an order nearly instantly instead of on the next tick or candle.
    - Leverage order book and order data in real time as part of strategies instead of just 1 minute candles, for example sniping mispriced orders.

plugins/trader/portfolioManager.js
```diff
@@ -44,12 +45,43 @@ var Manager = function(conf) {
...
+          this.websocket = false;
+      }.bind(this);
+      if(_.isNumber(conf.tradeSpeed)) {
+        this.tradeSpeed = conf.tradeSpeed;
+      } else {
+        this.tradeSpeed = 100;
+      }
+      log.debug('Websockets enabled, positions will be managed more quickly via websocket data.');
+      log.debug('Trade positions will be managed at ' + this.tradeSpeed + 'ms intervals.');
+      this.exchange.startWebsocket(ws);
```
cmroche, line 81: I don't think this is really the right place for this code. Ideally the exchange shouldn't need to export knowledge that it is a websocket to start the websocket, this should pretty much be implicit to the exchange implementation. Could this be moved to the GDAX constructor?

werkkrew: After working on it for a while, I have gotten sort of lost in what should go where.

Several things access the exchange object, portfolioManager, marketWatcher being the most notable, and I didn't want there to be multiple websocket streams getting opened.

My issue is that I want all of the modules which use an exchange to use the same websocket feed and corresponding arrays / objects created by it.  I got sort of stuck and I have been trying to figure it out / hoping someone with more JS expertise might take a look.

See my previous comment about some thoughts I have, specifically with respect to how the tick system works and in what ways gekko could leverage a websocket stream vs. api calls and how much of the core would have to be changed to accommodate it...

I think I got a little in over my head to be honest.

raiden017: Take at look at Yoyae's repo which uses ccxt to manage the connections to the exchanges. I think gekko would be better if it utilized ccxt as it is literally updated multiple times a day and the dev team can focus on core of the project.

werkkrew: I think using CCXT in gekko would be great, but that isn't really in the scope of this PR.  This PR is attempting to make Gekko capable of using Websocket data from an exchange in a meaningful way.

cmroche: CCXT also does not support websockets.

werkkrew:
I am closing this pull request as it is now in conflict and has not been adequately tested.  I may re-submit it once I am able to do further testing.

PR #1661, opened 2018-01-09
Add Pairs Bitttrex USDT  to ADA, BTG, OMG

eduardoarn:
Just add a few USD trading pairs that I use on Bittrex, the ADA, BTG and OMG
I do not know if I have to change some other file, but if you're sure it's a small contribution.

thank you


askmike:
Thanks!

PR #1666, opened 2018-01-11
Bugfix/fix cp debug port conflict

mronus:
Fixed the EADRRINUSE exception that is thrown when debugging

I see "Child process is died" message alert and EADDRINUSE exception in the console after a child process is started for a process when running the app in debug mode. (same in VS Code and IntelliJ)There are some open discussions in nodejs about this issue. Please see:
https://github.com/nodejs/node/issues/8495#issuecomment-246600965
https://github.com/nodejs/node/issues/9435

This is a temporary fix. I think using `cluster` instead of `child_process` solves the problem as mentioned here https://github.com/nodejs/node/issues/8495#issuecomment-246600965 and as applied here https://github.com/rocjs/roc-package-webpack-node/pull/13


askmike:
:+1:

Issue #1673, opened 2018-01-12
Error handling Binance API response


ansonphong:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Occurred while running a live trader on Binance.

* **Expected result** (what you hoped would happen)

* **Actual result** (unexpected outcome)

After running smooth for about an hour, this error happened and Node crashed:
```
2018-01-12 10:55:23 (DEBUG):    Requested XRP/BTC trade data from Binance ...
C:\...\gekko\exchanges\binance.js:62
    if(!_.isEmpty(body.code)) {
                       ^

TypeError: Cannot read property 'code' of undefined
    at C:\...\gekko\exchanges\binance.js:62:24
    at Request.request [as _callback] (C:\...\gekko\node_modules\binance\lib\rest.js:66:21)
    at self.callback (C:\...\gekko\node_modules\request\request.js:188:22)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at Request.onRequestError (C:\...\gekko\node_modules\request\request.js:884:8)
    at emitOne (events.js:121:20)
    at ClientRequest.emit (events.js:211:7)
    at TLSSocket.socketErrorListener (_http_client.js:387:9)
    at emitOne (events.js:116:13)

```

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

I think this could be due to the Binance API call giving an undefined response, we might want to first check that the `body` is defined before processing it.

ipansensoy:
is there any update about this error?

cmroche:
It's possible that the Binance module changed something about their error handling again, I'll check it out over the weekend.

If that is the case, running `npm install binance@1.0.2` should work around the issue for now.

cmroche:
Ok actually this was simpler than that, a missing null check. I'll push a fix shortly.

ansonphong:
Thanks @cmroche, yes that's how I ended up fixing it locally.

PR #1674, opened 2018-01-12
Fix check for data errors in binance response

cmroche:

Bugfix

Could crash on test for body.code in Binance api response when the response is empty.

Check for the empty body.

* **Other information**:


cmroche:
Should fix #1673

askmike:
:+1:

PR #1683, opened 2018-01-14
Scope update

oleersoy:
Still trying to understand the different between long and short.  Does short mean buy it at the current price only, whereas long means the trader has more flexibility to buy at higher price and should expend the entire budget to get as much as possible?  For example if I have 1 BTC and only 10,000 XLM are available at the current price of 0.000032 BTC, but there's more available at 0.000033 BTC then the long strategy execute two orders one at 0.000032 BTC and one at 0.000033 BTC in order to spend the entire budget?

### DIFF
--- a/docs/introduction/scope.md
+++ b/docs/introduction/scope.md
@@ -16,7 +16,7 @@ The limits of this design are::
 - Strategies do not see trades
 - Strategies cannot look at the order book.
 - Strategies do not know what the current portfolio looks like.
-- Strategies can only trigger LONG or SHORT advice.
+- Strategies can either trigger a `LONG` (Sell all shares) or `SHORT` (Buy up to the current budget allowance).
 
 ## Strategy Execution
 
--- a/docs/introduction/scope.md
+++ b/docs/introduction/scope.md
@@ -1,30 +1,29 @@
 # Scope
 
-Gekko is a free and open source tool that is designed as a starterskit for automated trading on cryptocurrency markets. Gekko aims to have a low barrier entry to writing your own strategies (however a basic scripting knowledge in javascript is required for users who create their own strategies). This page will discuss the scope of some components of Gekko, explaining how things are done and why they are done a certain way.
+Gekko is a free and open source starter kit for automated trading on cryptocurrency markets. Gekko has a low barrier to entry as only minimal knowledge of javascript is required to write trading strategies.
 
 ## Market data
 
-Gekko aggregates all market data into minutely candles (OHLC, VWP and amount of trades). This means Gekko only has to store candles on disk, which take up a predictable amount of space on your harddrive. The tradebot will use some additional market data (the orderbook) to make efficient execute orders efficiently, but this data is not visibile anywhere else.
+Gekko aggregates market data into [OHLC candles](https://en.wikipedia.org/wiki/Open-high-low-close_chart). Gekko only stores candle data on disk taking up a predictable amount of hard drive space. The tradebot will use some additional market data (Such as the orderbook) in order to efficiently execute orders, but this data cached in memory and not accessible.
 
 ## Strategies
 
-Strategies are simple scripts that handle new market data (OHLC candles) as well as calculated indicator results (strategies specify what indicators they want with which settings). Every time there is new data the strategy can determine to signal either LONG or SHORT. That's about it! This very simple design (candles + indicator values go in => signals come out) is quite powerfull.
+Strategies are scripts that receive market data in the form of (OHLC candles) and execute buy / sell orders which could involve the calculation and usage of trading indicators. See [Creating aa Strategy](../strategies/creating_a_strategy.html) for more information.
 
-Unfortunately this simple design can sometimes be limiting, here are some limitations:
+The limits of this design are::
 
-- Since strategies are only fed candles (and indicator values):
- - Strategies cannot act on a smaller timeframe than 1 minute.
- - Strategies cannot see any trades.
- - Strategies cannot look at the orderbook.
+- The smallest time frame a strategy can act on is one minute.
+- Strategies do not see trades
+- Strategies cannot look at the order book.
 - Strategies do not know what the current portfolio looks like.
-- Strategies can only trigger a LONG or SHORT which signals to go "all in".
+- Strategies can only trigger LONG or SHORT advice.
 
-## Execution strategy
+## Strategy Execution
 
-When you are using Gekko for a real tradebot Gekko will create orders at the exchange whenever your strategy signials an advice (long or short). If your strategy signal a long advice Gekko will try to buy as much "asset" as it can get with all your "currency" (if your strat is running on USD/BTC that would mean buying BTC with all your USD). As for creating the orders Gekko is conservative and stays on your side of the orderbook (this means you don not lose on the spread, slippage or taker fees). An example:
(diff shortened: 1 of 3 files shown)


docs/introduction/scope.md
```diff
@@ -1,30 +1,29 @@
...
+Gekko is a free and open source starter kit for automated trading on cryptocurrency markets. Gekko has a low barrier to entry as only minimal knowledge of javascript is required to write trading strategies.
 
 ## Market data
 
-Gekko aggregates all market data into minutely candles (OHLC, VWP and amount of trades). This means Gekko only has to store candles on disk, which take up a predictable amount of space on your harddrive. The tradebot will use some additional market data (the orderbook) to make efficient execute orders efficiently, but this data is not visibile anywhere else.
+Gekko aggregates market data into [OHLC candles](https://en.wikipedia.org/wiki/Open-high-low-close_chart). Gekko only stores candle data on disk taking up a predictable amount of hard drive space. The tradebot will use some additional market data (Such as the orderbook) in order to efficiently execute orders, but this data cached in memory and not accessible.
```
askmike, line 7: Besides OHLC Gekko also exposes VWP and amount of trades metrics, can we keep that in this document?

docs/introduction/scope.md
```diff
@@ -1,30 +1,29 @@
...
 
 ## Strategies
 
-Strategies are simple scripts that handle new market data (OHLC candles) as well as calculated indicator results (strategies specify what indicators they want with which settings). Every time there is new data the strategy can determine to signal either LONG or SHORT. That's about it! This very simple design (candles + indicator values go in => signals come out) is quite powerfull.
+Strategies are scripts that receive market data in the form of (OHLC candles) and execute buy / sell orders which could involve the calculation and usage of trading indicators. See [Creating a Strategy](../strategies/creating_a_strategy.html) for more information.
```
askmike, line 11: it's important to understand that the strategy does not execute any orders, it merely suggests to Gekko what to do (in the form of a position to take). Gekko has different plugins that will actually execute orders based on this advice.

Right now Gekko will always try to buy on long signals an try to sell on short signals, but the main point here is that strategies indicate "positions" to be in in. And selling and buying is right now the only way of moving from one position into another. At some point we will add margin trading which makes this less simple.

docs/introduction/scope.md
```diff
@@ -1,30 +1,29 @@
...
-- Since strategies are only fed candles (and indicator values):
- - Strategies cannot act on a smaller timeframe than 1 minute.
- - Strategies cannot see any trades.
- - Strategies cannot look at the orderbook.
+- The smallest time frame a strategy can act on is one minute.
+- Strategies do not see trades
+- Strategies cannot look at the order book.
 - Strategies do not know what the current portfolio looks like.
-- Strategies can only trigger a LONG or SHORT which signals to go "all in".
+- Strategies can only trigger LONG or SHORT advice.
```
askmike: Why did you remove the "all-in" part? This is one of the questions asked most.

askmike (commented):
See comments.

askmike:
Thanks a lot! I really like everything, except for the short/long stuff. To answer your question:

> Still trying to understand the different between long and short. Does short mean buy it at the current price only, whereas long means the trader has more flexibility to buy at higher price and should expend the entire budget to get as much as possible?

If you are trading BTC/USD:

- short means getting into a position where you do not have any BTC (sell BTC into USD)
- long means getting into a position where you have a lot of BTC (buy BTC from USD)

When your strat signals either of the above Gekko will try to buy or sell as much as possible.

> For example if I have 1 BTC and only 10,000 XLM are available at the current price of 0.000032 BTC, but there's more available at 0.000033 BTC then the long strategy execute two orders one at 0.000032 BTC and one at 0.000033 BTC in order to spend the entire budget?

This is a different question: if your strat signals "LONG" Gekko might try to buy XLM for your BTC. Gekko won't actually buy any XLM for the prices you mentioned, but will place a "bid" order at a lower price and will wait until someone sells into your bid order. This way you get a better price (but will have to wait a bit longer).

docs/introduction/scope.md
oleersoy: Absolutely.  I tried to google VWP and did not find any documents describing what that is, so I thought it might be confusing for others as well, which is why I took it out.  Ideally there would be a link to an API document that shows how to go about utilizing the VWP and trade volume metrics.  That would also make it clear what they are since the description would be detailed in the API JS Doc / TypeDoc or perhaps a something like [Compodoc](https://medium.com/vincent-ogloblinsky/compodoc-documentation-tool-for-angular-2-applications-44ec650e01a8) would work.

docs/introduction/scope.md
oleersoy: My bad.  I should have started some micro issues on topics I'm confused about before updating things that I'm not 100% certain of.  I'll try to clear up my misunderstandings and then hopefully I can write better docs.  Thanks for all your patient help with this.  The tool is really awesome.

askmike:
You are right about these docs, I am reworking the events that flow through
gekko (with proper documentation this time). A candle is one of these
events (the only one strategies can see - plugins can see everything).

If you have more questions feel free to ask them. I am already happy that
you are looking critically at the documentation!

docs/introduction/scope.md
oleersoy: I'm still a little confused WRT `LONG`, `SHORT`, and how orders are triggered in general.  I added an issue here to explore it further: https://github.com/askmike/gekko/issues/1686

askmike:
@oleersoy I cannot merge this mainly because of this comment: https://github.com/askmike/gekko/pull/1683#discussion_r161392697

Strategies really do not execute orders, and I cannot accept changes to the documentation indicating that they do. Even though I really like your other changes. If you can revert that paragraph I can at least accept your other changes.

oleersoy:
Would it be correct to say "Strategies can either trigger a `LONG` (Sell all shares) or `SHORT` (Buy up to the current budget allowance)"?

askmike:
@oleersoy yep that's fine :)

docs/introduction/scope.md
```diff
@@ -1,30 +1,29 @@
...
- - Strategies cannot look at the orderbook.
+- The smallest time frame a strategy can act on is one minute.
+- Strategies do not see trades
+- Strategies cannot look at the order book.
 - Strategies do not know what the current portfolio looks like.
-- Strategies can only trigger a LONG or SHORT which signals to go "all in".
+- Strategies can either trigger a `LONG` (Sell all shares) or `SHORT` (Buy up to the current budget allowance).
 
-## Execution strategy
+## Strategy Execution
```
askmike, line 21: This is definitely not strategy execution, this is the execution strategy: they are 2 different concepts and execution in this one refers to order execution.

docs/introduction/scope.md
askmike: see ^

askmike (changes requested):
Last two things are quite important!

So that's:

- this: https://github.com/askmike/gekko/pull/1683#discussion_r168984524
- and this: https://github.com/askmike/gekko/pull/1683#discussion_r161392697

askmike:
Closing this due to inactivity. I also think there will probably be some merge conflicts to figure out before we look at merging this.

Feel free to comment / push commits and we'll reopen :)

PR #1693, opened 2018-01-15
Minor GDAX fix and update to package 0.5.x

werkkrew:
Bug fix / feature

Using GDAX gives an error related to the new import changes.

Fixed the minor bug and also updated to support/require official GDAX 0.5.x module (coinbase/gdax-node) which required a couple of minor changes to the api calls to clear out deprecation warnings.

* **Other information**:


package.json
```diff
@@ -29,7 +29,7 @@
     "co-fs": "^1.2.0",
     "coinfalcon": "^1.0.3",
     "commander": "^2.9.0",
-    "gdax": "^0.4.2",
+    "gdax": "github:coinbase/gdax-node",
```
askmike, line 32: I think this requires git to be installed, Gekko's installation (using npm) should also work without Git. I'll check this asap.

askmike (commented):
see comment.

package.json
werkkrew: You might be right, I installed it this way based on information in their official repository.  The command was npm install coinbase/gdax-node. 

I am not sure why they haven't updated the other npm package.

werkkrew:
Closing as the primary issue I was trying to fix is resolved and updating to gdax-node from git is not ideal.

PR #1699, opened 2018-01-16
Binance processes the manually canceled order as filled

mronus:
Bug fix

Binance processes the manually canceled order as filled as seen in the following ss.
image

CANCELED, REJECTED and EXPIRED statuses are accepted as not FILLED.


mronus:
@cmroche I know you have more experience than me about binance. It would be good if you can check this.

cmroche (commented):
This looks reasonable to me, cancelled is spelt inorrectly though, is that the case in the API or a typo here?

mronus:
It is in the API. You can see it in logs i posted in the main comment, and here as well: https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#enum-definitions

cmroche:
I wonder if this feature should be implemented as a type of configurable option though, does anyone have a case where a manually cancelled action shouldn’t be handled this way?

Perhaps putting a new config block in the config file, called exchange, which allows us to specify exchange specific options? There has already been discussions in the past about using this for things like post only order support, or hidden order features on exchanges that support the option. What do you think?

This approach would probably disambiguate changes in behaviour amongst exchanges too.

mronus:
It would be really good to implement such a feature but currently, I think this is not a feature but a bug. The canceled, rejected and expired orders are considered as filled orders, and this is reflected in the trade as %100 gain or loss.

I think we can focus on a generic way, maybe an interface, to handle statuses after fixing this. I didn't check how it can be done but I believe that we need a better mechanism to handle statuses to not get similar exceptions. e.g #1703

askmike:
I agree that this PR is the right behaviour for now. Keep in mind that this does mean Binance is the only exchange where it is implemented like this. In the future some kind of interface (or a status number) could be used, but I also wouldn't know what the best way to handle a (manually) cancelled order would be..

Issue #1678, opened 2018-01-13
Binance new assets


gen4sp:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Trying to import some assets history data from Binance

* **Expected result** (what you hoped would happen)
All assets list should be present

* **Actual result** (unexpected outcome)
There aren't a few new assets

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
How to add new assets? Or maybe there is a way to fetch and use assets automatically?

cryosx:
edit exchanges/binance.js

add you asset to the list of assets and then add your pair down below it with the rest of the pairs.

Pair info: https://www.binance.com/exchange/public/product
tickSize -> precision
minTrade -> amount

askmike:
I would love a PR once you added some!

mronus:
PR created #1700

PR #1550, opened 2017-12-24
Use script to generate market data for binance, karken and bitfinex and exchange function for lot sizing

cmroche:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Feature

* **What is the current behavior?** (You can also link to an open issue here)

Product entries are done by hand, for exchanges with many assets this is tedious and prone to human error.

getMinimumSize on portfolioManager doesn't handle Binance's lot size requirements correctly result in occasional errors when posting small orders.

* **What is the new behavior (if this is a feature change)?**

Binance market data is now moved to a second .js file, which is updateable with a script. The script parses the product definitions for all available pairs and adds them to the product list with all necessary information.

Added the function getLotSize to the exchange and updated portfolioManager to check for this function and use it is available for computing the correct lot size of an order. General rules are that if we are blow the minimum, we don't trade. This will allow any exchange to define necessary behaviour for lot sizing to that specific exchange.

* **Other information**:

If we are happy with this approach for managing market data I'll apply the change to Kraken and Bitfinex as well.

PR #1700, opened 2018-01-16
Updated binance market data using script provided in PR #1550 by @cmroche

mronus:
Fix for #1678

Updated binance market data using the script provided in PR #1550 by @cmroche


cmroche (approved):

askmike:
:+1: thanks :)

nissimpradhan:
Can binance market data using the script be done automatically when using import option form the Gekko UI? I think that would be great. If it hasn't been done, l'd love to do it myself.

askmike:
No, you have to manually run the update file.

The reason we do it like this is so that we can make sure Gekko only works with markets that we have tested and seen. We could also dynamically look up markets whenever you start Gekko, but that defeats the purpose.

nissimpradhan:
What do you mean by tested and seen? Binance keeps adding new coins to the exchange, Gekko should support those trading pairs isn't it? I thought that was the purpose of the script.

It's not a ideal to run it every time we use the importer but perhaps, a "sync currency pairs" button in the import page definitely helps.

askmike:
The problem is that there are differences in how they report their markets
and how they actually are. We are not handling it yet but we will, see
https://github.com/askmike/gekko/pull/1805

askmike:
And honestly using gekko on brand new coins does not make sense. Gekko is
for trendwatching and momentum reversal. You can't do that on brand new
markets imo.

jbghub:
I am new to Gekko. This discussion is very confusing. I would like to see more coins available in Gekko for trading listed by Bianance. If an update has been prepared that will accomplish this I would like to know how to install the update. Thanks

stereohelix:
jbghub, here's a way to update for more coins without having to use any other tools:

-Make a backup copy of gekko/exchanges/binance.js
-Then open your current file, copy the contents of this into it, and save:
https://github.com/askmike/gekko/blob/c3de6f653930b1a8a58ef5071cfd0537664dc7d3/exchanges/binance.js

jbghub:
Thank you so much. I will implement later this evening.

jbghub:
I updated the binance.js file as stated. However, I am still only seeing the same 5 coins in the Gekko app. Any suggestions?

stereohelix:
Stop and restart gekko?

jbghub:
Yes, Stopped, Restarted Gekko; also rebooted PC. Any other ideas?

jbghub:
I am using the Admin Command window to start Gekko:
node gekko --ui
then going to http://localhost:3000/#/home in Chrome browzer.

stereohelix:
Hi jbghub,
Sorry, now I'm confused about this. I don't know how you could only show five coins after restarting, since the binance.js file is the only file where that information is read from. This might be beyond my ability to hack. Are you sure you edited the correct file? The only other thing I could suggest is restoring the files, pulling gecko again fresh (using git pull) and then trying again.

Good luck, hope that helps.

jbghub:
I found this code in the package-lock file:
"binance": {      "version": "1.3.1",      "resolved": "https://registry.npmjs.org/binance/-/binance-1.3.1.tgz"

Could this have anything to do with the binance file not working for me?

stereohelix:
I'm not sure. But that is a little out-of-date.

You could try 
npm uninstall binance
npm install binance

But I should really add a disclaimor that I'm in over my head here -- if this causes some other conflict, you might end up having to reinstall gekko. Which might be fine, and might actually solve whatever issue is happening.

jbghub:
I believe I found my problem. I was only looking at the USDT pairs. Their are many more coins for other currencies (coins) instead of USDT.

jbghub:
Thanks for all your help.

stereohelix:
Great!

jbghub:
Quesion: How did you determine the Precision of each coin?

stereohelix:
The easiest way is to look on the Binance website. The precision is the least significance digit of the price.

cmroche:
@stereohelix @jbghub Right now it happens to be the least significant digit of each coin, but Binance does publish somewhere in their documents and official list, and it is queryable through their API for official values as there is no guarantee of this relationship in the future.

stereohelix:
Here's the official document (it's actually pretty up-to-date, not 8 months old)
https://support.binance.com/hc/en-us/articles/115000594711-Trading-Rule

PR #1711, opened 2018-01-17
Fix kraken minimum order quantities

talentoscope:
Bug fix

Does not allow smaller orders

This fixes the current kraken order minimum quantities

* **Other information**:
See https://support.kraken.com/hc/en-us/articles/205893708-What-is-the-minimum-order-size-


askmike:
:+1:

PR #1736, opened 2018-01-20
Kraken: update CAD/ETH and USD/ETH min amount and precision

fouman75:
FIX CAD/ETH and USD/ETH min order amount and precision based on Kraken's updated info

Kraken api fails due to low volume

After updating values, no more low volume failures on CAD/ETH.

* **Other information**:
Updating CAD/ETH and USD/ETH based on the following information:

#### min amount
https://support.kraken.com/hc/en-us/articles/205893708-What-is-the-minimum-order-size-

#### precision
https://blog.kraken.com/post/1261/announcement-reducing-price-precision-on-all-trading-pairs/


askmike:
:+1: thanks

PR #1750, opened 2018-01-21
Manual update of market data (Bitfinex, Kraken)

cmroche:
Using the results from PR #1550 to update market data for Kraken and Bitfinex.


askmike:
@cmroche apologies about #1550, it completely escaped my attention. I think you rather have that merged than this, so see my comment there.

If there are more PRs you think deserve some priority let me know. There is a massive backlog of issues and PRs and it will take me a while to clear.

cmroche:
@askmike No problem at all, I will close this. #1550 is definitely preferred. I'll also ping you if there is anything else I see that is important (either mine, or someone else changes) in the future.

Thanks

PR #1791, opened 2018-01-26
more intuitive naming

dmitriz:
More intuitive variable naming

The same as before

No change in behaviour, only renaming the variables

* **Other information**:
I had found it very hard to read the code due to inconsistent naming.
This is an attempt of small improvement to ease the reading.


askmike:
:+1: thanks

Issue #1689, opened 2018-01-14
[GDAX] Strategy Execution Delayed or Never Starts


werkkrew:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Started a cli instance of gekko against a working config file

* **Expected result** (what you hoped would happen)
Gekko fills in the needed history to meet "requiredHistory" and starts the strategy.

* **Actual result** (unexpected outcome)
Gekko just collects new trades during the present timeframe but does not start the strategy. Occassionally after some time (10-30 minutes) the strategy will start, others it seems to go on like this indefinitely.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

```
    /      \ /        |/  |  /  |/  |  /  | /      \
  /$$$$$$  |$$$$$$$$/ $$ | /$$/ $$ | /$$/ /$$$$$$  |
  $$ | _$$/ $$ |__    $$ |/$$/  $$ |/$$/  $$ |  $$ |
  $$ |/    |$$    |   $$  $$<   $$  $$<   $$ |  $$ |
  $$ |$$$$ |$$$$$/    $$$$$  \  $$$$$  \  $$ |  $$ |
  $$ \__$$ |$$ |_____ $$ |$$  \ $$ |$$  \ $$ \__$$ |
  $$    $$/ $$       |$$ | $$  |$$ | $$  |$$    $$/
   $$$$$$/  $$$$$$$$/ $$/   $$/ $$/   $$/  $$$$$$/

        Gekko v0.5.12
        I'm gonna make you rich, Bud Fox.


2018-01-14 13:59:43 (INFO):     Setting up Gekko in realtime mode
2018-01-14 13:59:43 (INFO):
2018-01-14 13:59:43 (INFO):     Setting up:
2018-01-14 13:59:43 (INFO):              Candle writer
2018-01-14 13:59:43 (INFO):              Store candles in a database
2018-01-14 13:59:43 (INFO):

2018-01-14 13:59:43 (INFO):     Setting up:
2018-01-14 13:59:43 (INFO):              Trading Advisor
2018-01-14 13:59:43 (INFO):              Calculate trading advice
2018-01-14 13:59:43 (INFO):              Using the strategy: private-uber
[...]

PR #1793, opened 2018-01-26
Subtract candle length from start time to calculate isPremature

ansonphong:
Bug fix / feature

Gekko wasn't starting to trade on the first live candle, it was delaying a full candle before starting trades on the live market.

This causes it to start trading on the first live candle.

* **Other information**:
More info at : https://github.com/askmike/gekko/issues/1689#issuecomment-360933359


askmike:
I agree with the changes, but would you mind putting them inside the ifcheck? Else they might degrade performance of the backtester.

ansonphong:
@askmike Yea I see no problem, I'll move that in.

askmike:
Thanks!

PR #1794, opened 2018-01-27
Use declared var instead of undefined

dmitriz:
Bug fix

The variable `options` is unused.
The undefined `operation` is used instead

The `options` variable is now used correctly

* **Other information**:

### DIFF
--- a/core/util.js
+++ b/core/util.js
@@ -183,8 +183,8 @@ var util = {
       minTimeout: 1 * 1000,
       maxTimeout: 3 * 1000
     };
- 
-    retryHelper(fn, operation, callback);
+
+    retryHelper(fn, options, callback);
   },
   retryCustom: function(options, fn, callback) {
     retryHelper(fn, options, callback);


askmike:
Nice catch, bug was introduced here #1521.

Thanks for fixing!

dmitriz:
You are welcome.

It looks somewhat confusing in the `utils` with the hard-coded options,
among other more abstract methods.

Maybe move those options directly to the mailer, if that is where they are used?

askmike:
This is about retry functionality used in a lot of different places: whenever any piece needs to retry something more than once this code is used. But the `utils` module has grown quite big over time (exactly what you are referring too). This will be split up at some point.

dmitriz:
I could find `utils.retry` used only in one place in `mailer.js` but maybe it is called differently?

Other places I found use `util.retryHelper` which relies on the external `retry` package 
and does not seem to use this local `retry` method.

Possibly I am missing something but it looks confusing when trying to understand what the code is doing.

PR #1799, opened 2018-01-27
Fix varPPO strategy and TSI indicator

Paulovms:
Bug fix.

TSI values on varPPO and TSI strategies stay above 95 after a lot of candles.
The first momentum value is the full close value and this put a big error on next EMAs.
varPPO use the PPO value from this.indicators.ppo.ppo to calculate PPO histogram, but this value is not been updated and are always 0. In fact, since Gekko 0.5.11, the PPO indicator exposes the PPO histogram directly to the API on this.indicators.ppo.result.ppohist.

TSI indicator calculate the first value after get the first previous close and varPPO get ppoHist directly from PPO indicator.


exchanges/bitfinex.js
```diff
@@ -39,7 +39,7 @@ var retryForever = {
 };
 
 // Probably we need to update these string
-var recoverableErrors = new RegExp(/(SOCKETTIMEDOUT|TIMEDOUT|CONNRESET|CONNREFUSED|NOTFOUND|StatusCodeError: 429|StatusCodeError: 5)/)
+var recoverableErrors = new RegExp(/(SOCKETTIMEDOUT|TIMEDOUT|CONNRESET|CONNREFUSED|NOTFOUND|429|443|5\d\d)/g);
```
askmike, line 42: Is this supposed to be part of this PR?

askmike (commented):
see question, will merge after you can let us know :)

exchanges/bitfinex.js
Paulovms: I’m sorry, this will update the Bitfinex string. Maybe I’ll submit another PR.

askmike: Maybe it needs to be updated, did you add this specifically because it caught errors better?

Paulovms: Yes, this string caught erros better.It can be committed.

askmike:
thanks!

PR #1805, opened 2018-01-29
Fix Binance ICX/BNB minial order amount is 1

namns1412:
Fix Binance ICX/BNB minial order amount is 1

Current amount is 0.01

Current amount is 1

* **Other information**:
Fix strat stop when try to sell under 1 ICX

### DIFF
--- a/exchanges/binance-markets.json
+++ b/exchanges/binance-markets.json
@@ -1557,7 +1557,7 @@
         "ICX"
       ],
       "minimalOrder": {
-        "amount": 0.01,
+        "amount": 1,
         "price": 0.00001,
         "order": 1
       }


askmike:
@namns1412 you might be correct, but that JSON is generated by this file:

https://github.com/askmike/gekko/blob/develop/util/genMarketFiles/update-binance.js

Which we run occasionally to add new market support to Gekko. That file gets it's data from Binance itself ([see this link](https://www.binance.com/exchange/public/product)). So approving this PR wouldn't fix things for long. Would you mind also adding some the correction here as well:

https://github.com/askmike/gekko/blob/develop/util/genMarketFiles/update-binance.js#L42

so that we can apply that fix everytime we generate market files?

cc @cmroche.

cmroche:
@askmike @namns1412 Try regenerating the product list first to make sure the markets were not updated by their API. Last time I saw a descrepency like this it required it was due to an update.

If that doesn't fix the issue, then do as askmike suggested and put an exception in the update script.

n3us3ns3:
<iframe width="854" height="480" src="https://www.youtube.com/embed/3l2_KpfCcmA" frameborder="0" allow="autoplay; encrypted-media" allowfullscreen></iframe>

cmroche:
I reported the above post from @n3us3ns3 as spam, would you please remove it @askmike

PR #1806, opened 2018-01-29
Binance cancelOrder Fix

werkkrew:
Bug Fix

Portfolio Manager expects the exchange cancelOrder method to return true if the order has been filled.  The current Binance implementation does not do that.

The exchange cancelOrder method will return true if the order is not found (generally means it was filled)

* **Other information**:
This is probably a more robust way to ensure the order was filled, such as checking the order status with getOrder in the cancelOrder method to ensure it was in fact filled before returning true, but in my testing if the order id was valid, and binance returns the order was not found, it was filled, so the code I am publishing here does work but could be improved.

@cmroche @askmike


exchanges/binance.js
```diff
@@ -297,9 +297,15 @@ Trader.prototype.checkOrder = function(order, callback) {
 };
 
 Trader.prototype.cancelOrder = function(order, callback) {
+  // callback for cancelOrder should be true if the order was already filled, otherwise false
   var cancel = function(err, data) {
     log.debug(`[binance.js] entering "cancelOrder" callback after api call, err ${err} data: ${JSON.stringify(data)}`);
-    if (err) return callback(err);
```
askmike, line 302: I agree this is wrong: it shouldn't return the error. However I am not sure if there is an error `data.msg` is guaranteed to be set (if data is undefined this check will throw an error). Did you maybe manually cancel the order or so to get that response?

exchanges/binance.js
```diff
@@ -164,25 +164,25 @@ Trader.prototype.getFee = function(callback) {
 };
 
 Trader.prototype.getTicker = function(callback) {
-  var setTicker = function(err, data) {
-    log.debug(`[binance.js] entering "getTicker" callback after api call, err: ${err} data: ${JSON.stringify(data)}`);
-    if (err) return callback(err);
+    var setTicker = function(err, data) {
```
askmike: could you stick to 2 spaces indenting? Thanks!

askmike (commented):
I have a question! See above.

exchanges/binance.js
werkkrew: My editor is set to 4 space and sometimes it overrides the existing indenting, will fix.

exchanges/binance.js
werkkrew: In my experience running against binance for the last few days when it tries to cancel an order which was filled it always returns the "unknown_order" error, that said, there are other error cases that I did not account for as well as the possibility that an unknown_order error is not actually because the order was filled.  But assuming the upstream code is sound, it should never try to cancel a non-existent order.  As I mentioned in the PR notes, the likely correct way to do it is to have a getOrder or checkOrder call in the cancel order method to ensure the order was not filled before trying to cancel it, but this PR seems to work as a quick fix.

askmike: I've been getting a lot of cancels as well:

    2018-01-29 15:13:09 (INFO):     Starting to watch the market: Binance BNB/BTC
    2018-01-29 15:15:09 (INFO):     Trader Received advice to go long. Buying  BNB
    2018-01-29 15:15:09 (WARN):     Wanted to buy BNB but the amount is too small  (0.01014053 @ 0.00115250 ) at binance
    2018-01-29 15:25:10 (INFO):     Trader Received advice to go short. Selling  BNB
    2018-01-29 15:25:10 (INFO):     Attempting to SELL 16.36 BNB at binance price: 0.0011551
    2018-01-29 15:26:10 (INFO):     SELL order was not (fully) filled, cancelling and creating new order
    2018-01-29 15:26:10 (INFO):     Attempting to SELL 16.36 BNB at binance price: 0.0011547
    2018-01-29 15:27:10 (INFO):     SELL was successfull
    2018-01-29 15:27:10 (INFO):     (PROFIT REPORT) original simulated balance:      0.01887761 BTC
    2018-01-29 15:27:10 (INFO):     (PROFIT REPORT) current simulated balance:       0.01888588 BTC
    2018-01-29 15:27:10 (INFO):     (PROFIT REPORT) simulated profit:                0.00000827 BTC (0.04378277%)
    2018-01-29 16:35:11 (INFO):     Trader Received advice to go long. Buying  BNB
    2018-01-29 16:35:11 (INFO):     Attempting to BUY 16.43 BNB at binance price: 0.0011481
    2018-01-29 16:36:11 (INFO):     BUY order was not (fully) filled, cancelling and creating new order
    2018-01-29 16:36:11 (INFO):     Attempting to BUY 16.43 BNB at binance price: 0.0011486
    2018-01-29 16:37:11 (INFO):     BUY was successfull
    2018-01-29 16:37:11 (INFO):     (PROFIT REPORT) original simulated balance:      0.01887761 BTC
    2018-01-29 16:37:11 (INFO):     (PROFIT REPORT) current simulated balance:       0.01889297 BTC
    2018-01-29 16:37:11 (INFO):     (PROFIT REPORT) simulated profit:                0.00001536 BTC (0.08135346%)
    2018-01-29 19:45:14 (INFO):     Trader Received advice to go short. Selling  BNB
    2018-01-29 19:45:14 (INFO):     Attempting to SELL 16.44 BNB at binance price: 0.0011521
    2018-01-29 19:46:14 (INFO):     SELL order was not (fully) filled, cancelling and creating new order
    2018-01-29 19:46:14 (INFO):     Attempting to SELL 16.44 BNB at binance price: 0.0011519
    2018-01-29 19:47:14 (INFO):     SELL order was not (fully) filled, cancelling and creating new order
    2018-01-29 19:47:14 (INFO):     Attempting to SELL 16.44 BNB at binance price: 0.0011517
    2018-01-29 19:48:14 (INFO):     SELL was successfull
    2018-01-29 19:48:14 (INFO):     (PROFIT REPORT) original simulated balance:      0.01887761 BTC
    2018-01-29 19:48:14 (INFO):     (PROFIT REPORT) current simulated balance:       0.01894989 BTC
    2018-01-29 19:48:14 (INFO):     (PROFIT REPORT) simulated profit:                0.00007228 BTC (0.38289654%)
    2018-01-29 19:48:14 (INFO):     (ROUNDTRIP) entry date (UTC)    exit date (UTC)         exposed duration        P&L     profit
    2018-01-29 19:48:14 (INFO):     (ROUNDTRIP) 50049-07-16 19:07   50049-11-26 11:20       4 months                0.00    0.41

Good to see you finding the cause of those.

askmike: > but this PR seems to work as a quick fix.

Would you mind replacing `if(data.msg === 'UNKNOWN_ORDER')`  with `if(_.get(data, 'msg') === 'UNKNOWN_ORDER')`? Just so that if there is an other error and no data Gekko won't crash.

werkkrew: done

cmroche: @werkkrew Even if you added a checkOrder call before cancelling it, the condition could still happen. For that reason I would argue that we actually have to handle this case in the cancelOrder to at least benefit from a reduced round-trip (or api rate-limit).

cmroche (approved):

askmike:
@werkkrew thanks for the changes! If you can only clear the whitespace I will merge.

> Even if you added a checkOrder call before cancelling it, the condition could still happen.

Exactly: what we should do is let the cancel method return a specific value (or type of error) if the order was not cancelled because it was filled in the meantime. This is the only way to properly handle it imo. However this requires updating the [exchange spec](https://gekko.wizb.it/docs/extending/add_an_exchange.html#cancelOrder) and having to update all exchange implementations. Let's move that to another PR ;)

werkkrew:
@askmike should be all set thanks

askmike:
@werkkrew thanks!!

PR #1808, opened 2018-01-29
GDAX cancelOrder fix and misc other fixes

werkkrew:
Big Fix / Feature

Same as PR #1806 cancelOrder does not detect filled orders. 

Calls to cancelOrder that result in order not found error will presume the order was filled and return true to portfolioManager

Additional minor feature add is to have the ability to enable/disable sandbox using a config value instead of a hard coded boolean in the exchange file itself.

* **Other information**:
Also same as #1806 I am not 100% this approach is 100% as robust as it could/should be but it works as a stop-gap.

@askmike @cmroche


askmike:
ACK all your changes, would you mind reverting all the whitespace changes though? (2 space indents and the likes)? Thanks!

werkkrew:
Cleaned up the whitespace a good bit to clarify the changes introduced.

askmike:
:+1: thanks!

exchanges/gdax.js
```diff
@@ -204,10 +213,15 @@ Trader.prototype.getOrder = function(order, callback) {
...
 Trader.prototype.cancelOrder = function(order, callback) {
-  var result = function(err, data) {
-    // todo, verify result..
-    callback();
-  };
+    // callback for cancelOrder should be true if the order was already filled, otherwise false
+    var result = function(err, data) {
+        if(err) {
+            log.error('Error cancelling order:', err);
+            callback(true);  // need to catch the specific error but usually an error on cancel means it was filled
```
cmroche, line 220: If you don't return here you will also call callback(false).

cmroche (commented):
Small bug

exchanges/gdax.js
werkkrew: Since it was already merged can you implement this fix?

werkkrew: I updated my branch with the returns...

PR #1809, opened 2018-01-29
Added new "private" prefix to gitignore to ignore private strategy files

werkkrew:
Minor Feature

There is no gitignore wildcard for strategies, many people have strategies they do not want to make public. 

Naming a file in the strategies/ or config/strategies/ folders with a prefix of "private-" will ignore them for the purposes of git.

* **Other information**:


askmike:
:+1:

We could add this to the docs at some point.

PR #1813, opened 2018-01-29
Pushbullet Plugin Rewrite

werkkrew:
Pushbullet Rewrite / Minor Bugfix/cleanup to portfolioManager

The current pushbullet plugin is almost 2 years old and seems to only be useful for notifying on advice.

I essentially completely re-wrote Pushbullet.js to notify on actual trades as well as behave similar to the performanceAdvisor in that it will track and report on round trips / profits / value.

I modified this to suit my needs and figured I would introduce a PR for those changes since the existing plugin was so old and hadn't been updated in ages.

* **Other information**:
There is a change in portfolioManager plugin that may not belong in this PR but I found that both in pushbullet and elsewhere the trades that get emitted by portfolioManager had a trade.date field that was "corrupt" for lack of a better term.  The change I made seem to produce an actual usable and correct date in the date field of trades emitted by portfolioManager whereas the way it was did not produce usable dates.

There are also a few other text formatting/logging changes in portfolioManager.js


plugins/pushbullet.js
```diff
@@ -1,88 +1,186 @@
...
-            this.mail(title, body);
-        }else{
-            log.debug('Skipping Send message on startup')
-        }
-    };
-    setupPushBullet.call(this)
+  var setupPushBullet = function (err, result) {
+    if (config.pushbullet.startMessage) {
+      var subject = this.title + ' Started!'
+      var body = "Gekko has started and is trading LIVE on the " + this.exchange.toUpperCase() + " " + this.asset + "/" + this.currency + ' market using the strategy [' + this.strategy + ']';
```
askmike: We do not know for sure whether Gekko is "trading LIVE", it might be a paper trader.

plugins/pushbullet.js
```diff
@@ -1,88 +1,186 @@
...
+    body += '\n[' + this.currency + ']: ' + this.current.currency.toFixed(this.digitsCurrency); 
+    body += '\n[' + this.asset + ']: ' + this.current.asset.toFixed(this.digitsAsset);
+    body += '\nTotal Value: ' + balance.toFixed(this.digitsCurrency);
+    this.notify(subject, body);
+    this.prevAction = trade.action;
+    this.logRoundtripPart(trade);
+  }
+}
+
+Pushbullet.prototype.logRoundtripPart = function(trade) {
```
askmike, line 101: This code is recalculating everything that is being calculated by this plugin: https://github.com/askmike/gekko/blob/develop/plugins/performanceAnalyzer/performanceAnalyzer.js

If roundtrip information is needed by this plugin it should be exposed by the plugin above and fed into this one.

plugins/pushbullet.js
```diff
@@ -1,88 +1,186 @@
...
+        log.info('Pushbullet Message Sent')
+    }
+  });
 };
 
-module.exports = Pushbullet;
\ No newline at end of file
+module.exports = Pushbullet;
+
+function secToTimeString(duration) {
```
askmike, line 170: you can also use moment for this: http://momentjs.com/docs/#/durations/

askmike (changes requested):
This PR copies a lot of code from the performanceAnalyzer. I don't think that is the way to go: instead we should expose data calculated there to plugins. See #1814 for a (start of a) discussion on how to do that.

plugins/pushbullet.js
werkkrew: Agree, but I wanted this information now, so rather than start to introduce big changes to performanceAnalyzer I re-did the work here.

werkkrew:
Given that you have a number of other changes you want to do the performanceAnalyzer upstream that would make this redundant should I just close this PR?

plugins/pushbullet.js
werkkrew: Removed the word "LIVE" from the message

plugins/pushbullet.js
werkkrew: Yes, although I am unable to understand if the duration or duration.humanize() method can split it into multiple units of time like X hours Y minutes Z seconds, or if it can only work against 1 unit of measurement so I made a little helper function.

werkkrew:
Closing this in lieu of the upstream changes that can make it cleaner.

Issue #1812, opened 2018-01-29
DEMA indicator is not DEMA


AidasK:
* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
I have looked at dema indicator implementation and it is wrong.

* **Expected result** (what you hoped would happen)
It should be implemented as:
https://www.investopedia.com/ask/answers/121814/what-double-exponential-moving-average-dema-formula-and-how-it-calculated.asp

* **Actual result** (unexpected outcome)
https://github.com/askmike/gekko/blob/develop/strategies/indicators/DEMA.js

I suggest renaming current DEMA strategy. It currently calculates percentage of price change.

askmike:
> I suggest renaming current DEMA strategy.

We can also fix it, it shouldn't be too hard to implement what is suggested on that page. Does anyone feel like fixing it?

askmike:
I have offered a bounty of 0.003 BTC to whoever can fix this. See here for details and specific requirements: https://forum.gekko.wizb.it/thread-93.html

cloggy45:
@askmike Do you have the excel sheet that you used to originally calculate the formulas?

rkingy:
@askmike I've also created a new indicator here which from my back tests seems to be accurate. #1846

PR #1822, opened 2018-01-31
Fix Dema

Paulovms:
Bug fix.

#1812

It fixes DEMA formula.

* **Other information**:


askmike:
Great stuff, if this works you can claim this bounty: https://forum.gekko.wizb.it/thread-93.html

Quick question: where did you get the test data from?

Paulovms:
Thank you! 
First I've tried Excel to get the test data, but I'd got errors, then I got the values in another PR. 
I'd also think about get the test data from R.

config/strategies/DEMA.toml
```diff
@@ -1,6 +1,5 @@
-short = 10
-long = 21
```
AidasK, line 2: Have you wrote this strategy based on https://www.investopedia.com/articles/trading/10/double-exponential-moving-average.asp ? Because parameters should be long and short IMO

strategies/indicators/DEMA.js
```diff
@@ -1,27 +1,22 @@
...
 
 var Indicator = function(config) {
-  this.input = 'price'	
+  this.input = 'price'
   this.result = false;
-  this.short = new EMA(config.short);
-  this.long = new EMA(config.long);
+  this.inner = new EMA(config.weight);
+  this.outer = new EMA(config.weight);
+  this.sma = new SMA(config.weight); // It will be used in the strategy
```
AidasK: Indicator should not include any code needed for a strategy. How should I suppose to use this indicator later in my own strategies? 0 reusability

AidasK (changes requested):

config/strategies/DEMA.toml
Paulovms: Have you read this https://forum.gekko.wizb.it/thread-93.html ? “updates the DEMA strategy to only use a single period n”

AidasK: Yes and I don't know why he wrote this. I think DEMA should have same parameters as MACD, because DEMA supposed to be same thing as MACD but improved. We should wait for a comment from @askmike

Paulovms: You can make a PR with DEMA and MACD strategy.

strategies/DEMA.js
```diff
@@ -24,21 +25,26 @@ method.update = function(candle) {
...
-
-  log.debug('calculated DEMA properties for candle:');
-  log.debug('\t', 'long ema:', dema.long.result.toFixed(8));
-  log.debug('\t', 'short ema:', dema.short.result.toFixed(8));
-  log.debug('\t diff:', dema.result.toFixed(5));
-  log.debug('\t DEMA age:', dema.short.age, 'candles');
+  let dema = this.indicators.dema;
+  let sma = this.indicators.sma;
+  
+  log.debug('calculated DEMA and SMA properties for candle:');
```
greenbigfrog: capital `C`

greenbigfrog (changes requested):
👍 
(I'd recommend you look into how to use git properly though with proper commit messages etc instead of the web editor 😄 )

strategies/DEMA.js
Paulovms: I will learn! 😉

config/strategies/DEMA.toml
askmike: @AidasK are you sure? Looking at the first picture in your link:

> This one-minute chart of the e-mini Russell 2000 futures contract shows two different double exponential moving averages; a 55-period appears in blue, a 21-period in pink.

eg. most of the time when people use the DEMA indicator they use multiple. I was assuming [this definition](https://en.wikipedia.org/wiki/Double_exponential_moving_average):

askmike: @AidasK If my assumption is wrong (it might just be) let's fix that in a separate PR: this one is the solution to a bounty with specific requirements, if these turn out to be wrong this solution is still valid.

askmike:
@Paulovms thanks a lot, this looks great!

Final question before I can merge (as @greenbigfrog noted): can you clean up the commits a bit? You can do this by squashing all commits into a single one, read here how: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

If you can't I can also do this, but in that case git would reflect my as the author and these commits won't count towards your contribution (but we all know you did it and you'd still be the person to complete the bounty obviously).

config/strategies/DEMA.toml
AidasK: @askmike DEMA indicator implementation is correct. It has only one variable. But I was talking about the strategy. Strategy should be configurable with long and short parameters, not only weight. Though, this is just my opinion how it should work.

Paulovms:
@askmike I have tried to clean up the commits. If this doesn't work you can do it.

askmike:
Done! Now the last thing I need is your bitcoin address to payout the bounty :)

Paulovms:
@askmike  Thank you! The address is 38YTsUrcxPxwxf8UnQewjF87pBDDnjxoEx
https://gist.github.com/Paulovms/aaa277d249ba2a147e23b359404c9df4

askmike:
Bounty send: https://live.blockcypher.com/btc/tx/0744f7076f3820898aca87dce31f53962e8b2e80d4956d918792409c4f4f1a5d/

glegarand:
Will the current DEMA (stable) strategy be featured on the next version or is there any simple way to do it with another strategy ?

PR #1831, opened 2018-01-31
Allow correct print of Entryr/Exit for exchanges that use millisecond time stamps

EToreo:
Enhancement  to allow table to show correct entry and exit times for Binance and other exchanges that use milisecond timestamps

Incorrect display of time stamps when exchange is Binance or any other exchange that uses millisecond timestamps

Correct display of time stamps when exchange is Binance or any other exchange that uses millisecond timestamps

* **Other information**:


EToreo:
Oops!  Ignore this PR, please.  Problem exists in live gekkos, not backtester.  This is a data plumbing issue.

PR #1492, opened 2017-12-16
Limit trading

Ali1:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Feature

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**
Possibility to configure the maximum amount of fiat currency available to gekko to trade with

* **Other information**:
#937

Just an idea for an implantation. What do you think?

PR #1834, opened 2018-02-01
Limit Trading Feature

Ali1:
Feature

* **What is the new behavior (if this is a feature change)?
Possibility to configure the maximum amount of fiat currency available to gekko to trade with.
For example if you have 1 BTC and 100 USD in an exchange account which gekko is performing BTC/USD trading on, you can set currencyAmountAvailableToTrade in your config file to something like 50 USD. If a buy comes in, only 50 USD is used and further buys are prevented. If sells occur, all USD proceeds get added to the available tradable balance ready for the next BUY order. In this way, the other 50 USD of your exchange balance remains untouched.

I don't use the web version and haven't looked into it's code so it is not compatible with it.

* **Other information**:
Cleaned up and resolved conflicts of PR #1492


mauroprojetos:
The only thing I don't understand is where you set up the amount or percentage that will be negotiated.

Ali1:
See sample-config.js

This is passed to portfolioManager near line 59.

greenbigfrog (changes requested):
I think we should stick to one stile for comments and capitalize the first letter if it fits

sample-config.js
```diff
@@ -218,6 +218,7 @@ config.trader = {
   username: '', // your username, only required for specific exchanges.
   passphrase: '', // GDAX, requires a passphrase.
   orderUpdateDelay: 1, // Number of minutes to adjust unfilled order prices
+  currencyAmountAvailableToTrade = false, // set to a number to limit the amount of fiat currency available to Gekko to trade at initiation. Set to false to disable.
```
askmike: > fiat currency

You can run Gekko over non fiat markets, Gekko uses the terminology to distinguish between the two sides of the market: main one is callled "Currency" and the secondary one is called "Asset" (also called base/quote by some people).

askmike:
This is the single most requested feature of Gekko, thanks for the PR.

A few notes on it:

- I love your new added functionality where you track the amount already traded on these trades. However I propose we add it in a way where this code is always used (not just when this feature is enabled), mostly to track bugs and such.
- When you are bailing out because the amount is to low you are returning without calling proper callbacks (such as `checkOrder`).
- you are calling checkOrder after every individual buy/sell, not when they are all done (this is also a bug in the current implementation, quite possibly the bug described in #1852).
- We need to think about implications: this calculates the max amount to buy/sell based on current rates, this means that if you trade USD/BTC and the price goes up and down you might lose more than 100 USD over a timespan. Is this really how we want to implement this?

askmike:
Given this in mind I propose to create something a bit more clear and subtle: a new class that is responsible for doing a single trade (let's call him broker or execution strategy or so), it needs to:

- Figure out if there is a limit on the size of the trade we can do.
- If we can't do any trade (low funds) don't do anything.
- If we can do a trade figure out what kind of trade we aim to do (sell X USD) and trigger an event (see #1850).
- Try to buy sell according to current "limit order do not cross spread" strategy (explained [here](https://gekko.wizb.it/docs/introduction/scope.html#Execution-strategy)).
- only when fully completed figure out trade statistics (average execution rate, amount that was traded, etc).
- only after this upstream a `trade` event with the statistics.

The reason I want to pull this out is so that in the future we can have different execution strategies: the current one tries to get the best price - at the cost of execution speed (since it will always make orders it might take a long time before the trade is completed, especially when trying to catch a trend). Once it's split out we can easily create another one that will do a market order instead. Or one that tries to split the order out into multiple small ones (bigger traders need ways to take pressure of the market for example).

----

@werkkrew @Ali1 what are your thoughts?

sample-config.js
Ali1: thanks, will clarify with next commit

Ali1:
@askmike 

> I love your new added functionality where you track the amount already traded on these trades. However I propose we add it in a way where this code is always used (not just when this feature is enabled), mostly to track bugs and such

Thank you. This code always runs, even without limit trading feature enabled, this is what you will get:
`2018-01-27 02:56:12 (INFO):	BUY order was not (fully) filled, cancelling and creating new order`
`2018-01-27 02:56:12 (INFO):	So far, traded 0.01787247 ETH for approx 15.2437871124 EUR Approx average price: 852.92` **<-- new**
`2018-01-27 02:56:12 (INFO):	Attempting to BUY 0.8334388630927254 ETH for 710.3982837343464 EUR price: 852.37`
`2018-01-27 02:57:13 (INFO):	BUY was successful after 12 tries, trading 0.852421 ETH for approx 735.2437871124 EUR Approx average price: 852.42` **<-- improved**

> When you are bailing out because the amount is to low you are returning without calling proper callbacks (such as checkOrder).

I haven't changed the bail out procedure when the balance is too low (i.e. 0 or lower than exchange minimum). If you mean trading limit is set to 0, then I'm using the same bail out procedure which is for there not to be any callback.

> you are calling checkOrder after every individual buy/sell, not when they are all done (this is also a bug in the current implementation, quite possibly the bug described in #1852).

Sorry not sure about this, I didn't change any callbacks. I didn't actually play a lot with orders that don't fill completely except for adding additional logging. My code was more related to whether the trade should go ahead to begin with or reducing the amount that should be traded to begin with.

> We need to think about implications: this calculates the max amount to buy/sell based on current rates, this means that if you trade USD/BTC and the price goes up and down you might lose more than 100 USD over a timespan. Is this really how we want to implement this?

Again not sure about what you are saying here. Do you mean current rates at initiation (in which case, the use case is to set a maximum USD rate so that multiple bots can share the pot of USD) or current rae at each trade (in which case I think the way I've implemented it makes sense). Let me know.

> Given this in mind I propose to create something a bit more clear and subtle: a new class that is responsible for doing a single trade (let's call him broker or execution strategy or so), it needs to:

This is a fantastic idea. I don't think I'm the right person to do it since I've learnt node for this PR, I've come from different development languages - I'm still trying to get my head around things. I think what I have added seems like a lot, but it's not really - a lot of my code is extra logging which I thought might be useful. Perhaps the next step, through a further PR, can be to package the trade method along with buy and sell methods and callbacks to a new class.

@greenbigfrog 

There's a lot of inconsistency already.

ansonphong:
@askmike 
>Given this in mind I propose to create something a bit more clear and subtle: a new class that is responsible for doing a single trade (let's call him broker or execution strategy or so)

I've been thinking about this for a few weeks, I like the idea of breaking it out into it's own class, that would provide much more flexibility. Brilliant, I added an issue for it here: https://github.com/askmike/gekko/issues/1942

@Ali1 
This is great stuff, I was literally sitting down tonight to code this exact feature in, and I came across this. I hope we can get this merged in soon. 

This is also a good point by @askmike :
> We need to think about implications: this calculates the max amount to buy/sell based on current rates, this means that if you trade USD/BTC and the price goes up and down you might lose more than 100 USD over a timespan. Is this really how we want to implement this?

Another way to look at it, say you had $1000 total portfolio, with only $100 set as the amount you want to trade with on this given strategy, presumably you want $900 to be left untouched. Now Gekko starts trading with BTC, and it makes a trade which loses 20%, now there is $980 remaining in your portfolio. When Gekko initiates a new trade, then it will take out another $100, now you only have $880 remaining. With a bad strategy left over a long period of time this could drain out your entire portfolio.

I quite prefer this approach @Ali1 has implemented over a "keep" approach, as for my own use case I would like to have multiple instances of Gekko running on different BTC pairs from one exchange account. So if I had 1BTC, I could have 10 bots trading all with a maximum of 0.1BTC, on different pairs / strategies.

I suggest a NET value, which is updated after each round trip to include the amount gained/lost. Then you could specify with a boolean option if you want to compound gains. This NET value could optionally be stored in a local DB for consistency across multiple runs, under a key which is the hashed value of the config object, so if you changed the config object at all, or set another boolean inside, it would forget the previous NET value.

``` javascript
{
  currencyAmountAvailableToTrade: 0.1,
  compoundGains: true
}
```

Perhaps we can come up with some more simplified property keys. Thoughts??

BTW @Ali1 In your `sample-config.js` where you define the property `currencyAmountAvailableToTrade = false` should read `currencyAmountAvailableToTrade: false` ;)

aapsoftware:
is there an ETA for merging this PR?

ansonphong:
@aapsoftware This PR has no merge conflicts so you could merge it into your local repository and you should be good to start testing it as-is (I'd suggest with very low amount of funds until you are sure it's working as you would expect). I'm sure any field testing / feedback would be beneficial.

It looks like this is the latest branch:
https://github.com/Ali1/gekko/commits/new-limit-trading

Ali1:
@aapsoftware 
Would be useful for us if you tried it and go ahead and use Ali1/new-limit-trading

@ansonphong 

> Another way to look at it, say you had $1000 total portfolio, with only $100 set as the amount you want to trade with on this given strategy, presumably you want $900 to be left untouched. Now Gekko starts trading with BTC, and it makes a trade which loses 20%, now there is $980 remaining in your portfolio. When Gekko initiates a new trade, then it will take out another $100, now you only have $880 remaining. With a bad strategy left over a long period of time this could drain out your entire portfolio.

Yeah this is the easy way to do it but I haven't gone that way. The behaviour is supposed to work in a way where the most you could lose is the full "asset" (e.g. NANO) value stored in the exchange  plus the amount of currency (e.g. USD) that is set in the config. No matter what happens to the asset, how bad the strategy and how long the bot runs, there's no way that more USD can be lost than the value set in the config (if this works).

> BTW @Ali1 In your sample-config.js where you define the property currencyAmountAvailableToTrade = false should read currencyAmountAvailableToTrade: false ;)

Fixed!

Getting a busier now so have stopped day trading, don't have an appetite to add more features at the moment, sorry. But shouldn't be difficult for more PRs to add extra features.

> Perhaps we can come up with some more simplified property keys. Thoughts??

Suggestions welcome. Also I'm a little new to node. Would "let" in the trade method work for many of the trader class property keys?

Sorry getting quite busy so have stopped day trading for now. Don't have an appetite to add further features at the moment.

Ali1:
Might be useful to be able to alter the currency available while the bot is running, is this something that's possible? Would just need to change the currencyAmountAvailableToTrade property in portfolioManager and it will get picked up at the next trade.

ansonphong:
@Ali1 Yes, `let` is the generally preferred way to scope variables in ES6. So let is for defining scoped variables, not object properties. It is generally a replacement for `var`, which orders definitions linearly, otherwise it is similar. For instance, you would use `let` when defining any variable in any scope, in place of `var`.

I'm going to try and build on what you've done here and progress it to the next stage. Thanks for the PR, it'll take a bit before we can get this merged, although it's such a critical feature right now it's worth taking the time to get it right.

I really like the Trader class concept though, that really got me thinking... Although honestly I don't know if there's anybody better than @askmike to implement something like that, at least take the first swipe at it. @askmike is that anywhere on your upcoming timeline to make a Trade class? A lot of the things that I've been imagining would build on top of that class. If it's far off on your agenda I could prioritize it on mine...

Ali1:
If more work is going to be done, perhaps the same mechanism here can be used to limit the amount of asset traded. Then the "keepasset" feature can be removed. Not that I think people would be using this feature as often.

ansonphong:
@Ali1 I'll have a look at that, the Keep Asset functionality seems considerably simpler so it should be an issue.

@askmike Are you adverse to using ES6 syntax, or do you want to keep everything old school for the time being?

ansonphong:
As I'm digging deeper into this, @askmike's idea about the `Trade` class is seeming more and more inevitable. I'm going to go ahead and do a prototype of the `Trade` class and submit a PR for that after some testing.

aapsoftware:
I took a look at the code; it seems to limit the amount the trader is allowed to trade with not the size of the trade. That is, you can limit your trading amount to 50% of your initial balance (that amount will grow if the trades are profitable) but it does not help if you want to trade a specific amount, always (i.e always place buy orders of 0.1 BTC).
It is helpful, but not quite what I was expecting.
I will give it a go and report back any issues I find.

Ali1:
@aapsoftware Your right that you can't set an option that limits every trade by a fixed amount. Out of interest, why would you want to do that?

aapsoftware:
@Ali1  limiting every trade to a fixed amount would not make much sense; what I was truly hoping for it the possibility to specify the amount I want to trade for each transaction; when you place an order it should be `advice('buy/sell', price, amount)`

On a different note, here's some field testing results:2018-02-22 04:44:43 (INFO):	In uptrend since 
```
2018-02-22 04:49:23 (INFO):	In uptrend since 2 candle(s): @ 0.00330168
2018-02-22 04:49:23 (INFO):	BUY@  0.00330168  ( 0.0033099342 )
2018-02-22 04:49:23 (INFO):	Trader Received advice to go long. Buying  ETC
2018-02-22 04:49:25 (INFO):	Poloniex portfolio:
2018-02-22 04:49:25 (INFO):		 BTC: 0.007276840000
2018-02-22 04:49:25 (INFO):		 ETC: 0.000000000000
2018-02-22 04:49:25 (INFO):		 **Buy orders currently restricted to 0.0035 BTC**
2018-02-22 04:49:25 (INFO):	Attempting to BUY 1.05999576001696 ETC at Poloniex price: 0.00330189
2018-02-22 04:54:23 (INFO):	In uptrend since 3 candle(s): @ 0.00330012
2018-02-22 04:59:23 (INFO):	In uptrend since 4 candle(s): @ 0.00330825
2018-02-22 05:04:24 (INFO):	In uptrend since 5 candle(s): @ 0.00330826
2018-02-22 05:11:04 (INFO):	In uptrend since 6 candle(s): @ 0.00330368
2018-02-22 05:15:24 (INFO):	In uptrend since 7 candle(s): @ 0.00330673
2018-02-22 05:19:04 (INFO):	In uptrend since 8 candle(s): @ 0.00330635
2018-02-22 05:19:27 (INFO):	BUY was successful after the first attempt
2018-02-22 05:19:27 (INFO):	**Buy orders currently restricted to 0 BTC**
2018-02-22 05:19:28 (INFO):	(PROFIT REPORT) original simulated balance:	 0.00727684 BTC
2018-02-22 05:19:28 (INFO):	(PROFIT REPORT) current simulated balance:	 0.00727632 BTC
2018-02-22 05:19:28 (INFO):	(PROFIT REPORT) simulated profit:		 -0.00000052 BTC (-0.00714709%)
```
Just a minor bug with the message **Buy orders currently restricted to 0 BTC**

Ali1:
> advice('buy/sell', price, amount)

@aapsoftware sorry still haven't understood your needs. how would you be able to dynamically specify the amount you want before each transaction.

> Just a minor bug with the message Buy orders currently restricted to 0 BTC

That's expected. After you have used up the 0.0035 BTC that you wanted your bot limited to, then the bot should not be able to execute further buy orders (limit = 0). However if a sell order get executed, then any currency that you earn off that trade becomes tradable again. For example if your bot keeps running and now sells ETC for double the price and you earn 0.007 BTC from the sale, you will get a message saying: ```Buy orders currently restricted to 0.007 BTC```

In this way, the other 0.0037 BTC you have in your exchange balance never gets touched.

aapsoftware:
@Ali1 I would like to have the option to either let the bot trade the full amount available (which is the current behaviour) or specify exactly how much I want to trade - consider this example :
   

-   the limit amount is 0.004BTC
-  advice (buy, 0.002) will  only trade half of the limit amount.

This is the first step towards making it possible to implement more flexible  strategies, where you can:
  

-   place an order at a specific price, before the market reaches that price (i.e. 10% above current market price)
-   place multiple, consecutive buy/sell orders (currently you can only do **buy, sell, buy, sel**l and so on. I would be nice to have the possibility to do **buy, buy, buy, sell, buy, sell, sell** , etc.

Ali1:
Right, now I understand you want the strategy to decide how much to buy. I recommend you create an issue or PR and see what discussion comes out of that.

ansonphong:
@aapsoftware I know exactly what you're talking about, you want to go in with different position sizes which are defined by the strategy, that would be a very good feature.

This PR is not at all about that, what you're wanting would be a layer built on top of this. What this feature is about is limiting the maximum amount which can be traded by the portfolio manager, so that for instance you could have multiple strategies running at the same time on the account which all had a maximum trade limit measured in the base currency.

BTW, @askmike @Ali1 I'm actively working on this feature now including a separate trade class responsible for getting a trade executed, which could be sub-classed to enable different trade execution strategies, will submit a PR after testing.

aapsoftware:
I've been testing this feature for about a week and so far everything looks fine. There is one thing I would change though.  I'd split the `trader.orderUpdateDelay` into an delay for buy and one for sell. If you think about it makes sense - you want to re-enter a buy order fairly quick, while for a sell order you may be willing to hold out longer.

ansonphong:
This is up to @askmike ultimately if this is merged, although we are working on reworking the structure of how trades are handled to allow a more robust framework for trades and orders starting with this PR https://github.com/askmike/gekko/pull/1968. Reguardless, to maintain backward compatibility, I'd suggest we setup the parameter names so they can be consistent with the new developments.

To maintain consistency then I suggest to change the `currencyAmountAvailableToTrade` parameter name to `currencyAllowance`. Thoughts?

aapsoftware:
@ansonphong  `currencyAllowance` is a shorter name and it would make the code easier to read. I think It's a good idea.

askmike:
@aapsoftware @ansonphong @Ali1 I am not comfortable with the amount of if checks which lead to different codepaths based on configuration and or portfolio situation, without any debug statements in here it becomes extremely hard to figure anything out once something goes wrong: we need to know how Gekko was configured and estimate what choices the code made at the time to come to any sort of conclusion, let alone actually fix it.

I want this feature, but the way this is now implement scares me from a maintenance point of view.

aapsoftware:
@askmike valid point (the debug statements). The code as is could use some more logging for when things don't go as expected.
About the `if`s perhaps it would be better if the data paths would be decided at initialisation; you could have multiple methods and decide which one to use an initialisation, depending on the config options. It would also be a bit more efficient, but the code does require a bit of rework (which is well worth doing) to avoid repetition.
What do you think?

askmike:
> About the ifs perhaps it would be better if the data paths would be decided at initialisation; you could have multiple methods and decide which one to use an initialisation, depending on the config options.

Yes this a 100%: calculate all values in the beginning and log them (in debug or maybe even more). After that the codepath can stay the same for all everyone and there are no situations where the code goes down  a different path.

ansonphong:
@askmike Yes having spent several hours studying this PR, I understand where you're coming from regarding maintenance.

I'm currently testing this PR on 4 live gekkos and it's running as expected, it had no merge conflicts with the develop branch, so should be easy to get working for any early adopters who want to use it in the meantime. @Ali1 thanks for putting this together.

weeder:
I just tried your code, and after few trades it forget the config and start to trade with whole portfolio.
https://pastebin.com/dN34KMRF

Ali1:
Hi @weeder 

Looks like the problem happens at 2018-03-08 21:15:38. A buy order is partially filled before new trade advice to sell is received. Gekko forgets to take into account the partially filled order before going on to try and sell. This is because calculations about trade allowance are only made after the order fills completely should be expected to happen. But of course there will be times when it does not so perhaps you or someone could provide a fix for these cases since I am no longer able to be involved at the moment.

It does look like your strategy makes a lot of buy and sell decisions quickly and so really, these orders shouldnt be taking so long to fill. You can look at your configuration and also I think another issue/PR is looking into adding more options to allow orders to get filled faster.

plugins/trader/portfolioManager.js
```diff
@@ -99,6 +110,14 @@ Manager.prototype.setPortfolio = function(callback) {
 
     this.portfolio = portfolio;
 
+    if(this.currencyAmountAvailableToTrade !== false && this.getBalance(this.currency) < this.currencyAmountAvailableToTrade) {
+      // reduce currencyAmountAvailableToTrade to equal the current balance if the current balance is lower than the configured currencyAmountAvailableToTrade
+      if(this.getBalance(this.currency) < 0)
+        this.currencyAmountAvailableToTrade = 0;
```
ansonphong, line 116: @Ali1 I've been running this PR live, and it makes a successful first buy, though it's unable to finish a second trade if one of the orders wasn't filled, it seems as though the `currencyAmountAvailableToTrade ` is being set to 0 here and never changed back later. Is this the intended behavior?

ansonphong:
@Ali1 I've been running this PR live, and it makes a successful first buy, though it's unable to finish a second trade if one of the orders wasn't filled, it seems as though the `currencyAmountAvailableToTrade` is being set to `0` here and never changed back later. Is this the intended behavior?

See `plugins/trader/portfolioManager.js` line `116`

ghost:
It will be great to add something to this feature like for exemple say max amount by trade is 50 USD.
If a buy occure it buy 50 USD then if the rate down and an other long advice happen it buy again 50 USD ...
I want develop a strategy which buy limited amount but not just one time.
I don't know if it possible ...

pabloaugusto:
I trade buying many orders as long as price go up, this minimize the risk of teh operation. A way to create this kind of settup in geckko will be great!

Istead put 1 order at 9300 price for buy 1 BTC:

I buy 0,1 at 9300, then is pice go up
I buy 0,1 at 9325, then is pice go up
I buy 0,1 at 9425, then is pice go up
...

If price drops a certain % i start sell previusly buyed BTC with some rules...

If theres a way to implement it directly from strategy file it  will be asome. I really dont need profit report, i have ir on exchange

askmike:
> I trade buying many orders as long as price go up, this minimize the risk of teh operation. A way to create this kind of settup in geckko will be great!

Creating the orders is easy, reporting this in a simple way in the UI is a lot harder. Unforutnately this is out of scope for now.

--

ALL: This PR was perfect to get the ball rolling, I'm working on a massive upgrade for the execution logic that will include all func described in this PR. See #2118. Let's keep all still alive discussion either in there OR on the forum.

PR #1871, opened 2018-02-06
Return newly created indicator instance in addIndicator method

ansonphong:
Update / modification

Return the instance of the indicator when creating it.

* **Other information**:
This helps to store a reference of the indicator for use in more advanced strategies, and is fully backward compatible.

### DIFF
--- a/plugins/tradingAdvisor/baseTradingMethod.js
+++ b/plugins/tradingAdvisor/baseTradingMethod.js
@@ -224,11 +224,11 @@ Base.prototype.propogateTick = function(candle) {
   // whether candle start time is > startTime
   var isPremature;
 
-  // Subtract number of minutes in current candle for instant start
-  let startTimeMinusCandleSize = startTime.clone();
-  startTimeMinusCandleSize.subtract(this.tradingAdvisor.candleSize, "minutes"); 
-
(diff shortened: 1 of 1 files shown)


askmike:
I don't see the problem of adding this, however I fail to see what it allows for that you can't currently do.

ansonphong:
Right, reason for this PR being, I employ substrategies which are selected by the main Strat, and they create their own indicators, so for initializing them and storing internal refs, it is cleaner and simpler if it returns the indicator on creation.

Also in all cases I like to store a ref to the indicator object on creation, so I don't need to always dig it out of subobjects.

askmike:
sounds great! :+1:

I really do want to do a major refactor of the indicator logic, it's not really reusable and very testable code at the moment. This won't happen before the end of the month though.

ansonphong:
Yes, a full (backwards compatible) refactor of the indicators would be great. Along with that I would love to see a way to visualize indicator graphs and histograms in backtesting, it would help so much for strategy development.

askmike:
> Along with that I would love to see a way to visualize indicator graphs and histograms in backtesting, it would help so much for strategy development.

Will be part of #1850.

PR #1882, opened 2018-02-08
Fix to paper trader when a sell is the first trade and multiple short…

werkkrew:
… advices emit from the strategy before a long

Bug Fix

When backtesting or using the paper trader, if a strategy emits a short advice first, and it emits another short advice before it ever emits a long, the round trips in performanceAnalzer get messed up.

This change makes the paperTrader behave the same way as the actual trader in that it will ignore duplicate instances of the same advice.  The code added to paperTrader is nearly identical to what is in baseTradingMethod so it should not adversely affect anything.

* **Other information**:


askmike:
This shouldn't happen since a short shouldn't be broadcasted twice in a row, see here:

https://github.com/askmike/gekko/blob/d953a37f7ebe04262b7c2d311fa2dbcf536097e3/plugins/tradingAdvisor/baseTradingMethod.js#L319-L321

PR #1898, opened 2018-02-10
Bitfinex - fix providesHistory property

cmroche:
The value wasn't set, data stitching won't work for this exchange without this being set to 'date'.


askmike:
thanks!

PR #1899, opened 2018-02-10
Bitfinex - fix cancelOrder

Paulovms:
Bug fix

If an order wasn't fulfillment when using Bitfinex it is not canceled.

The order is canceled and Gekko makes a new one.

* **Other information**:


cmroche (approved):

askmike:
awesome, thanks!

Issue #1855, opened 2018-02-04
[ERROR] Binance Live Trade: Timestamp for this request was 1000ms ahead of the server's time


ansonphong:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[X] bug report ?

* **Action taken** (what you did)
Run live trader on Binance.

* **Actual result** (unexpected outcome)
Node crashes, with the following error:

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

```
2018-02-03 20:56:12 (INFO):     Attempting to BUY 7.89 NEO at binance price: 0.128
2018-02-03 20:56:12 (DEBUG):    [binance.js] (addOrder) BUY 7.89 NEO @0.128 ETH
2018-02-03 20:56:12 (ERROR):    [binance.js] (addOrder) returned an irrecoverable error: Error: Response code 400
2018-02-03 20:56:12 (DEBUG):    [binance.js] entering "setOrder" callback after api call, err: [binance.js] Response code 400 data: {"code":-1021,"msg":"Timestamp for this request was 1000ms ahead of the server's time."}

Gekko encountered an error and can't continue

Error:
[binance.js] Response code 400

Meta debug info:

Gekko version: v0.5.13
Nodejs version: v8.9.3
```

This is very strange, it's a non-persistent issue, so it's happened about 4-5 times in a row, sometimes after a few hours trading, sometimes right away after starting the trader. I reset the time on my computer and it seems to be working now (maybe, we will see for how long). It's only happening on one of my strategies on 5M candles, not ones running on the same system on 1M candles.

Does anybody have any ideas about this? It would be nice to continue to make Gekko more reliable or be able to deal with this kind of anomalies. Is it related to my internet connection perhaps or my local clock? Not sure what to make of this. Thanks.

askmike:
There have been a ton of issues specifically with Binance in the past. It's pretty stable now (compared to a few weeks back)..

Pinging @cmroche & @aarongarvey

zoeyg:
Are thing setup such that they only utilize `(start|end)TimeSync`?  If you pass a `handleDrift: true` as part of the options to the constructor it will retry the request after calculating the drift.  It will only reattempt a request once though.  If it's still failing after that you must have a very inconsistent network.

TadejPanjtar:
I found this error reproducible on Windows. On Linux works fine.

[...]

PR #1908, opened 2018-02-11
Binance, handling clock drift automatically

cmroche:

Bugfix

Doesn't auto adjust for drift after failed query to Binance

If a request failed due to time, drift is rechecked and handled automatically and the request re-sent.

* **Other information**:

As per @aarongarvey


cmroche:
Fixes #1855

askmike:
:+1:

PR #1920, opened 2018-02-13
Fix issue with trade events being deferred too long

cmroche:

Bugfix

Deferring the advice event causes the back testing to receive advice events in large batches at the end of backtesting.

This is an issue if the strategy listens the to outcome of trades due to performance updates or trade events to feedback into the decision logic. Those events will not happen until far too late in the process. This does not appear to be an issue outside of backtesting.

Removing the deferred causes real-time processing of these events, allow for performance update and trade updates to feed back into the strategies.

* **Other information**:

@askmike I saw that you deferred this event some time ago, but not really sure why. Is this an issue to remove it? Everything looks to work as I would expect without the call to _.defer here.

@werkkrew Your work to bubble trade data into the strategies may be impacted by this, and backtesting results may not accurate be reflecting if you are operating on this data in your strategy.


werkkrew:
I haven't noticed an issue with trade events getting deferred in back testing, but yes, in my current strategy that I have been evolving/developing I rely on the completed trade event coming back into my strategy when it is expected.  So far I have not experienced this bug, or at least, I haven't noticed it.

What I have noticed, and this might be of interest to you regarding portfolio management and even the websocket work, is that Gekko does not react to the same advice twice.  So depending on where the portfolio starts and how you want to manage your portfolio if you strategy emits advice to short more than once in a row before switching to a long advice, only the first instance of each advice makes it into portfolio manager logic.  It will likely be desirable in the future to allow all advice through from the strategy and allow the portfolio manager to decide if it should react to it or not (based on available balance or whatever other logic).

cmroche:
@werkkrew I already fixed that problem in my local code, gekko also has an issue where you cannot cancel and advice and take a neutral action, also fixed. One of us should send a PR for all that :)

Thanks.

askmike:
@cmroche since the advice includes the candle that is used by the performance analyzer this should not be impacted.

Note that #1850 will have some differences in the event flows, so this will be fixed later anyway, but there is no reason not to merge it now :)

PR #1921, opened 2018-02-13
Fix roundtrip calculations

cmroche:

Bugfix

Roundtrip calculations only work if you follow a specific order of sequential events, I think it's buy then sell, ad infinitum. If you start with all your holdings in asset then move to currency you first roundtrip does not compute properly, and if you execute multiple sells/buys consecutively it messes up P&L computations (and shows incorrect durations).

This change records the P&L entry state by adding an ID so we can see when new positions are opened, and when an existing position is being updated by consecutive events, this also gives inherent stability if your first action is a sell (IIRC, it will just not report this case since the information is incomplete).

This change leads way to changes I will prepare and push in the future that allows strategies to specify a percentage of the portfolio to trade, which is desirable as it allows risk management in the strategy advice.

* **Other information**:

@werkkrew Based on other changes you've made I wonder if this will be relevant for you.


werkkrew:
It is, and I have done some work in this area.  

I think outside of just calculating round trips and on the subject or portfolio management: the ability for a position to start as a long or a short, and end on a "closePosition" event is likely just as important as being able to use a portion of the portfolio on each bit of advice.  In looking at it, since most of gekko assumes a round trip always starts with a buy, it appears to me that it might take some doing to overhaul portfolioManager to have a better way to manage actual positions, including true short positions and not just the closing of a long position by selling. 

Also, and not to derail the scope of this PR, the other functionality you mention with respect to percentages of the portfolio should probably also be able to accept fixed amounts and not just percentages.

askmike:
Ah great stuff! I started refactoring the performance Analyzer to make room for all the new events but I haven't included these fixes. Thanks!

PR #1929, opened 2018-02-15
Missing 'args' definition in quadriga

### LINKED ISSUE #1923: Quadriga async.series(getOrders, relay) in porfolioManager crashes
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[ ] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
async.series(getOrders, relay) in porfolioManager Manager.prototype.relayOrder = function(done) { } crashes right after first trade (buy or sell) on quadrigacx.

* **Expected result** (what you hoped would happen)

* **Actual result** (unexpected outcome)

cmroche:

Bugfix

Missing definition of 'args' in getOrder and checkOrder, means that it couldn't fail gracefully.

Fails gracefully

* **Other information**:

May resolve #1923 

@jchudon would you please test these changes to ensure they address your problem. Hitting this error means that the API was about to return an error, but that may be normal in this case since it takes time for the order to propagate into the order books, the function will retry until it succeeds, I would just like to confirm that it does and there are no other problems.


askmike:
Thanks! I am okay with merging this and giving your fix some exposure, quadriga is not an exchange used very often.

Issue #1912, opened 2018-02-12
Error with the size of trade at Bitfinix


DerBeweis:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Traded with CL on Bitfinex. EOS/BTC and NEO/USD
With enough BTC and USD
* **Expected result** (what you hoped would happen)
The Bot will do the trades and if it is not filled 100% it will just continue with what it filled

* **Actual result** (unexpected outcome)
It did a partial trade, and then broke out because the rest was not filled

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
![acd57b0b-d43b-4075-9b29-b3d09b4c5993](https://user-images.githubusercontent.com/34221585/36087965-624c43b2-0fd5-11e8-9aee-7f695dfc1653.png)

DerBeweis:
@askmike 
Should I just try the development branch? If so
Would that be correct : 
$ git pull
$ npm install --Development 

To update my version to the development one?
Thanks

DerBeweis:
I got it now, running with the development branch. 
If you think it is solved in this branch you can close the bug

DerBeweis:
I still get this: please help 
![4c1d18b0-2168-4235-871d-b40143709f4c](https://user-images.githubusercontent.com/34221585/36111338-16e4bc14-1026-11e8-94ac-d58af1eaf824.jpeg)

Issue #1469, opened 2017-12-13
Bitfinex: BUY order was not (fully) filled - ordering at wrong amount


magicdude4eva:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
Auto-trade. Buy was not fulfilled

* **Expected result** (what you hoped would happen)
New order should have had correct quantity.

* **Actual result** (unexpected outcome)

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
```
2017-12-13 19:41:35 (INFO):	We have new trading advice!
2017-12-13 19:41:35 (INFO):		 Position: long
2017-12-13 19:41:35 (INFO):		 Market price: 0.41339
2017-12-13 19:41:35 (INFO):		 Based on market time: 2017-12-13 17:40:00

2017-12-13 19:41:35 (INFO):	Trader Received advice to go long. Buying  XRP
2017-12-13 19:41:36 (INFO):	Attempting to BUY 764.6142323602439 XRP at Bitfinex price: 0.4145

2017-12-13 19:42:38 (INFO):	BUY order was not (fully) filled, cancelling and creating new order
2017-12-13 19:42:40 (INFO):	Attempting to BUY 0.5937004705882354 XRP at Bitfinex price: 0.4249
2017-12-13 19:42:41 (ERROR):	unable to buy Error: Invalid order: minimum size for XRP/USD is 52.0
    at Request.request [as _callback] (/Users/gekko/node_modules/bitfinex-api-node/rest.js:61:17)
    at Request.self.callback (/Users/gekko/node_modules/request/request.js:188:22)
    at emitTwo (events.js:126:13)
    at Request.emit (events.js:214:7)
    at Request.<anonymous> (/Users/gekko/node_modules/request/request.js:1171:10)
    at emitOne (events.js:116:13)
    at Request.emit (events.js:211:7)
    at IncomingMessage.<anonymous> (/Users/gekko/node_modules/request/request.js:1091:12)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1056:12)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
[...]

PR #1944, opened 2018-02-18
Bitfinex - Fixes wrong values after an order cancellation 

Paulovms:
Bug fix.

#1912 #1469
Gekko sometimes gets a wrong balance from Bitfinex servers after an order has not been fulfilled.

Gekko waits 10 seconds before makes a new order and Bitfinex servers can completed the calculations after the order cancellation and give the correct balance.

* **Other information**:
There are an issue in Bitfinex-api-node repository about this behavior.
bitfinexcom/bitfinex-api-node#141
This is a server issues (not in bitfinex-api-node or in Gekko), and has to do with the delay between closing an order and the associated balance update. 
It has to do with the fact that balance calculations are queued and not instantaneous.


askmike:
:+1: thanks!

Issue #1592, opened 2017-12-31
Child process has died when I  scan for Market Watcher Statistics ,How??


Jackzhangpan:
```
E:\Gekko\gekko\node_modules\bitcoin-co-id\index.js:65
    requestOptions[x] = options[x];
                   ^

ReferenceError: x is not defined
    at Bitcoincoid._request (E:\Gekko\gekko\node_modules\bitcoin-co-id\index.js:65:20)
    at Bitcoincoid._get (E:\Gekko\gekko\node_modules\bitcoin-co-id\index.js:20:8)
    at Bitcoincoid.getTrades (E:\Gekko\gekko\node_modules\bitcoin-co-id\index.js:119:8)
    at Trader.getTrades (E:\Gekko\gekko\exchanges\bitcoin-co-id.js:113:20)
    at Trader.bound [as getTrades] (E:\Gekko\gekko\node_modules\lodash\dist\lodash.js:729:21)
    at Fetcher._fetch (E:\Gekko\gekko\core\budfox\marketFetcher.js:75:16)
    at Fetcher.bound [as _fetch] (E:\Gekko\gekko\node_modules\lodash\dist\lodash.js:729:21)
    at Fetcher.fetch (E:\Gekko\gekko\core\budfox\marketFetcher.js:88:8)
    at Fetcher.bound (E:\Gekko\gekko\node_modules\lodash\dist\lodash.js:729:21)
    at Manager.retrieve (E:\Gekko\gekko\core\budfox\marketDataProvider.js:31:15)
RECEIVED ERROR IN GEKKO 535418895635132
Child process has died.
```

Jackzhangpan:
@cloggy45 can help me? thank you

cloggy45:
I will look into it.

cloggy45:
From what I can tell, it relates to this [node package issue](https://github.com/danKHUU/bitcoin-co-id/commit/7426836e0bd805ebe78f7808881e9e19eebc1fad).

cloggy45:
@Jackzhangpan To solve the problem use ``npm install --only=production`` this will get all the updated dependencies.

Jackzhangpan:
OK，thank you

Hanna96:
@cloggy45 it doesn't work for me

masbenx:
@cloggy45 doesn't work for me too
I still get error

> ReferenceError: x is not defined
[...]

PR #1945, opened 2018-02-19
update package for bitcoin-id

cloggy45:
* **What kind of change does this PR introduce?** (Bug fix, feature, docs 

Fetches the update bitcoin-id package using the github repo, since the NPM package is not updated to include the fix. 

Issue when using bitcoin-id
https://github.com/askmike/gekko/issues/1592#issuecomment-354579581

Resolves the issue mentioned above

* **Other information**:


askmike:
This only works when people have git installed. Else npm won't be able to install the package (AFAIK). A better way would be to fork the project and publish a new package that includes the fix, and point Gekko to that one.

cloggy45:
@askmike I was not aware of that. I will do what you advised.

askmike:
:+1: thanks

PR #1948, opened 2018-02-19
Quadriga Quick Bugfix

cmroche:
Just adding toLowerCase at two more lines that were initially missed.


askmike:
Thanks as always :)

TheWarlu:
Does this mean we will get Quadriga import soon?

askmike:
unfortunately not :(

cmroche:
@TheWarlu Quadriga does not support historical data import through their API.

Issue #1942, opened 2018-02-18
A new class that is responsible for doing a single trade...


ansonphong:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[ ] bug report
[X] feature request
[ ] question about the decisions made in the repository

So this is an idea that @askmike came up with in this PR: https://github.com/askmike/gekko/pull/1834. I would like to surface it here so that we don't loose track of it, as it deserves it's own issue.

>I propose to create something a bit more clear and subtle: a new class that is responsible for doing a single trade (let's call him broker or execution strategy or so), it needs to:
> - Figure out if there is a limit on the size of the trade we can do.
> - If we can't do any trade (low funds) don't do anything.
> - If we can do a trade figure out what kind of trade we aim to do (sell X USD) and trigger an event (see #1850).
> - Try to buy sell according to current "limit order do not cross spread" strategy (explained here).
> - only when fully completed figure out trade statistics (average execution rate, amount that was traded, etc).
> - only after this upstream a trade event with the statistics.

>The reason I want to pull this out is so that in the future we can have different execution strategies: the current one tries to get the best price - at the cost of execution speed (since it will always make orders it might take a long time before the trade is completed, especially when trying to catch a trend). Once it's split out we can easily create another one that will do a market order instead. Or one that tries to split the order out into multiple small ones (bigger traders need ways to take pressure of the market for example).

Ali1:
Would be good if the trade processing code that is executed is separated from the attempt order code which may repeat a few times until the trade gets fully filled. Currently the trade method keeps executing itself if it's only partially filled.

There may be better terminology options for the full trade and each trade attempt that is made until fully filled:

- Trade and Attempt
- Trade and Order
- Execute Advice and Trade/Order/Trade Order
- Trade Action and Trade/Order/Trade Order
- Execute Trade and Trade Order/Order

The class name and method name should reflect the chosen terminology for clarity and consistency.

askmike:
Great point.

> Currently the trade method keeps executing itself if it's only partially filled.

This wasn't a problem before your PR since the whole balance was supposed to get swapped, but this is definitely limiting us right now in a lot of ways.

> There may be better terminology options for the full trade and each trade attempt that is made until fully filled

Could you elaborate on what you mean with those names? I'm not following them.

Ali1:
[...]

PR #1968, opened 2018-02-24
Introduce Trade Class

ansonphong:
A new class that is responsible for doing a single trade.
Related to discussion in Issue https://github.com/askmike/gekko/issues/1942

Currently all the trade and order functionality is mashed into the portfolio manager, making it  impossible to deploy different trade and order execution strategies, as everything is hard-coded to one specific type of trade execution which is a single limit order for the full balance amount right across from the spread.

This PR strictly breaks out the trade specific functionality (as-is) into a separate Trade class, with minimal modification and no refactoring. 

* **Other information**:
This is an initial proposal for the Trade class. I have tested it on the live market and it works as expected without any errors so far. I'm curious to see others kick the tires and try to shoot holes in it and see how it works in different circumstances, and see if I missed anything.

This PR is far from ideal as a final implementation, and is part of an incremental series of PRs which will see a total refactor of the `Portfolio Manager`, this `Trade` class, and the introduction of a more robust `Order` model. Refer to the Issue https://github.com/askmike/gekko/issues/1942 for further discussion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,15 +94,6 @@
         }
       }
     },
-    "JSONStream": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
-      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
-      "requires": {
-        "jsonparse": "1.3.1",
-        "through": "2.3.8"
-      }
-    },
     "accepts": {
       "version": "1.3.4",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
@@ -2090,6 +2081,15 @@
       "resolved": "https://registry.npmjs.org/jsonpointer/-/jsonpointer-4.0.1.tgz",
       "integrity": "sha1-T9kss04OnbPInIYi7PUfm5eMbLk="
     },
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "jsprim": {
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/jsprim/-/jsprim-1.4.1.tgz",
@@ -2794,9 +2794,9 @@
       "resolved": "https://registry.npmjs.org/node.bittrex.api/-/node.bittrex.api-0.4.4.tgz",
       "integrity": "sha512-zNrwiSufttRBfPeSJfQLRDd9AHQuAL2IVxJEdEtNvwqvqHsdRvPkiQfANOzPy+0jFM/J8/t6/+gJ8Df+0GkgiQ==",
       "requires": {
-        "JSONStream": "1.3.1",
         "event-stream": "3.3.4",
         "jsonic": "0.3.0",
+        "JSONStream": "1.3.1",
         "request": "2.83.0",
         "signalr-client": "0.0.17"
       }
@@ -4424,6 +4424,14 @@
             }
           }
         },
+        "string_decoder": {
+          "version": "1.0.3",
+          "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.0.3.tgz",
+          "integrity": "sha512-4AH6Z5fzNNBcH+6XDMfA/BTt87skxqJlO0lAh3Dker5zThcAxG6mKz+iGu308UKoPPQ8Dcqx/4JhujzltRa+hQ==",
+          "requires": {
+            "safe-buffer": "5.1.1"
+          }
+        },
         "string-width": {
           "version": "1.0.2",
           "bundled": true,
@@ -4431,14 +4439,6 @@
             "code-point-at": "1.1.0",
             "is-fullwidth-code-point": "1.0.0",
             "strip-ansi": "3.0.1"
-          }
-        },
-        "string_decoder": {
-          "version": "1.0.3",
-          "resolved": "https://registry.npmjs.org/string_decoder/-/string_decoder-1.0.3.tgz",
-          "integrity": "sha512-4AH6Z5fzNNBcH+6XDMfA/BTt87skxqJlO0lAh3Dker5zThcAxG6mKz+iGu308UKoPPQ8Dcqx/4JhujzltRa+hQ==",
-          "requires": {
-            "safe-buffer": "5.1.1"
           }
         },
         "stringstream": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,6 +94,15 @@
         }
       }
     },
+    "JSONStream": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
+      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
+      "requires": {
+        "jsonparse": "1.3.1",
+        "through": "2.3.8"
+      }
+    },
     "accepts": {
       "version": "1.3.4",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
@@ -2081,15 +2090,6 @@
       "resolved": "https://registry.npmjs.org/jsonpointer/-/jsonpointer-4.0.1.tgz",
       "integrity": "sha1-T9kss04OnbPInIYi7PUfm5eMbLk="
     },
-    "JSONStream": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.1.tgz",
-      "integrity": "sha1-cH92HgHa6eFvG8+TcDt4xwlmV5o=",
-      "requires": {
-        "jsonparse": "1.3.1",
-        "through": "2.3.8"
-      }
-    },
     "jsprim": {
       "version": "1.4.1",
       "resolved": "https://registry.npmjs.org/jsprim/-/jsprim-1.4.1.tgz",
@@ -2794,9 +2794,9 @@
       "resolved": "https://registry.npmjs.org/node.bittrex.api/-/node.bittrex.api-0.4.4.tgz",
       "integrity": "sha512-zNrwiSufttRBfPeSJfQLRDd9AHQuAL2IVxJEdEtNvwqvqHsdRvPkiQfANOzPy+0jFM/J8/t6/+gJ8Df+0GkgiQ==",
       "requires": {
+        "JSONStream": "1.3.1",
         "event-stream": "3.3.4",
         "jsonic": "0.3.0",
-        "JSONStream": "1.3.1",
         "request": "2.83.0",
         "signalr-client": "0.0.17"
       }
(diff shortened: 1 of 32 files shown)


package-lock.json
```diff
@@ -94,6 +94,15 @@
         }
       }
     },
+    "JSONStream": {
```
askmike: do we need these changes?

plugins/trader/trade.js
```diff
@@ -0,0 +1,293 @@
+/*
+  @askmike
```
askmike: better to link to the discussion than copy snippets of it in a comment :)

plugins/trader/trade.js
```diff
@@ -0,0 +1,293 @@
...
+      this.manager.setFee
+    ], _.bind(act, this));
+
+  };
+
+  // first do a quick check to see whether we can buy
+  // the asset, if so BUY and keep track of the order
+  // (amount is in asset quantity)
+  buy(amount, price) {
+    let minimum = 0;
```
askmike, line 105: pretty sure we had a minimum, right?

plugins/trader/portfolioManager.js
```diff
@@ -137,195 +139,50 @@ Manager.prototype.setTicker = function(callback) {
...
+    this.tradeHistory.push(this.currentTrade)
+    this.currentTrade = false
   }
-
-  this.orders.push(order);
-
-  // If unfilled, cancel and replace order with adjusted price
-  let cancelDelay = this.conf.orderUpdateDelay || 1;
-  setTimeout(this.checkOrder, util.minToMs(cancelDelay));
+  return this.currentTrade = new Trade(this,{action: what})
```
askmike: If we want to stick to OO: can you not pass the parent (portfolioManager) into the child (trade), these circular references are tightly coupled.

askmike (commented):
See few comments, also: does the ES6 class syntax work nodev6? If it doesn't there is a node update process we need to go through.

package-lock.json
ansonphong: No, any changes to the `package-lock.json` are based on local package preferences, and aren't required for this PR, I'll remove that.

plugins/trader/trade.js
ansonphong: Agreed, will update.

plugins/trader/portfolioManager.js
ansonphong: Yes, it's not ideal, let's fix this.

Currently there are multiple references to the manager inside the trade class, which would call for some restructuring. This is what we could do:

- A new exchange object could be created inside the Trade class so it's not referencing it in the Manager, thus removing references to `manager.exchange`
- We can pass in the current pair currency/asset handles into the Trade class on creation, and remove `manager.asset` and `manager.currency`
- Currently the following methods are being called by the Trade class, how do you suggest we handle this?
```
this.manager.setTicker
this.manager.setPortfolio
this.manager.setFee
this.manager.getMinimum
this.manager.convertPortfolio
```

ansonphong:
Yes, Classes should work since Node `V6.5`, it also seems to support 97% of all ECMA 6 features listed:

http://kangax.github.io/compat-table/es6/

plugins/trader/trade.js
ansonphong: Yes, that code is just directly pasted from the previous implementation, looks like the minimum is defined later on line `133` before the callback is executed.

aapsoftware:
what's the current status with this branch? is it usable or are there still changes to be made?

askmike:
> Yes, Classes should work since Node V6.5, it also seems to support 97% of all ECMA 6 features listed:

Gekko should work for Gekko v6.0, we can upgrade the minimum but as explained above this is a different process. I do want to move to use classes more so I'll look into any possible negative effects of requiring a higher version of nodejs.

plugins/trader/portfolioManager.js
askmike: > this.manager.convertPortfolio

This is currently only needed for emitting events upstream, the trade class doesn't need it.

Some other functions won't be needed anymore by the portfolioManager, like setFee and setTicker (we will need this one later again, for the [portfolioValueChange](https://github.com/askmike/gekko/blob/0d975707009456720f59b98c102c334c05e35585/docs/internals/events.md#portfoliovaluechange-event) event for example). All the other functions will be used differently anyway, so I propose the portfolioManager passes the exchange (`this.exchange`) to the tradeclass which reimplements what it needs, we can also drop in the portfolioManager what we don't need. What do you think?

ansonphong:
@askmike Yes, I'm open to go through a Node upgrade process whatever is required, I think it's important to be progressive with adopting new standards.

plugins/trader/portfolioManager.js
ansonphong: @askmike OK that makes sense, yes let's keep a single Exchange instance, and move the methods into the Trade class. I'll make that happen, test it and update this PR.

ansonphong:
@askmike I made the modifications that we discussed earlier in the code comments.

Since the "Portfolio" object and associated methods needed to be accessed by both the Trade class and the Portfolio Manager class, I broke it out into a "Portfolio" class / object which has a shared reference between both classes, which contains the portfolio data and ticker.

I tested this extensively in high frequency / high tick rate and high position reversal conditions on the live market and it works as expected.

This is not ready to merge yet until the logging and syntax is cleaned up, I'd like to make it more consistent. I'm curious to get your feedback on the last commit.

After this, the next thing I'd like to develop on this is a proper data model for `Orders`, which will help with having a perfect calculation of the status of a trade, and precisely measuring averages, slippage, etc, which will be needed for limiting position sizes, currency allowance, compounding gains, etc. Although that will require another pull request which may require modification to the exchange classes to collect additional data from orders such as how much was actually filled, order status, etc. We can discuss on another issue.

Let me know your thoughts on this last commit, it's ready to be live tested.

astralmedia:
I am looking forward to this! Thanks everyone involved.

ansonphong:
@askmike Also what is required to conduct a Node 6.5+ upgrade so we can get ES6 support? While we are at it why not update to Node 8? I've been running Node 8 on Windows and Linux without any issues thus far.

aapsoftware:
@ansonphong  node8 is diffcult to install on some raspberry pi versions. So you loose a little flexibility if you go to node 8 directly. Plus node 8 does not bring that much to the game to be worth making it a requirement. 
Even with node 6 you have to be careful choosing your version; my raspberry pi runs node 6.2.1 and it the latest  version you can install using yum

ansonphong:
@aapsoftware How about Node V6.5, that supports ES6, is that manageable on Raspi?

plugins/trader/portfolioManager.js
```diff
@@ -17,381 +15,95 @@ var log = require(dirs.core + 'log');
...
-  var getOrders = _.map(
-    this.orders,
-    order => next => this.exchange.getOrder(order, next)
-  );
+  return this.currentTrade = new Trade({
+    action: what,
+    exchange:this.exchange,
+    currency: this.conf.currency,
+    asset: this.conf.asset,
+    portfolio: this.portfolio,
```
ansonphong, line 91: I do not like all of these property values passed in, ideally it's just `action`, `currency`, `asset`, and then an `amount` which could have an undefined value which would then default to the system settings for currency allowance, keep asset, etc.

@askmike You know the architecture best, is there an currently established way by which we could store references to globally accessible objects / dependencies without resorting to a singleton?

ansonphong:
@askmike Can you have a look at this when you have some time? I'd like to make a decisive move on this PR, as a lot of future work I'd like to do builds on this. Thanks!

askmike:
Apologies for the delay. I am testing this now, If I don't find problems I
will merge it today.

positonic:
@ansonphong I'm interested in moving this forward and have availability, please let me know how to work with you in getting there faster...

@askmike I'm also eagerly awaiting this :)

ansonphong:
Great thanks, @jamespfarrell you are welcome to test this PR, I'm only 99% confident that it works as expected, having others help testing is always helpful.

@askmike Check my comments above, I can do a final pass at it before merging.

minhnguyenvan95:
@ansonphong interesting

ansonphong:
I made another push which basically just cleans up some of the debugging language and the comments a little bit, it should be ready to merge. I'm not content with the solution for passing the `Portfolio` object into the Trade, although it does work and makes a shared Portfolio object. Having a single object for the Portfolio itself is nice, although having a better way to access it via the core architecture would be ideal (IMHO), rather than having to pass it around.

That said, this PR simply breaks out the existing functionality into separate classes, and does not introduce any new functionality. The next work will require refactoring the methods to allow for an architecture which will allow for sub-classes of the Trade class which can carry out different trade strategies. For instance, alternate strategies could include layering multiple orders, placing market orders, etc, and pave the way for more sophisticated systems in Gekko which could allow for position sizing, etc. The refactoring can happen on another PR.

The very next step I'd like to take with it is to build an `Order` model which tracks all of the orders and their status, within a given Trade. This will require minimal updating of the Exchange class, which we can do on another PR. That will allow for a limited currency allowance to be traded with and then a ratio could be defined on how much profits to keep and how much to continue future trades to compound gains. This would allow Gekko users to basically run many instances of Gekko trading on the same exchange account on different pairs.

positonic:
> That said, this PR simply breaks out the existing functionality into separate classes, and does not introduce any new functionality. The next work will require refactoring the methods to allow for an architecture which will allow for sub-classes of the Trade class which can carry out different trade strategies. For instance, alternate strategies could include layering multiple orders, placing market orders, etc, and pave the way for more sophisticated systems in Gekko which could allow for position sizing, etc. The refactoring can happen on another PR.

If I understand you correctly, with this refactoring a Portfolio refers to an exchange portfolio, rather and an entire portfolio across many exchanges? 

Also, if I understand this approach correctly you we will be able to apply many strategies on the market, but since this is operating through plugins on the GekkoStream, this will still only be able to act on 1 market at a time?

@ansonphong @askmike I would love to test the PR but can't get a copy of it besides making the change manually. I have done a lot of googling and tried a lot of commands, but can't seem to get a copy... Can you help? I'm super keen to test and further this issue...

I have tried:

git fetch origin pull/1968/ansonphong:trade-class
git fetch upstream pull/1968/head:trade-class      

[I have also run through this with no joyl](https://gist.github.com/piscisaureus/3342247).

and few other things...

@ansonphong are you working on a fork?

ansonphong:
@jamespfarrell You can easily pull off my fork to test:
https://github.com/ansonphong/gekko/tree/trade-class

In your local developent Gekko, just go 👍 
`git pull git@github.com:ansonphong/gekko.git trade-class`

Still will only be on one market at a time, though you will be able to run multiple instances.

askmike:
This is long overdue, again apologies for the delay. This PR is perfect in that it only moves around code but doesn't change behaviour / introduces new concepts. Love to work on upgrades and will reply to your latest comment @ansonphong asap!

askmike:
Before this can land in stable we do need to upgrade the minimal node version (as stated before), but as part of #1850 this will be necessary anyway. I will focus on that.

ansonphong:
@askmike Sounds good, happy to get this merged. Are you open to eliminating all semicolons on line endings for ES6 code?

zawsx:
>>**eliminating all semicolons** 
Google say semicolons required - they are not the authority of course, but many people will follow their [JavaScript Style Guide](https://google.github.io/styleguide/jsguide.html#formatting-semicolons-are-required). The tendency is clear, lots of code I see has none, they're not needed. 
I can see that the occasional horrible bug would be avoided if you litter up the code with superfluous symbols but not here, surely?

askmike:
> @askmike Are you open to eliminating all semicolons on line endings for ES6 code?

What do semicolons have to do with ES6? Did anything change? As far as I know they were optional and they still are optional. Most styleguides I know recommend using them though.

PR #1970, opened 2018-02-25
Prevent dataset scanning from depleting memory

cmroche:

Bugfix

If you have large datasets, with many markets imported the dataset scanner is going to fork for every single exchange and market combination all at once. This depletes memory very fast, and doesn't provide a significant gain in processing performance (can often lead to worse performance even).

AWS and Docker machines often crash due to the scanning requiring several gigs of memory.

The dataset scanner now queues and runs only as many forks as there are CPU cores on the system. Greatly reducing the impact on memory to a couple 10s of MB.

* **Other information**:


askmike:
super slick!

PR #1991, opened 2018-03-04
Incomplete datasets no longer fool the datastitcher

cmroche:
A simple bugfix that recognizes incomplete local data sets when trying to seed trading, the condition now defaults to re-importing the needed history.


askmike:
Ah nice catch!

PR #2012, opened 2018-03-08
Fix Bitstamp precision for BTC pairs and XRPUSD

kevinmershon:
Bug fix

XRP/USD, ETH/BTC, BCH/BTC, XRP/BTC, and LTC/BTC trading pairs only trade with 2 decimals of precision and frequently trades won't complete

* **What is the new behavior?**
All aforementioned trading pairs now use appropriate precision per the exchange

* **Other information**:
None


askmike:
great stuff, thanks!

PR #2018, opened 2018-03-09
Integrated Coingi and added a new feature "exchangeMaxHistoryAge"

jmcklondonuk:

feature

Supported exchanges don't cointain coingi. Data import is only supported where exchanges return their full data.

Coingi is among supported exchanges. Data import is supported also when exchanges return only the last X days (for example the last 30 days)

* **Other information**:
https://forum.gekko.wizb.it/thread-1447.html


askmike:
Great! Thanks a lot!

PR #2087, opened 2018-04-05
Add OMG, ZRX, CVC, GAS assets to BTC and ETC currencies

xFFFFF:
should be script to auto updates like for binance

pairs update

havent gas, cvc, omg,zrx pairs

have pairs :D

* **Other information**:
havent, but I added and now have :)


VeeeneX:
It would be better if it could be automated

xFFFFF:
Its true. But I haven't skills to write script in javascript to auto update.

GNO is missed too but only in array: assets. How I can edit file from this PR?

askmike:
thanks!

askmike:
@VeeeneX yes it would be, this is automated for a number of exchanges such as bitfinex and binance. Very open to PRs to extend this to polo :)

PR #2095, opened 2018-04-09
fixing candle writer

nichdiekuh:

It fixes the candle-writer for livebots.

When a livebot is startet (via cli), it fetches the local dataset and tries to stich it with the exchange data. The reader ends the DB connection once it's done reading and the candle writer fails silently, since the db connection is already closed by the reader.

The connetion stays open


askmike:
Are you using postgresql?

nichdiekuh:
I used to run gekko with postgres, but switched to mysql a few weeks ago.

Issue #2106, opened 2018-04-15
Gdax Trade Class Error


Kris191:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a Gekko CLI crash error...**
[x ] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Rn Gekko bot on GDax exchange after pulling latest update including the Trade class PR, when the bot attempts to trade on advice it crashes.

**Expected result** (what you hoped would happen)
It would trade

**Actual result** (unexpected outcome)
It crashes

**Other information** (e.g. Detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

CLI error code below...

/root/gekko/plugins/trader/trade.js:312
      return minimum = this.minimalOrder.amount;
                     ^

ReferenceError: minimum is not defined
    at Trade.getMinimum (/root/gekko/plugins/trader/trade.js:312:22)
    at Trade.sell (/root/gekko/plugins/trader/trade.js:185:22)
    at act (/root/gekko/plugins/trader/trade.js:89:18)
    at /root/gekko/node_modules/async/dist/async.js:3853:9
    at /root/gekko/node_modules/async/dist/async.js:484:16
    at replenish (/root/gekko/node_modules/async/dist/async.js:1025:25)
    at iterateeCallback (/root/gekko/node_modules/async/dist/async.js:1015:17)
    at /root/gekko/node_modules/async/dist/async.js:988:16
    at /root/gekko/node_modules/async/dist/async.js:3850:13
    at apply (/root/gekko/node_modules/async/dist/async.js:41:25)
    at /root/gekko/node_modules/async/dist/async.js:76:12
    at set (/root/gekko/plugins/trader/portfolio.js:89:9)
    at Trader.getFee (/root/gekko/exchanges/gdax.js:140:3)
    at Trader.bound [as getFee] (/root/gekko/node_modules/lodash/dist/lodash.js:729:21)
    at Portfolio.setFee (/root/gekko/plugins/trader/portfolio.js:91:19)
    at /root/gekko/node_modules/async/dist/async.js:3845:9
    at replenish (/root/gekko/node_modules/async/dist/async.js:1030:17)
    at iterateeCallback (/root/gekko/node_modules/async/dist/async.js:1015:17)
[...]

Issue #2083, opened 2018-04-04
Reference Error -> minimum is undefined (trader crashes)


H256:
**I'm submitting a ...**
[x ] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
From command-line
Start live trading @bitstamp using current develop version (incl. new Trade-Class)

**Expected result** (what you hoped would happen)
It trades...

**Actual result** (unexpected outcome)
Errormessage and process crashes.
```
C:\gekko\plugins\trader\trade.js:310
      return minimum = this.minimalOrder.amount / price;
                     ^

ReferenceError: minimum is not defined
    at Trade.getMinimum (C:\gekko\plugins\trader\trade.js:310:22)
    at Trade.buy (C:\gekko\plugins\trader\trade.js:140:22)
    at act (C:\gekko\plugins\trader\trade.js:83:18)
    at C:\gekko\node_modules\async\dist\async.js:3853:9
    at C:\gekko\node_modules\async\dist\async.js:484:16
    at replenish (C:\gekko\node_modules\async\dist\async.js:1025:25)
    at iterateeCallback (C:\gekko\node_modules\async\dist\async.js:1015:17)
    at C:\gekko\node_modules\async\dist\async.js:988:16
    at C:\gekko\node_modules\async\dist\async.js:3850:13
    at apply (C:\gekko\node_modules\async\dist\async.js:41:25)
```

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
This seems to happen on the first advice, the trader gets (regardless of being a short/long).
It happened only at BitStamp so far, I have other instances running where no crash occured.

I have also tried it with a fresh clone of the repo - the result is the same.

askmike:
ah nice catch! ping @ansonphong

dmuzizian:
Error. How to FIX? i'm using BITTREX ... BTC-DOGE
![screenshot_1](https://user-images.githubusercontent.com/15715474/38730717-d7492662-3f41-11e8-9887-ee9d1e88b197.png)

H256:
@dmuzizian I have just temporarily altered the line to this: 
`return this.minimalOrder.amount`
This seems to work.

dmuzizian:
where i can find this line ? @H256

H256:
[...]

PR #2108, opened 2018-04-16
Fix minimum error

H256:
#2106, #2083

Fixing error message due to syntax error in function `getMinimum()`

An error is thrown.

The minimum-amount is returned correctly... ;)

* **Other information**:
I finally had time to fork and open this PR :)


xFFFFF:
I really like Your UI, but now i need fix minimum error only gekko on my remote server. Im not using UI there. Which files I should change? :)

H256:
Will open a cleaned up one - so that review is easier...

PR #2111, opened 2018-04-19
Add total exposure percentage to Performance Analyzer

stereohelix:
This patch adds total exposure as a percentage to Performance Analyzer.

The logger reports the exposure as a number between 0 and 1, depending on how much time is spent in trades. an exposure of 1 means the asset is bought immediately and never sold, an exposure of 0 means the asset is never bought. A lower total exposure may indicate a lower risk strategy when weighed against profit and/or sharpe ratio.

I think this could be a useful stat for optimization in general, but plan on referring to it specifically in gekkoga. I've been using this to help optimizing in gekkoga and find it quite helpful for my strategy, so I thought I'd add it here (and then patch gekkoga if it gets merged.)

### DIFF
--- a/plugins/performanceAnalyzer/logger.js
+++ b/plugins/performanceAnalyzer/logger.js
@@ -104,6 +104,7 @@ if(mode === 'backtest') {
     log.info(`(PROFIT REPORT) timespan:\t\t\t ${report.timespan}`);
     if(report.sharpe)
       log.info(`(PROFIT REPORT) sharpe ratio:\t\t\t ${report.sharpe}`);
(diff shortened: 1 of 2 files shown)


askmike:
Apologies for the late response.

I am loving this! Would you mind PRing this against this branch instead: https://github.com/askmike/gekko/tree/feature/sync-gekko-events

I think this touches code that has since been changed in that branch.

stereohelix:
Hi Mike,

No problem. I made a new PR (https://github.com/askmike/gekko/pull/2178) and added a couple other things I'd been talking about.

askmike:
closing this one in favor of #2178.

PR #2118, opened 2018-04-22
[WIP] Gekko broker

askmike:
Finally a proper execution engine. This is the first step to getting advanced orders (stop losses, take profits, etc) into Gekko. It's also a standalone library to easily create and manage advanced oders on all of Gekko's supported exchanges.

Docs: https://github.com/askmike/gekko/blob/gekko-trader/docs/gekko-broker/introduction.md
Status: https://github.com/askmike/gekko/blob/gekko-trader/docs/gekko-broker/introduction.md#status

After this PR lands (which will take a while) we can look into an API for strategies to use:

- Allow strategies to place stop losses
- Allow strategies to place take profits
- Allow strategies to decide order priority (should we trade RIGHT NOW at market price or we want the best possible price or somewhere in between)?


askmike:
See #2216 for follow up!

PR #2137, opened 2018-04-27
Set docker entrypoint to run gekko by default and use options in cmd

lutangar:
feature

When running a container in another mode than `--ui` we need to specifiy the base command again (`node gekko`).

When running the container in another mode we just need to specify the custom options. The default behavior is unchanged and will run the ui by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,5 +24,4 @@ EXPOSE 3000
 RUN chmod +x /usr/src/app/docker-entrypoint.sh
 ENTRYPOINT ["/usr/src/app/docker-entrypoint.sh"]
 
-
-CMD [ "npm", "start" ]
+CMD ["--config", "config.js", "--ui"]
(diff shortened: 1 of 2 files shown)


askmike:
I think most people want to run a single Gekko instance via the UI. Doing so should be easy enough, how would that work after this PR?

lutangar:
The default command was `npm start` and was running `node gekko --config config.js --ui`.
This PR use `node gekko` as the *entrypoint* and ` --config config.js --ui` as the *command*, so it basically don't change the default behavior and will run the *ui* by default.

Docker natural behavior is to allow to easily override the *command* part, here it's the configuration. 
For example, to run gekko in import mode, giving the docker image `gekko/gekko` exists, we could do:
```
docker run gekko/gekko --import --config config.js
```
There would be no need to override the *entrypoint*.

askmike:
Thanks for clearing that up, I haven't used Docker for a while.

That sounds awesome!

PR #2152, opened 2018-05-06
sample-config: fix mongodb host + datarange

xFFFFF:
I do not know if you like such changes (I wanted it to be user-friendly), you will reject it the most.

Fix host name, and example dataranges for import and backtest

host mongodb doent exist :D

localhost always exist

* **Other information**:
Peace


askmike:
Loving it, thanks!

PR #2153, opened 2018-05-06
Indicator/UO: Fix calculation of third ratio

joelving:
Fix error where secondLow was used as denominator in calculation of third ratio.

Bug fix

The second low result is used in calculating the third ratio.

The third low result is used instead.

* **Other information**:


strategies/indicators/UO.js
```diff
@@ -35,7 +35,7 @@ Indicator.prototype.update = function(candle) {
 
   var first = this.firstHigh.result / this.firstLow.result;
   var second = this.secondHigh.result / this.secondLow.result;
-  var third = this.thirdHigh.result / this.secondLow.result;
+  var third = this.thirdHigh.result / this.thirdLow.result;
```
lee101, line 38: looks like a clear bugfix to me :)

askmike:
great stuff, thanks!

askmike:
I would love a unit test similar to this one: https://github.com/askmike/gekko/blob/develop/test/indicators/ema.js for the UO indicator as a follow up PR, if anyone wants to do that that would be great!

Issue #2138, opened 2018-04-28
Cannot BACKTEST with varPPO strategy


shmutalov:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a bug**
[ ] bug report

**Action taken** (what you did)
Loaded BTC-XLM data from Binance (2018-02-01 - 2018-04-27)
Ran backtest with `varPPO` strategy:
![image](https://user-images.githubusercontent.com/8098898/39391802-e5ab1cce-4ac2-11e8-8c93-85085bf49452.png)

**Expected result** (what you hoped would happen)
Successful execution

**Actual result** (unexpected outcome)
```js
2018-04-28 08:59:00 (INFO):     Setting up Gekko in backtest mode
2018-04-28 08:59:00 (INFO):
2018-04-28 08:59:00 (INFO):     Setting up:
2018-04-28 08:59:00 (INFO):              Trading Advisor
2018-04-28 08:59:00 (INFO):              Calculate trading advice
2018-04-28 08:59:00 (INFO):              Using the strategy: varPPO
2018-04-28 08:59:00 (WARN):     TALIB indicators could not be loaded, they will be unavailable.
D:\Dev\git\gekko\strategies\indicators\PPO.js:9
  this.short = new EMA(config.short);
                             ^

TypeError: Cannot read property 'short' of undefined
    at new Indicator (D:\Dev\git\gekko\strategies\indicators\PPO.js:9:30)
    at Base.addIndicator (D:\Dev\git\gekko\plugins\tradingAdvisor\baseTradingMethod.js:308:34)
    at Base.bound [as addIndicator] (D:\Dev\git\gekko\node_modules\lodash\dist\lodash.js:729:21)
    at Base.method.init (D:\Dev\git\gekko\strategies\varPPO.js:27:8)
    at Base.bound [as init] (D:\Dev\git\gekko\node_modules\lodash\dist\lodash.js:729:21)
    at new Base (D:\Dev\git\gekko\plugins\tradingAdvisor\baseTradingMethod.js:87:8)
    at Actor.setupTradingMethod (D:\Dev\git\gekko\plugins\tradingAdvisor\tradingAdvisor.js:62:17)
    at Actor.bound [as setupTradingMethod] (D:\Dev\git\gekko\node_modules\lodash\dist\lodash.js:729:21)
[...]

PR #2157, opened 2018-05-07
Update varPPO.js

TheHalcyonSavant:
Bug fix for #2138

Backtesting with varPPO is failing at gekko\strategies\indicators\PPO.js:9

* **Other information**:

### DIFF
--- a/strategies/varPPO.js
+++ b/strategies/varPPO.js
@@ -2,12 +2,20 @@
 var _ = require('lodash');
 var log = require('../core/log');
 
+var util = require('../core/util');
+var fs = require('fs');
+var toml = require('toml');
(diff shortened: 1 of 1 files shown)


askmike:
The current config situation is a bit of a mess, but unfortunately this solution does not always work. Here is a gist:

There are 2 ways to run Gekko: via CLI and via the UI:

- CLI: one big config file in javascript is used.
- UI: multiple TOML config files are parsed (by the frontend) and all used together.

Gekko should automatically deal with the config, I will have a look why it's not properly passed to the varPPO when using the UI asap.

TheHalcyonSavant:
You're right. You can reject my PR. It was only a temporary solution for UI

PR #2162, opened 2018-05-10
Correct CCI Indicator

Dalekxy:
This is my first pull request on GitHub :)
New calculation  method for avgtp.
CCI Indicator is now correctly calculated.

After some weird backtests and debuging, I've noticed that cci.result is different from the CCI Indicator on TradingView.
In the previous version the average trending price (avgtp) was calculated with LRC(history),
But cci.result differs vastly from generally accepted formula CCI = (tp - SMA(tp))/(constant * md))
with tp = trending price, SMA(tp) = simple moving average of trending price, md = mean deviation

cci.result now matches with TradingView's CCI(hlc3). hlc3 means tp (=(high+low+close)/3)).

* **Other information**:


askmike:
This looks great, thanks!

Would you mind adding a test similar to these ones: https://github.com/askmike/gekko/blob/develop/test/indicators/ema.js

There was none for CCI just yet.

---

If that's to much work no problem!

askmike:
Merging this already.

Issue #1811, opened 2018-01-29
[Poloniex] Gekko does not properly calculate minimum orders


AidasK:
* **I'm submitting a ...**
[x] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
I have launched live bot at poloniex and saw thousand of lines with errors like:
```
2018-01-23 02:33:16 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:17 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:17 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:27 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:27 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:27 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:27 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:37 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:37 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:37 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:37 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:48 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:48 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:48 (ERROR):    unable to buy: null { error: 'Nonce must be greater than 151667122795101. You provided 151667122795100.' }
2018-01-23 02:33:48 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:58 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:58 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:33:58 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
2018-01-23 02:33:58 (DEBUG):    Poloniex returned an error, retrying..
2018-01-23 02:34:08 (ERROR):    unable to buy: null { error: 'Total must be at least 1.' }
[...]

PR #2167, opened 2018-05-11
Fix Poloniex minimum order calculations (issue #1811)

jonbrew:
Bug fix

Issue #1811 - Running live trading on Poloniex will cause the bot to be stuck in a buy or sell loop without the funds necessary to execute a trade when consecutive buy or sell signals are sent.

Implemented getLotSize to calculate minimum order amount based off of asset minimum and currency minimum
Moved markets object into a seperate file poloniex-markets.json to reduce bloat inside of poloniex.js

* **Other information**:


askmike:
I will merge this for now, but note that I already have another poloniex implementation working here: https://github.com/askmike/gekko/blob/gekko-trader/exchange/wrappers/poloniex.js

That will be merged as part of 0.6 in a few weeks.

PR #2170, opened 2018-05-12
Fixed QuadrigaCX market detection of outstanding orders

jcoder76:
- Fixed an error in the getOrder function where it was using lookup_oder instead of lookup_order as the remote path


jcoder76:
@cmroche this change fixes detection of the order books

askmike:
:+1:

Issue #2022, opened 2018-03-11
Issue with Gekko not retrieving QuadrigaCX balances


larathydo:
**Note: for support questions, please join our [Discord server](https://discord.gg/26wMygt)**

* **I'm submitting a ...**
[X] bug report
[ ] feature request
[ ] question about the decisions made in the repository

* **Action taken** (what you did)
When I turn on the trader after configuring my key/secret/username for QuadrigaCX, I see the following messages with debug turned on:

2018-03-11 14:53:26 (INFO):	Setting up:
2018-03-11 14:53:26 (INFO):		 Trader
2018-03-11 14:53:26 (INFO):		 Follows the advice and create real orders.
2018-03-11 14:53:26 (DEBUG):	getting ticker, balance & fee from quadriga
2018-03-11 14:53:26 (INFO):	

2018-03-11 14:53:27 (DEBUG):	Unable to find "CAD" in portfolio provided by exchange, assuming 0.
2018-03-11 14:53:27 (DEBUG):	Unable to find "LTC" in portfolio provided by exchange, assuming 0.
2018-03-11 14:53:27 (INFO):	trading at quadriga ACTIVE
2018-03-11 14:53:27 (INFO):	quadriga trading fee will be: 0.5%
2018-03-11 14:53:27 (INFO):	quadriga portfolio:
2018-03-11 14:53:27 (INFO):		 CAD: 0.000000000000
2018-03-11 14:53:27 (INFO):		 LTC: 0.000000000000

* **Expected result** (what you hoped would happen)
Gekko would have been able to retrieve my CAD/LTC balance from the API.

* **Actual result** (unexpected outcome)
Gekko assumes the balance is zero given it's inability to determine account balance.

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
I have double checked my API settings on Quadriga and it is set with the following options enabled:

-> List Balances Enabled (Display all of your account's fiat and crypto balance data.)
-> List Transactions Enabled (Display all of your account's trade, funding and withdrawal data.)
-> List Open Orders Enabled (Display list of Open Orders from your account.)
-> Lookup Order Enabled (Query an Order to fetch more details.)
-> Cancel Open Order Enabled (Ability to cancel an Open Order.)
-> Trading Enabled (Allow API to create Buy or Sell orders.)
-> Funding DISABLED (Display the current funding address.)
[...]

PR #2171, opened 2018-05-12
Fixed the precision handling in the QuadrigaCX market fixing buy/sell

jcoder76:
- Fixed an invalid cast when trying to buy/sell because it wasn't finding the market properly since the closure "this" object was not the exchange "this" object


jcoder76:
@cmroche this fixes the precision handling mentioned in #2022

askmike:
:+1:

PR #2172, opened 2018-05-12
Quadriga market fixes

jcoder76:
- Updated the supported markets and moved them to a separate JSON file with an order book
- Applied a proper fix for the uppercase currencies vs Quadriga order book format because there was inconsistencies between the Quadriga page using XBT/CAD which has order book btc_cad

### DIFF
--- a/exchanges/quadriga-markets.json
+++ b/exchanges/quadriga-markets.json
@@ -1,0 +1,136 @@
+{
+  "assets": [
+    "XBT",
+    "ETH",
+    "LTC",
+    "BCH",
+    "BTG"
+  ],
+  "currencies": [
+    "CAD",
+    "USD",
+    "XBT"
+  ],
+  "markets": [
+    {
+      "pair": [
+        "CAD",
+        "XBT"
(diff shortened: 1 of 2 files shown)


jcoder76:
@cmroche it is up to you if you want to pull this.  It is a breaking change for anyone with a config using the BTC order book since I changed that to match Quadriga's website as XBT.

jcoder76:
The top level declaration cleanup is in another pull request that should be accepted first.

askmike:
Thanks a lot!

> Updated the supported markets and moved them to a separate JSON file with an order book

I have no objections against this but note: the reason some other exchanges have this specified in a JSON file is because those are dynamically generated, see here: https://github.com/askmike/gekko/tree/develop/util/genMarketFiles

>  It is a breaking change for anyone with a config using the BTC order book since I changed that to match Quadriga's website as XBT.

There will be multiple breaking changes coming soon, I'm not too worried about this right now. This will be reflected in the changelog when we bump the stable release.

PR #2185, opened 2018-05-17
Create testi/indicators/cci.js

Dalekxy:
As requested from @askmike in #2162, I created a cci indicator test file.
Unfortunately I had to limit the decimals, because excel or gekko cause rounding mistakes in the calculation.

I also think it would be great if there had been a manual how to use the indicator test, it took me about an hour with google to find out, that I have to use "npm test" for starting the test cycle (I suppose you also have to install chai using "npm install chai").

No CCI test Indicator

npm test is completed without error

* **Other information**:

### DIFF
--- a/test/indicators/cci.js
+++ b/test/indicators/cci.js
@@ -1,0 +1,59 @@
+var chai = require('chai');
+var expect = chai.expect;
+var should = chai.should;
+var sinon = require('sinon');
+
+var _ = require('lodash');
+
+var util = require('../../core/util');
+var dirs = util.dirs();
+var INDICATOR_PATH = dirs.indicators;
+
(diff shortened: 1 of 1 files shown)


askmike:
> Unfortunately I had to limit the decimals, because excel or gekko cause rounding mistakes in the calculation.

That is not a problem.

> it took me about an hour with google to find out, that I have to use "npm test" for starting the test cycle (I suppose you also have to install chai using "npm install chai").

Ah yes! I will add a contribution doc explaining this kind of stuff. Apologies for that.

Minor question, are you able to upload the excel file (preferably here) so everyone can verify the numbers found in the test? Thanks!

PR #2192, opened 2018-05-18
Update other_software.md

xFFFFF:
Backtesters are sorted by the date the repository was created. From the oldest to the newest. That no one would be mad at me:D

We also place unofficial plugins, improvements that are found in forks or PR? Like: https://github.com/PixelGithub/gekko/blob/PixelCrunch-DiscordPlugin/plugins/discordbot.js


askmike:
Thanks for submitting this PR! I'll make sure it reflects on the online doc site asap.

> We also place unofficial plugins, improvements that are found in forks or PR?

Sure! If the page gets to big we figure it out from there.

PR #2201, opened 2018-05-23
Draw indicators on backtest using Highcharts

Antreasgr:
This is a feature. 

D3 library is used with no indicators shown

Uses highcharts to draw the backtest results with all the indicators

* **Other information**:


askmike:
Hey! Unfortunately highcharts is not compatible with Gekko for licensing reasons. You can find more in depth information if you search around on this issue tracker.

Note that I am working on a new UI that includes better charts as well as drawn indicator results. You can see some of that on this site: https://gekkoplus.com/ - on top of that site the UI shown there will also be the new UI for the open source gekko project. It will be written completely in react :)

Antreasgr:
Its ok i understand not wanting to merge this. But i use it in my own installation, and maybe some people would want this too. Leaving it here for reference.

askmike:
You can also submit another PR to add your fork to this page: https://github.com/askmike/gekko/blob/develop/docs/extending/other_software.md once I update the docs that are now live here: https://gekko.wizb.it/docs/extending/other_software.html I think it will get more exposure :)

I'm trying to tidy up this issue tracker a bit.

PR #2203, opened 2018-05-24
UI - Live runner add type (paper trader or live) and count trades

julesGoullee:
This pr improve ui to add some meta on live runner:
- type of runner (live or paper trader)
- Number of trades
screen shot 2018-05-24 at 17 26 00


askmike:
Awesome! Keep in mind that I am working on a new UI that includes all of this (and more). But it will take a while to complete, see this: https://forum.gekko.wizb.it/thread-1429.html

PR #2208, opened 2018-05-26
Extra Indicators.

Gab0:
Extra indicators, from https://github.com/Gab0/gekko-extra-indicators. Those are the stable ones. 
Strategies to compare their output with TA-LIB counterparts are included.

### DIFF
--- a/strategies/indicators/ADX.js
+++ b/strategies/indicators/ADX.js
@@ -1,0 +1,42 @@
+// Average Directional Movement Index indicator;
+// usable on gekko trading bot. Same license as gekko.
+// "ported" from tulip: https://tulipindicators.org/adx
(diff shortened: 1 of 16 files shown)


askmike:
Awesome stuff! Before we merge I'd like to get the tests in line with the current test suite:

- pre calc the TAlib numbers and hardcode them in the tests (with comments on how they were generated)
- put the tests here: https://github.com/askmike/gekko/blob/develop/test/indicators/macd.js

---

I am already super happy with this PR, let me know if you don't feel like doing it, in that case I will do it myself.

DragonBtc93:
How do we test? Would really like to help this project. Really think Gekko could be be the future of bot trading.

PR #2213, opened 2018-05-28
Ui Strat runner: add type of runner, candle size , history size

julesGoullee:
Add missing data on live runner, that are set on start runner:
- Type of trader (paper or live)
- candle size
- history size (warmup)
screen shot 2018-05-28 at 02 37 57


askmike:
:+1:

PR #2214, opened 2018-05-28
Ui Strat runner fix profit report value and add percentage

julesGoullee:
Fix live runner profit report value, add percentage

screen shot 2018-05-28 at 03 02 11


web/vue/src/components/gekko/singleStratrunner.vue
```diff
@@ -57,15 +57,15 @@
                 .grd-row-col-3-6 {{ round(report.balance) }}
               .grd-row
                 .grd-row-col-3-6 Market
-                .grd-row-col-3-6 {{ round(report.market) }} {{ data.watch.currency }}
+                .grd-row-col-3-6 {{round(report.market / 100 * report.startPrice)}} {{ data.watch.currency }} ({{ round(report.market) }} %)
```
askmike, line 60: This now describes market as notional with % behind it? Sounds cool!

askmike (approved):

askmike:
Thanks a lot! I will try to push a release out before the new UI asap, keep this in mind though:

> I am working on a new UI that includes all of this (and more). But it will take a while to complete, see this: https://forum.gekko.wizb.it/thread-1429.html

web/vue/src/components/gekko/singleStratrunner.vue
julesGoullee: yep absolute value in currency en relative in %

julesGoullee:
I looking for new ui before a while I prefer a this meta & fix incorrect value :hear_no_evil:

PR #2244, opened 2018-06-16
updating documentation location

arimourao:
Docs update

The links for the current documentation are broken. Also the description of the links are pretty vague.

The links point to the right location of the current documentation


askmike:
Thanks for your PR. With the current URLs they work on github, in a standalone markdown editor. When pushed to the docs website the links are converted automatically, see here: https://github.com/askmike/gekko.wizb.it/blob/master/md.js#L23-L51

Do you experience any issues with the current links?

PR #2263, opened 2018-06-30
PostgreSQL connection pooling / candle writing fix

mark-sch:
PostgreSQL connection handling bugfix. Introduced connection pooling and resolved race condition conflicts against database, e.g. prevents table creation attempt before db is created, candle writing when no table is present ... 

Before the fix, often no candle data was written do postgresql db.

Fix with connection pooling and transactional upsert candle writing.

* **Other information**:
Requires PostgreSQL 9.5+


package.json
```diff
@@ -66,6 +66,8 @@
   "devDependencies": {
     "chai": "^4.1.2",
     "mocha": "^5.0.0",
+    "mongojs": "^2.6.0",
+    "pg": "^7.4.3",
```
askmike: Would you mind not removing this here? `pg` shouldn't be a dev dependency.

importers/exchanges/kraken.js
```diff
@@ -43,9 +43,8 @@ var handleFetch = (err, trades) => {
         return fetcher.emit('trades', []);
     }
 
-    var last = moment.unix(_.last(trades).date);
+    var last = moment.unix(_.last(trades).date).utc();
```
askmike, line 46: you mentioned some bugs related to UTC, do you have more information regarding what this exactly fixes?

plugins/postgresql/handle.js
```diff
@@ -1,6 +1,8 @@
-var _ = require('lodash');
-var fs = require('fs');
-var pg = require('pg');
+const _ = require('lodash');
+const fs = require('fs');
+const pg = require('pg');
+//const pg = new Pool();
```
askmike: would you mind removing these comments?

plugins/postgresql/reader.js
```diff
@@ -19,163 +19,180 @@ Reader.prototype.mostRecentWindow = function(from, to, next) {
...
+      done();
+      next(null, _.first(rows));
+    });
+  });  
 }
 
 Reader.prototype.close = function() {
-  this.db.end();
+  //obsolete due to connection pooling
+  //this.db.end();
```
askmike, line 195: does this mean that the process can completely quit without having to close anything? If yes please remove the completed Reader.prototype.close` function!

askmike (commented):
Great stuff! Looks like a lot of work went into this, big thhanks. See few small comments :)

package.json
mark-sch: removed both packages

importers/exchanges/kraken.js
mark-sch: This helped me to get --imports completely, otherwise 2h were missing. UTC vs. GMT+2 prob.

plugins/postgresql/handle.js
mark-sch: 2 comments removed

plugins/postgresql/reader.js
mark-sch: Exactly, this is the goal of connection pooling. You do not close a connection anymore, instead you hand it back to the pool with callback done(). Not all DB plugins support this pooling method and therefore I think we can not remove this call from the core. For now, I left it in postgresql plugin code to do nothing, so it can be still called from core without any errors.

mark-sch:
I just tested current v0.6.1 code (without this postgres patch). I have the same error like in v0.5 before - no candles are written to db in realtime mode. I suggest to apply this patch to v0.5 and v0.6.

plugins/postgresql/reader.js
askmike: :+1:

askmike:
Works great! I will continue to work on this, but since this is such a big improvement on the postgresql adapter I am merging this already.

Thanks a lot!

mark-sch:
Hi Mike, I am happy it works for you. My tests today with this postgres pooling on v0.6.x went also fine. Btw, v0.6 was an awesome step, I am going to switch into production soon. For this step, I am close to finish hitbtc exchange wrapper conversion for v0.6.

Issue #2266, opened 2018-07-04
v0.6.0: "sed: can't read /usr/src/app/web/vue/UIconfig.js: No such file or directory"


bald123:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[X ] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Pulled v0.6.0, built Docker container and started it.

**Expected result** (what you hoped would happen)
A running container without errors.

**Actual result** (unexpected outcome)
I get the error: "sed: can't read /usr/src/app/web/vue/UIconfig.js: No such file or directory" and the webinterface is not running.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

bald123:
docker-entrypoint.sh references this file.

PR #2267, opened 2018-07-04
Fix UIconfig path for docker

arcxyz:

Bug fix

#2266 Cannot start gekko using docker

After changing the UIConfig path to the correct one it works!

* **Other information**:


askmike:
Awesome, thanks!

Issue #2284, opened 2018-07-07
Send notification from my strategy


thanhlmm:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[ ] bug report
[ ] question about the decisions made in the repository
[x] feature request

**Action taken** (what you did)
Sometimes, I need to get more information currently running in my strategy (New advice with additional information, current status of my running strategy,...). So I need a function for send notification to my plugin (Slack, Email, Discord,...).

**Expected result** (what you hoped would happen)
`this.notice('Hello, I can send additional information here!');` and I can see that `notification` in Slack or Email

**Actual result** (unexpected outcome)
None

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)
None

askmike:
great idea, discussion on implementation in #2285.

PR #2285, opened 2018-07-07
Add function for send notification from strategy

thanhlmm:
#2284

```
this.notice('Hello, I can send additional information here!');
```
To send a notification from my strategy file

* **Other information**:

### DIFF
--- a/docs/internals/events.md
+++ b/docs/internals/events.md
@@ -14,6 +14,7 @@ Note that all events from Gekko come from a plugin (with the exception of the `c
 - [stratWarmupCompleted](#stratWarmupCompleted-event): When the strategy is done warming up.
 - [advice](#advice-event): Every time the trading strategy is fed a new candle.
 - [stratUpdate](#stratUpdate-event): Every time the strategy has processed a new strat candle.
+- [stratNotification](#stratNotification-event): Every time the strategy emit new strategy notification.
 - [tradeInitiated](#tradeInitiated-event): Every time a trading plugin (either the live trader or the paper trader) is going to start a new trade (buy or sell).
 - [tradeCompleted](#tradeCompleted-event): Every time a trading plugin (either the live trader or the paper trader) has completed a trade.
 - [tradeAborted](#tradeAborted-event): Every time a trading plugin (either the live trader or the paper trader) has NOT acted on new advice (due to unsufficiant funds or a similar reason).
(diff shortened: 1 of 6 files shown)


askmike:
Hey @thanhlmm, thanks for the PR!

I like this idea, just a few tiny things:

- can we call it something else than "notification"? It's not very descriptive that it can only come from a strategy. Maybe something like "strategyNotification" or "strategyMessage" (also instead of notice we can have notify).
- can you add it to the docs here: https://github.com/askmike/gekko/blob/develop/docs/internals/events.md

askmike:
Also we probably need to think about how we want those to behave during backtests, if these notifications are for realtime stuff they should probably be ignored before even being propagated.

What kind of notifications do you feel like sending?

askmike:
fixes #2284.

thanhlmm:
> can we call it something else than "notification"? It's not very descriptive that it can only come from a strategy. Maybe something like "strategyNotification" or "strategyMessage" (also instead of notice we can have notify).

`this.notify('Somethings');` is a nice name !

> can you add it to the docs here: https://github.com/askmike/gekko/blob/develop/docs/internals/events.md

Sure then

> Also we probably need to think about how we want those to behave during backtests, if these notifications are for realtime stuff they should probably be ignored before even being propagated.

It rarely when user config any notification (email, slack,...) while backtesting but we should handle this case. I think we should have a config for determining if users should send `strategyNotification` while backtesting or not. In paper trade or real trade mode, I sure they will need `strategyNotification` to work.

How do you think, @askmike ?

askmike:
Can you give an example of a notification you want to send?

thanhlmm:
I use `Gekko` as a tool to watch market and give advice. So that, I need more information about what is happening like:
"We got 10 green candles, should we go long?"
"RSI is going beyond 120, MACD diff is now 20,..."

Or sometime user wanna get information of what is about to go like
"I detect an uptrend, wait for 2 more confirm candles to go `Long`"

askmike:
That makes sense, and in that case it makes sense to have the possibility of having them available in backtests (to plot the notifications on a chart or so).

> this.notify('Somethings'); is a nice name !

For the inside strat? Yes that's great.

> strategyNotification

I like this one for the event name! Do you feel like fixing these names in a commit? After that (& added to the event doc) I can merge it :)

thanhlmm:
Ok, I will make new pull request soon

askmike:
You don't have to perse, you can push new commits to your branch `feature-notification` and they will be automatically show up here. If you run into anything just let me know.

thanhlmm:
Hey @askmike ,
I have just edit some code base on your idea.
I changed `strategyNotification` event name to `stratNotification` for the same naming convention in your code base.

askmike:
great stuff! Thanks a lot :)

askmike:
great stuff! Thanks a lot :)

thanhlmm:
I'm glad to contribute for great product 😎

PR #2304, opened 2018-07-13
Remove docker restart always

sylvainbaronnet:
Remove `restart: always` from docker-compose.yml ? It made the project "up" automatically every time docker start and it's a bit anoying as it takes a port used by other project.


askmike:
:+1: thanks

Issue #2150, opened 2018-05-05
Cannot run Gekko


ChadTaljaardt:
**I'm submitting a ...**
[x] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
```
 node gekko --ui
```

**Expected result** (what you hoped would happen)
Gekko to run

**Actual result** (unexpected outcome)
```
 gekko git:(stable) ✗ node gekko --ui


    ______   ________  __    __  __    __   ______
   /      \ /        |/  |  /  |/  |  /  | /      \
  /$$$$$$  |$$$$$$$$/ $$ | /$$/ $$ | /$$/ /$$$$$$  |
  $$ | _$$/ $$ |__    $$ |/$$/  $$ |/$$/  $$ |  $$ |
  $$ |/    |$$    |   $$  $$<   $$  $$<   $$ |  $$ |
  $$ |$$$$ |$$$$$/    $$$$$  \  $$$$$  \  $$ |  $$ |
  $$ \__$$ |$$ |_____ $$ |$$  \ $$ |$$  \ $$ \__$$ |
  $$    $$/ $$       |$$ | $$  |$$ | $$  |$$    $$/
   $$$$$$/  $$$$$$$$/ $$/   $$/ $$/   $$/  $$$$$$/

        Gekko v0.5.14
        I'm gonna make you rich, Bud Fox.


/Users/chadtaljaardt/Desktop/tradingbot/gekko/node_modules/promisify-node/utils/args.js:9
  var args = func.toString().match(/function\s.*?\(([^)]*)\)/)[1];

TypeError: Cannot read property '1' of null
    at module.exports (/Users/chadtaljaardt/Desktop/tradingbot/gekko/node_modules/promisify-node/utils/args.js:9:63)
    at processExports (/Users/chadtaljaardt/Desktop/tradingbot/gekko/node_modules/promisify-node/index.js:61:29)
    at module.exports (/Users/chadtaljaardt/Desktop/tradingbot/gekko/node_modules/promisify-node/index.js:164:10)
    at Object.<anonymous> (/Users/chadtaljaardt/Desktop/tradingbot/gekko/web/routes/scanDateRange.js:4:14)
    at Module._compile (internal/modules/cjs/loader.js:678:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:689:10)
    at Module.load (internal/modules/cjs/loader.js:589:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:528:12)
[...]

PR #2309, opened 2018-07-15
Update promisify and sqlite versions

hilkoc:
Necessary package version updates as suggested in
https://github.com/askmike/gekko/issues/2150#issuecomment-394829897


askmike:
Ah the promisify update looks great, that might solve a problem a lot of people have been having. As for the sqlite upgrade, I want to quickly make sure there are precompiled binaries availble for this version before I merge this.

askmike:
> node-pre-gyp WARN Using request for node-pre-gyp https download 
> [sqlite3] Success: "/Users/pear/projects/gekko/node_modules/sqlite3/lib/binding/node-v59-darwin-x64/node_sqlite3.node" is installed via remote

It is available for me (on MacOS), assuming it's also available on windows.

Thanks!

Issue #2326, opened 2018-07-17
Multiple advise together kills gekko


hiyan:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[X] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Testing exchange using DEBUG_toggle-advice.js.

**Expected result** (what you hoped would happen)

**Actual result** (unexpected outcome)
Multiple advice within "same time period" crashes gekko. Rare situation... i know :)
note: cancelOrder will callback(error) if id is not valid.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

logs:
```
2018-07-17 18:30:31 (DEBUG):    scheduling ticks
2018-07-17 18:31:32 (DEBUG):    trigger SHORT

2018-07-17 18:31:32 (INFO):     We have new trading advice!
2018-07-17 18:31:32 (INFO):              Position: short
2018-07-17 18:31:32 (INFO):              Market price: 0.0703
2018-07-17 18:31:32 (INFO):              Based on market time: 2018-07-17 10:30:00

2018-07-17 18:31:32 (INFO):     NOT selling, already no exposure

2018-07-17 18:35:32 (DEBUG):    trigger LONG

2018-07-17 18:35:32 (INFO):     We have new trading advice!
2018-07-17 18:35:32 (INFO):              Position: long
2018-07-17 18:35:32 (INFO):              Market price: 0.0706
2018-07-17 18:35:32 (INFO):              Based on market time: 2018-07-17 10:32:00

2018-07-17 18:35:32 (INFO):     Trader Received advice to go long. Buying  ETH
2018-07-17 18:35:32 (DEBUG):    Creating order to buy 0.0561657223796034 ETH
2018-07-17 18:35:32 (DEBUG):    trigger SHORT

2018-07-17 18:35:32 (INFO):     We have new trading advice!
2018-07-17 18:35:32 (INFO):              Position: short
2018-07-17 18:35:32 (INFO):              Market price: 0.0706
2018-07-17 18:35:32 (INFO):              Based on market time: 2018-07-17 10:34:00

2018-07-17 18:35:32 (INFO):     Received advice to sell however Gekko is already in the process to undefined
[...]

PR #2331, opened 2018-07-18
temp workaround for #2326

hiyan:
temp workaround for #2326


askmike:
Thanks for the PR! Unfortunately this is not a good workaround, it has 2 problems:

- The problem is inside the trader, changing this specific strategy does not make the problem go away (for other strategies for example).
- this PR breaks this strategy inside all backtests, since during a backtest it will most definitely fire more than 3 times a second.

hiyan:
Thanks. I dont understand the trader to do any meaningful fixes. This is something i added to just continue testing.

PR #2335, opened 2018-07-19
 v0.6 support for bitx (aka Luno)

hiyan:
Update exchange for v0.6.


exchange/wrappers/bitx.js
```diff
@@ -0,0 +1,348 @@
+const BitX = require("bitx");
+const _ = require('lodash');
+const moment = require('moment');
+const log = require('../../core/log');
```
askmike: can you not import anything from outside the exchange folder? That way people can use Gekko Broker as a standalone lib. Thanks!

askmike:
Woa great PR Thanks a lot! Just one small nit, see comment :)

Side note: did this take you very long and where the docs very clear?

exchange/wrappers/bitx.js
hiyan: Alright... i'll push an update later.

Curious, how do you plan to handle logging for exchanges moving forward?

hiyan:
I didnt depend on the docs much, just did an quick diff of the docs and mainly "diff-ing" the new binance + polonex exchanges.

Oh yes, this reminds me, I had to add in roundAmount and roundPrice. This wasnt documented.

askmike:
Seems to work great!

askmike:
TODO Follow up (I will pick those up, unless you want to do another PR):

- fix dependency: should be managed by gekko broker (`gekko/exchange/package.json`)
- rename exchange name + file to luno.
- add to GB docs that luno is now supported.

hiyan:
Sure.. that seems simple enough :)
I've already changed bitx/luno support in supported_exchanges.md. Is there anywhere else that needs editing?

PR #2352, opened 2018-07-23
rename bitx to luno

hiyan:
rename bitx to luno


askmike:
Woa awesome! I can remove this from my todo list now.

This is much appreciated.

PR #2356, opened 2018-07-23
Fix method name for stratNotification event

ptdev:
The documentation was missing the "process" prefix on the method name for the stratNotification event


askmike:
:+1: thanks

PR #2374, opened 2018-07-27
Change two debug messages to warn messages

crypto49er:

feature

User will not see these two messages in terminal if they don't have debug set to true.

User will see they messages as warnings

* **Other information**:

The "Not creating order" message was a warn message in Gekko 0.5.


askmike:
great idea!

PR #2417, opened 2018-08-08
Update Luno

hiyan:
- re-add logging
- Retry insufficient balance due to server update delay
- Enhanced getPorfolio
- getOrder now return fees


exchange/wrappers/luno.js
```diff
@@ -243,54 +312,55 @@ Trader.prototype.checkOrder = function(order, callback) {
...
     if (data && !data.success) {
-      console.log('cancelOrder() --> status:', data.success);
+      log('cancelOrder() --> status:', data.success);
       return callback(undefined, false);
     }
 
     this.checkOrder(order, (error, orderStatus) => {
       if (error) {
-        console.log(name, 'cancelOrder\'s checkOrder failed. What do i do here?');
+        log(name, 'cancelOrder\'s checkOrder failed. What do i do here?');
```
askmike, line 337: Good question.. At that point we are sure that the order was cancelled but we do not have fill information. Let's keep it like this: we bubble up the error and have Gekko figure out how to proceed.

Assuming it was or was not partially filled will probably lead to nasty bugs down the line.

askmike (commented):
Great stuff! I'll be testing this out asap but it looks very nice. Thanks a lot :)

exchange/wrappers/luno.js
```diff
@@ -51,20 +67,45 @@ const recoverableErrors = [
...
+
+      if (includes(error.message, ['error 500'])) {
+        error.notFatal = true;
+        error.backoffDelay = 15000;
+      }
+
+      if (includes(error.message, ['Insufficient balance'])) {
+        error.notFatal = true;
+        error.backoffDelay = 2000;
+        tradeAttempt++;
```
askmike, line 104: this is okay for now, but we should aim to avoid these kind of things in the future!

askmike:
Seems to be working fine! (based on limited tests)

Issue #2457, opened 2018-08-18
gekko instances (load/workers/loadCandles/child) never killed when using Chrome, consuming memory


eusorov:
**I'm submitting a ...**
[x ] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
1. In Chrome start live gekko (market watcher)
2. close Chrome and open again.
3. click on live market watcher, now loadCandles/child get instantiated, with ps -aux | grep gekko you can see it.
4. close Chrome and open again and click market watcher. Again new instance of is created.
5. now leave Chrome. This intances willl run forever on a server and will cause OutOfMemory.

With Firefox I can not reproduce it, the instances of loadCandles are not created? with ps -aux | grep gekko I can not see any of them.
And also not tried with IE.

**Expected result** (what you hoped would happen)
When closing chrome all  loadCandles/child processes should be killed.

**Actual result** (unexpected outcome)
Intances of oad/workers/loadCandles/child will run forever on a server and will cause OutOfMemory.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

example with one UI, one market watcher and 3 loadCandle children.
ps -aux | grep gekko
node gekko.js --config sample-config.js --ui
evgeny   29230  0.1  0.3 816976 50868 pts/4    Sl+  14:32   0:00 /home/evgeny/.nvm/versions/node/v10.9.0/bin/node /home/evgeny/workspace/angular/gekko-master/core/workers/pipeline/child
evgeny   29252 60.8  0.5 632492 83664 pts/4    Rl+  14:32   5:42 /home/evgeny/.nvm/versions/node/v10.9.0/bin/node /home/evgeny/workspace/angular/gekko-master/core/workers/loadCandles/child
evgeny   29374 59.9  0.5 630380 82112 pts/4    Rl+  14:33   5:00 /home/evgeny/.nvm/versions/node/v10.9.0/bin/node /home/evgeny/workspace/angular/gekko-master/core/workers/loadCandles/child
evgeny   29467 59.3  0.5 630992 81700 pts/4    Sl+  14:34   4:18 /home/evgeny/.nvm/versions/node/v10.9.0/bin/node /home/evgeny/workspace/angular/gekko-master/core/workers/loadCandles/child

askmike:
Nice find. And chrome should not have much to do with it. Will test

eusorov:
[...]

PR #2467, opened 2018-08-21
bugfix #2457

eusorov:
Sometimes child.kill() failes. Better disconnect the parent and let the
child exit itself.

bugfix #2457

Sometimes child.kill() failes.

child get killed

* **Other information**:


askmike:
Nice one, thanks!

Issue #762, opened 2017-05-29
Request: multi timeframes emit to strategy


thegamecat:
Hi again, Tomih posted something a while ago about his fork here: https://github.com/tomih/gekko which suggest we can use multiple time frames in 1 strategy. Is this something that be merged into the gekko Master?

I'm manually working with an effective strategy at the moment and would like to automate it but it does need the read across multiple time frames.

askmike:
There are no plans for that, main reasons are that this goes against the main design principles of Gekko:

The idea of Gekko is that it is a simple starters kit for building TA strats, keeping simplicity is achieved by sacrificing more advanced (and complex) functionality that only a small portion of users would need but makes everything more confusing for everything else. Examples of this are:

- Simple flow based codebase: 90% of the code for importing is the same as for running backtests is the same as running against a live market.
- No possibilities for execution strategies: your strat advices long or short and based on that portfolio will be reallocated ASAP.
- No data below 1 minute resolution
- etc.

---

Gekko is build modular so not to say it's not possible, just that I won't be supporting it (I won't add features that won't be accessible through the UI - and I  don't think this is easy to add to the UI in a way that people actually understand what is going on).

More details on this specific topic can be found here: https://github.com/askmike/gekko/pull/737#issuecomment-299634318

askmike:
That's not to say I am not open to PRs, if someone opens an issue to go into details on how to do this properly and than submits a PR that implements it (with the intention to keep it up to date as well) it might just land in Gekko at some point.

generalectric:
@askmike, can we talk more about this? Ok so you do not want to implement multiple candlesize support due to the fact that it wouldnt be easily worked into the UI. Ok... fair enough. So cant we just write large candle strat to control small candle strat. Large candle strat instead of buy or sell would start or stop small candle strategy process and listen... then upon advice execute... Or... have the small candle strat just execute the advice then pass back the information to the parent process where it can be emitted as a trade for the ui. The problem is small candle strats have ups and downs... but they perform differently during the big candle ups and downs... So what you end up with is settings that are optimal over the entire spectrum... or settings that are overall optimal instead of settings that are geared toward a specific market trend. Now my small candle strat has settings for up and down trends... but it needs to have total different settings for those ups and downs based off of large candle indicators to maximize results. Any thoughts or considerations i should have going into this? I know i dont always take the right approach but if adding multi candlesize support to gekko isnt an option i think this is what im left with unless you have a better idea?

askmike:
[...]

Issue #2461, opened 2018-08-20
Market Watcher: Strat Calculations queue when no trades available for processing


storshun:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[ x] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
1. "No new trades" are available from the exchange (binance in this case) and 2 or more candles of time elapse
2. A new trade is processed from the exchange at the next "request" interval

**Expected result** (what you hoped would happen)
1 candle of strategy calculation should be processed

**Actual result** (unexpected outcome)
All strategy calculations that should have run at the end of each candle, but no trades were available to process, ends up running at the same time as soon as a new trade is processed.

For instance: 
Candle Size: 3 Minutes
Strat: CCI

On XLM/ETH - No trades occurred for about 8 minutes.
As soon as a new trade could be processed -  and was processed - 3 CCI calculations were processed, and resulted in a false indication of a trend for 2 candles.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

The scenario above should - in my opinion - result in a single calculation and not trigger subsequent trend indications off a single data-point. Ideally, any time no data is available for a candle, that candle should not contribute to the calculation of a trend.

askmike:
This is a design decision! The hard thing comes from some exchanges (in some scenarios) not sending a trade that happened at 11:59 until something like 12:06 (7 minutes later). By waiting for the next trade Gekko is not making any assumptions about lag anywhere between ME and Gekko (slow APIs under load, network issues, etc).

Note that is only a problem in very iliquid markets & on small candles. If you trade on 1 hour candles you'll hardly notice this when there are some trades.

There are a number of reasons why running Gekko with small candles on an illiquid market is a bad idea. For example see https://gekko.wizb.it/docs/features/backtesting.html#Simplified-simulation and https://github.com/askmike/gekko/issues/2380#issuecomment-408744682. I will try to explain this better in the docs.

Closing this but feel free to comment and let me know your thoughts.

thisisloze:
[...]

PR #2429, opened 2018-08-11
[WIP] native trailing stops

askmike:
More background here: https://forum.gekko.wizb.it/thread-57314.html

TODO:

- [x] core trailing stop class
- [x] unit tests for core trailing stop class
- [x] strategy trigger API
- [x] trigger events
- [x] implement in paper trader
- [x] implement in live trader
- [x] test! Both paper trader and live trader.

Log when running with live trader:

    root@vps459808:~/gekko# node gekko -c b.js

        ______   ________  __    __  __    __   ______
       /      \ /        |/  |  /  |/  |  /  | /      \
      /$$$$$$  |$$$$$$$$/ $$ | /$$/ $$ | /$$/ /$$$$$$  |
      $$ | _$$/ $$ |__    $$ |/$$/  $$ |/$$/  $$ |  $$ |
      $$ |/    |$$    |   $$  $$<   $$  $$<   $$ |  $$ |
      $$ |$$$$ |$$$$$/    $$$$$  \  $$$$$  \  $$ |  $$ |
      $$ \__$$ |$$ |_____ $$ |$$  \ $$ |$$  \ $$ \__$$ |
      $$    $$/ $$       |$$ | $$  |$$ | $$  |$$    $$/
       $$$$$$/  $$$$$$$$/ $$/   $$/ $$/   $$/  $$$$$$/

            Gekko v0.6.5
            I'm gonna make you rich, Bud Fox.

    2018-08-27 12:18:43 (INFO):     Setting up Gekko in realtime mode
    2018-08-27 12:18:43 (INFO):
    2018-08-27 12:18:43 (INFO):     Setting up:
    2018-08-27 12:18:43 (INFO):              Candle writer
    2018-08-27 12:18:43 (INFO):              Store candles in a database
    2018-08-27 12:18:43 (INFO):

    2018-08-27 12:18:43 (INFO):     Setting up:
    2018-08-27 12:18:43 (INFO):              Trading Advisor
    2018-08-27 12:18:43 (INFO):              Calculate trading advice
    2018-08-27 12:18:43 (INFO):              Using the strategy: DEBUG_single-advice
    2018-08-27 12:18:43 (INFO):

    2018-08-27 12:18:43 (INFO):     Setting up:
    2018-08-27 12:18:43 (INFO):              Trader
    2018-08-27 12:18:43 (INFO):              Follows the advice and create real orders.
    2018-08-27 12:18:43 (INFO):

    2018-08-27 12:18:49 (INFO):              Portfolio:
[...]

PR #2468, opened 2018-08-21
Feature strategy #762

eusorov:
Feature #762: pass one minute candle to strategy for stoploss or making with CandleBatcher other candles.

I use in my strategy various candles and the easy way to implement it it to pass one min to the strategy. There is also default implementation in baseTradingStrategy if the strategy don't implemnt updateOneMin.

Now the strategy get only fixed candle which we can specify at the beginning. But for stoploss and other complex strategies we need different candle sizes.

There is a default implementation of updateOneMin in baseTradingStrategy. In the strategy we can now define updateOneMin(candle) and get the small candles.

* **Other information**:

### DIFF
--- a/plugins/tradingAdvisor/tradingAdvisor.js
+++ b/plugins/tradingAdvisor/tradingAdvisor.js
@@ -86,6 +86,9 @@ Actor.prototype.setupStrategy = function() {
 // HANDLERS
 // process the 1m candles
 Actor.prototype.processCandle = function(candle, done) {
+  // notify strategy about one min
+  this.strategy.updateOneMin(candle);
+
   this.candle = candle;
   const completedBatch = this.batcher.write([candle]);
   if(completedBatch) {
--- a/config/strategies/MACD_multiple_timeframes.toml
+++ b/config/strategies/MACD_multiple_timeframes.toml
@@ -1,0 +1,8 @@
+short = 10
+long = 21
+signal = 9
+
+[thresholds]
+down = -0.025
+up = 0.025
+persistence = 1
--- a/core/workers/loadCandles/child.js
+++ b/core/workers/loadCandles/child.js
@@ -22,3 +22,7 @@ process.on('message', (m) => {
   if(m.what === 'start')
     start(m.config, m.candleSize, m.daterange);
 });
+
+process.on('disconnect', function() {
+  process.exit(0);
+})
--- a/core/workers/loadCandles/parent.js
+++ b/core/workers/loadCandles/parent.js
@@ -56,7 +56,7 @@ module.exports = (config, callback) => {
 
     // else we are done and have candles!
     done(null, m);
-    child.kill('SIGINT');
+    this.disconnect();
   });
 
   child.on('exit', code => {
--- a/plugins/tradingAdvisor/baseTradingMethod.js
+++ b/plugins/tradingAdvisor/baseTradingMethod.js
@@ -66,6 +66,9 @@ var Base = function(settings) {
   if(!this.onTrade)
     this.onTrade = function() {};
 
+  if(!this.updateOneMin) //get OneMinCandles!
+    this.updateOneMin = function() {};
+
   // let's run the implemented starting point
   this.init();
 
--- a/strategies/MACD_multiple_timeframes.js
+++ b/strategies/MACD_multiple_timeframes.js
@@ -1,0 +1,140 @@
+/*
+
+  MACD - DJM 31/12/2013
+
+  (updated a couple of times since, check git history)
+
+ */
+
+// helpers
+var _ = require('lodash');
+var log = require('../core/log.js');
+
+// let's create our own method
+var method = {};
+
+// prepare everything our method needs
+method.init = function() {
+  // keep state about the current trend
+  // here, on every new candle we use this
+  // state object to check if we need to
+  // report it.
+  this.trend = {
+    direction: 'none',
+    duration: 0,
+    persisted: false,
+    adviced: false
+  };
+
+  // how many candles do we need as a base
+  // before we can start giving advice?
+  this.requiredHistory = this.tradingAdvisor.historySize;
+
+  // define the indicators we need
+  this.addIndicator('macd', 'MACD', this.settings);
+}
+
+// what happens on every new candle?
+method.update = function(candle) {
+  // nothing!
+}
+
+// for debugging purposes: log the last calculated
+// EMAs and diff.
+method.log = function() {
+  var digits = 8;
+  var macd = this.indicators.macd;
+
+  var diff = macd.diff;
+  var signal = macd.signal.result;
+
+  log.debug('calculated MACD properties for candle:');
+  log.debug('\t', 'short:', macd.short.result.toFixed(digits));
+  log.debug('\t', 'long:', macd.long.result.toFixed(digits));
+  log.debug('\t', 'macd:', diff.toFixed(digits));
+  log.debug('\t', 'signal:', signal.toFixed(digits));
+  log.debug('\t', 'macdiff:', macd.result.toFixed(digits));
+}
+
+method.check = function() {
+  var macddiff = this.indicators.macd.result;
+
+  if(macddiff > this.settings.thresholds.up) {
+
+    // new trend detected
+    if(this.trend.direction !== 'up')
+      // reset the state for the new trend
+      this.trend = {
+        duration: 0,
+        persisted: false,
+        direction: 'up',
+        adviced: false
+      };
+
+    this.trend.duration++;
+
+    log.debug('In uptrend since', this.trend.duration, 'candle(s)');
+
+    if(this.trend.duration >= this.settings.thresholds.persistence)
+      this.trend.persisted = true;
+
+    if(this.trend.persisted && !this.trend.adviced) {
+      this.trend.adviced = true;
+      this.advice('long');
+    } else
+      this.advice();
+
+  } else if(macddiff < this.settings.thresholds.down) {
+
+    // new trend detected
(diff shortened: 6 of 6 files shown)


askmike:
> But for stoploss

Note that I am very close to releasing "native stoplosses" the first version will poll the ticker every few seconds (more stable than simple watching the close of 1 min candles) in the live trader, paper traders defers to 1 min candle close. But I'm planning on using actual native exchange stops on exchanges that support them (even more stable - even work after Gekko goes down / connection problems).

First version is currently undergoing more testing, see here: #2429.

So instead of increasing complexity in userland (strategies) I rather have Gekko provide more features so that people can create complex behaviour without polluting their strategies (which should be focused on alpha, not execution nits).

> and other complex strategies we need different candle sizes.

I do understand that more complex strategies won't have enough on a single timeframe. I'll keep this PR open for a little bit to see if others like the implementation.

plugins/tradingAdvisor/tradingAdvisor.js
```diff
@@ -86,6 +86,9 @@ Actor.prototype.setupStrategy = function() {
 // HANDLERS
 // process the 1m candles
 Actor.prototype.processCandle = function(candle, done) {
+  // notify strategy about one min
+  this.strategy.updateOneMin(candle);
```
askmike, line 90: Though maybe do this bellow line 92? The docs do state that this.candle reflects the current candle. In your proposal it reflects the old one (since you run a strat function before setting it). Also we should probably document/think about how we want the order of functions to be called.

If a strategy has async indicators Gekko should defer all other 1min candles until after they are calculated, but the buffering of additional 1min candles while the async indicators are calculated happens below this point. Meaning that (depending on how the node stream write candles) this might trigger `updateOneMin` functions of candles of candles that happen after a bigger candle might arrive first.

For example: If you you have a strategy that:

- uses a tulip indicator
- runs over 10min candles
- has a  `updateOneMin` function
- is running in a backtest (or realtime markets with burst of candles, see #2461)
- Candles are from `:00`, `:10`, `:20`, etc.

I am not sure if right now you are guarenteed to get your 10min candle from 10:10 before any 1 min candles of 10:11 fire (since the big candle might be buffering below this code, while the `10:11` is already coming in).  Whether this matters depends on your strategy.

askmike (commented):
see ^.

plugins/tradingAdvisor/tradingAdvisor.js
askmike: I am not 100% sure of the above, Gekko uses [Node streams](https://nodejs.org/api/stream.html) to pump candles into plugins (they end up in  plugins/tradingAdvisor/tradingAdvisor.js:88). And with a stream you pass a callback once you are done with consuming (a candle), the callback is only called after Gekko [is done with calculating all the async stuff](https://github.com/askmike/gekko/blob/b99c5a2f9ad766b203608b006f99a9f7e22713f0/plugins/tradingAdvisor/tradingAdvisor.js#L105). However the streams API has changed and I am currently not sure how much buffering of candles is guarenteed to happen before you call this callback function. As such Gekko implements another layer of buffering here: https://github.com/askmike/gekko/blob/develop/plugins/tradingAdvisor/asyncIndicatorRunner.js#L28-L36

eusorov: As I understood the code we have a flow: 
market stream -> 1min candle -> batcher buffer -> emit big candle -> cal. indicators sync/async -> strategy.update -> done() get new 1 min candle from the stream.

than we can put updateOneMin just before the buffer
market stream -> 1min candle -> strategy.updateOneMin -> batcher buffer -> emit big candle -> cal. indicators sync/async -> strategy.update -> done() get new candle from the stream.

The strategy gets always first 1 min candle and than we get the buffered big candle. And when buffering takes very long, than the programm still waits for the done() and doesn't get new candle. 

As long the updateOneMin has sync code there should be no problem. If you start async code in  updateOneMin with callbacks than yes we might get into trouble. But now the update-method also expects only sync code. 

And yes putting   this.strategy.updateOneMin(candle); after line 92 is good.

"I am not sure if right now you are guarenteed to get your 10min candle from 10:10 before any 1 min candles of 10:11 fire"

I would say yes: we get 1-1-1-1-1 - 1-1-1-1-1-10min because of the done().

askmike: > The strategy gets always first 1 min candle and than we get the buffered big candle. And when buffering takes very long, than the programm still waits for the done() and doesn't get new candle.

To make sure we are on the same line, I am not talking about any buffering from the candleBatcher. That's a sync buffer and has no effect here.

> I would say yes: we get 1-1-1-1-1 - 1-1-1-1-1-10min because of the done().

The buffer I was talking about: (the one after the candleBatcher emits a big candle, but before gekko runs lifecycle methods of strategy) the one inside the asyncIndicatorRunner. It's there because of the many issues regarding race conditions from async indicators (issues reported by users before that file existed).

> As long the updateOneMin has sync code there should be no problem. If you start async code in updateOneMin with callbacks than yes we might get into trouble. But now the update-method also expects only sync code.

This is not what I mean, we can easily document that users should not do this (the same as other lifecycle methods).

I am specifically talking about the about how many candles will ever pass through the stream (meaning hitting the processCandle of the tradingAdvisor without the previous one having called done). I agree that no writing stream should ever push more candles when the tradingAdvisor isn't done processing yet. But I am not sure if that's how all writing streams (called markets, 3 inside this repo and some more floating around the internet) are actually implemented. Or if they are not maybe the node streams API gives some guarantees?

Gekko had so many problems as a result of running async things related to strategies (especially in backtests, but also in realtime markets) that I only accept code for which we are sure of the behaviour.

mark-sch:
Great to find a discussion about this feature. I did a very similar extension to the tradingAdvisor today to get the one minute candles into the strategy also.  So I would really look forward to have this feature in the main gekko development repo -  with Mike's best practice advice and experience.
In addition to the one minute candles I have remote candles (from tcp/ip socket server)  in mind and remote advices. My goal is to watch candle trends from other markets for use in gekko trading strategies. E.g. the bitcoin pump and dump often happens a couple of minutes before the altcoins follow this trend. Is there any implementation approach for this already? To avoid changing the gekko one market concept an extention (with empty default implementations) of the tradingAdvisor.js seems most reasonable so far for me. @Mike Would you suggest a different approach?

plugins/tradingAdvisor/tradingAdvisor.js
eusorov: do you mean this 4 Readables? 

budfox (no get function) only feeds the db,
backtest (get function retrieves data from db),
leech (get function retrieves data from db),
importer (get function retrieves data from exchange and feeds db)

Only leech and backtest are used with strategies and they build very similar. I don't see any obstacles. 
In baseTradingMethod you call afterAsync() explicitly after asyncIndicatorRunner.processCandle() is finished, so that after propogateTick->update(candle) -> return done() line 127 is called. There is no way how  asyncIndicatorRunner can interfere with the gekko stream. 

Or do I miss something?

askmike: > In baseTradingMethod you call afterAsync() explicitly after asyncIndicatorRunner.processCandle() is finished, so that after propogateTick->update(candle) -> return done() line 127 is called. 

True, what I mean is: The baseTradingMethod will run asyncIndicators, and only after they are run the callback which came from the tradingAdvisor.processCandle. This is all great, however are we sure the stream will halt pumping new candles until this callback is ran? I am not sure the stream behaves like this in the realtime market. But I might be wrong, I just want to know for a fact given the amount of historical bugs regarding async candles.

eusorov: As far I understand, when we start realtime/backtesting we always start the leecher to read data from db: 

1. market data ( leecher readable Stream) pipes db data to 
2. gekkoStream (Writeable Stream) calls every plugin with processCandles (), and it will halt if done() function is not called.

if updateOneMin(candle) is sync and takes very long, than gekkoStream will wait and it could lead to : backpressure https://nodejs.org/en/docs/guides/backpressuring-in-streams/ 

"There are instances where a Readable stream might give data to the Writable much too quickly — much more than the consumer can handle!

When that occurs, the consumer will begin to queue all the chunks of data for later consumption. The write queue will get longer and longer, and because of this more data must be kept in memory until the entire process has completed."

There are solution to this problem which are described below in that article. 

The question is this the only concern?

Another question is what about async code in updateOneMin(candle) .  Because we might want to have other candles and we want to have asyncIndicators with this candles. Than we need to provide done callback like updateOneMin(candle, done)

mark-sch: > Another question is what about async code in updateOneMin(candle)

The one minute candle works nice until you are going to use async code inside. I am also using this oneMin candles with async talib+tulip indicators, but in this case we need either a callback param or leave the current params and use the ECMAScript await/async/promise pattern.

Which flavor you take is a personal preference, but none of them gives you the ability to use the asyncIndicatorRunner out of the box with this custom/different candle size, or did I miss something?. How to tell this.addTulipIndicator that we want to use it with a different candle size? Therefor I added/suggested  async Tulip and Talib wrappers, see PR, to use these async indicators like any other native Gekko indicator (I am using it in my one minute candles side by side to other candle sizes)

Having this in mind it showed, that it will become a pretty nested callback flow in the end, so I personally went the await/async way. All I can say, it works pretty nice for me.

https://github.com/askmike/gekko/pull/2530/commits/affbb47de0db8b4f86d63abf7d21649d182fa65f

See line 102 for the one minute candles

PR #2500, opened 2018-09-03
[BUGFIX] Fixes IPC disconnected error

VegtMedia-Alex:
Fixes the Error: IPC channel is already disconnected which is happening when we are trying to disconnect an already disconnected process

* **Other information**:

### DIFF
--- a/core/workers/loadCandles/parent.js
+++ b/core/workers/loadCandles/parent.js
@@ -56,7 +56,9 @@ module.exports = (config, callback) => {
 
     // else we are done and have candles!
     done(null, m);
-    this.disconnect();
+    if (this.connected) {
+      this.disconnect();
+    }
   });
 
   child.on('exit', code => {


VegtMedia-Alex:
After more tests, this is not a full fix. 

More information about error:

screen shot 2018-09-03 at 23 38 21

This only happens when the redisBeacon is on! For now I disabled redis and then its working fine.

askmike:
ping @eusorov 

> This only happens when the redisBeacon is on! For now I disabled redis and then its working fine.

Gekko with or without this fix?

VegtMedia-Alex:
It happens without the fix and redis on very fast. With the fix and redis on it still happens but it takes longer for the error to appear.

askmike:
Looking at the stack trace you shared this error gets thrown async after the `disconnect` call. Else we could simply wrap the `disconnect` in a try/catch. Seems we are not the only ones with this problem, I could find this less than ideal solution: https://stackoverflow.com/a/18753988/843033

> With the fix and redis on it still happens but it takes longer for the error to appear.

Sounds like a race condition, it might be that during the if check the client is connected but by the time it actually tries to disconnect it isn't anymore.

VegtMedia-Alex:
Nice work! I think if its a race condition it should fail silently and not make it crash the whole UI.

askmike:
agreed, but is pretty much nodejs acting a bit strange. I suggest we merge this since it already improved the behaviour. And after that: we could do some testing to see if the child process fires more messages (in which case gekko would try to disconnect them multiple times). However I can't reproduce this as is..

PR #2501, opened 2018-09-03
Added ETC support for GDAX

crypto49er:

feature

GDAX wrapper doesn't support ETC even though GDAX added it last month.

Gekko will be able to trade ETC with other currencies.

* **Other information**:


askmike:
:+1:

Issue #2503, opened 2018-09-04
Postgresql: uncaughtException TypeError: Cannot read property 'query' of null


VegtMedia-Alex:
Running Gekko v0.6.6 and Gekko UI v0.2.3 inside docker with the postgresql adapter. When starting multiple market watches a few are crashing after some time.

Error below:

<img width="1066" alt="screen shot 2018-09-04 at 09 25 14" src="https://user-images.githubusercontent.com/3672692/45016392-c7ab3e80-b024-11e8-84e9-c413af057219.png">

PR #2504, opened 2018-09-04
[BUGFIX] Fixes empty client on postgresql plugin

VegtMedia-Alex:
@askmike probably will need some more attention, because I am not able to find out why the client is empty?

https://github.com/askmike/gekko/issues/2503

* **Other information**:

### DIFF
--- a/plugins/postgresql/writer.js
+++ b/plugins/postgresql/writer.js
@@ -33,14 +33,16 @@ Store.prototype.writeCandles = function() {
     `;
     
     this.db.connect((err,client,done) => {
-      client.query(stmt, (err, res) => {
-        done();
-        if (err) {
-          log.debug(err.stack)
-        } else {
-          //log.debug(res)
-        }
-      });
+      if(client) {
+        client.query(stmt, (err, res) => {
+          done();
+          if (err) {
+            log.debug(err.stack)
+          } else {
+            //log.debug(res)
+          }
+        });
+      }
     });
   });
 


askmike:
Yea this is not ideal: if client is undefined Gekko won't run the query and never run the next callback (basically freeze).

Instead of doing this, we should probably check the first `err` parameter (right now it seems ignored). Would you mind doing that instead? You can use `util.die` (import util first) to crash Gekko with the error. I guess it would be cleaner to actually wrap the pool exposed by `handle.js` in a function that checks the error and passes only proper connections on, but you can do either.

VegtMedia-Alex:
Hi @askmike my node knowledge is basic. I am still learning, i will see what i can do.

askmike:
@alexvandervegt no problem! If the client is not defined it means that postgresql encountered an error. Gekko should check that like so:

https://github.com/askmike/gekko/blob/daa4eb6dacb64b3d9571c4dff04e9af07606bcc1/plugins/postgresql/handle.js#L54-L56

(if there is a DB error Gekko can crash, since there isn't much Gekko can do if the DB is down).

askmike:
Let me know if you rather have me fix it, but I'll keep this open so you can finish up the PR :)

VegtMedia-Alex:
@askmike I think best would be if you fix this one. I have looked over and over but I am not able to see it yet :)

askmike:
fixed here: fe8dbcd

askmike:
If you run it again now Gekko will spit out why client is undefined (there was a postgres error along the way).

ezgoProject:
@alexvandervegt how did you change the database from sqlite to postgresql

PR #2529, opened 2018-09-14
Updated API Endpoint, frequency of update msg

crypto49er:

Feature

https://github.com/askmike/gekko/issues/2517

In addition, the message 'Scanning back in the history needed... false' shows up in console after every candle, even after the all the historical data has been scanned in.

It uses the new API URL and the message is logged to the console only if 'since' is valid.

* **Other information**:

Changed API to https://api.pro.coinbase.com and sandbox API to https://api-public.sandbox.pro.coinbase.com. Modified the else statement history scanning process so the message 'Scanning back in the history needed...' only shows if 'since' is valid. This corrects the issue where the message is log to console at every candle even after scanning process is complete.


askmike:
great, thanks!

PR #2544, opened 2018-09-18
Added async Tulip indicator

mark-sch:

Feature

Right now any native Gekko indicator need to run synchronous, impossible to use async execution or callbacks. Since talib und tulip indicators are async, so they need to use the asyncIndicatorRunner, where Gekko is calculating it behind the scenes before the strategy update is called.

The existing approach works nice until you need async behavior inside the strategy on your own. E.g. this is the case if you write multi-timeframe strategies with custom candle batching. With this async tulip indicator implementation, it is possible to add and calculate tulip indicators like any other native gekko indicator, from inside the strategy.

* **Other information**:

### DIFF
--- a/strategies/indicators/TulipAsync.js
+++ b/strategies/indicators/TulipAsync.js
@@ -1,0 +1,92 @@
+// ****************************************************************************
+// *** TulipSync.js                                                         ***
+// ****************************************************************************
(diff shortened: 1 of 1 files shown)


askmike:
let's discuss first in here: https://github.com/askmike/gekko/pull/2530#issuecomment-422379725

askmike:
see ^

PR #2545, opened 2018-09-18
Added ForecastV3 strategy to showcase async tulip+talib usage

mark-sch:

Feature

Strategies use this.addTalibIndicator inside the strategy init function to use any Talib or Tulip indicator right now. This is a different approach than using native gekko indicators.  

In multi-timeframe strategies for example it is necessary to apply a talib/tulip indicator against a custom candle size. In this case the asynIndicatorRunner (used above) is not available.

This is an advanced sample strategy using async talib/tulip implementations to showcase strategies using async/await. This approach can be used for many other, async things inside a strategy, e.g. requesting a twitter feed from the internet and add it to your strategy decisions, or remote markets/candles, or any other async functionality you can imagine for use inside a gekko strategy.

* **Other information**:
The above Forecast strategy was heavily discussed on gekko discord channel, good backtest result but impossible to run live due to async problems. The conversion of this strategy to async tulip/talib is also a new test case whether this strategy can be enabled for live trading (in a multi-timeframe strat).

### DIFF
--- a/strategies/ForecastV3.js
+++ b/strategies/ForecastV3.js
@@ -1,0 +1,139 @@
+var _ = require('lodash');
+var log = require('../core/log.js');
+const util = require('../core/util');
+const TULIPASYNC = require('../strategies/indicators/TulipAsync.js');
+const TALIBASYNC = require('../strategies/indicators/TalibAsync.js');
+
+
+function getScale(val) {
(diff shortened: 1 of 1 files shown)


askmike:
see #2530.

PR #2549, opened 2018-09-19
Pushbullet Plugin - extended functionality

GryphonR:
Feature - Plugin Expansion

Pushbullet Messages on new advice.

Pushbullet optionally messages on new advice and also optionally messages on TradeComplete event with additional information of:
Advice Price
Trade Price
Slip %
Advice Time
Trade Completion Time
Time Taken to fill order

Pushbullet message on startup now also states whether it is a Live or Paper trader.

* **Other information**:
I extended this for my own use as I wanted more details on the trades that gekko was making and have found it very helpful. 

Hopefully helpful to others too if it's wanted?

My first PR so I think I've done everything right - happy to change anything that's needed though!

### DIFF
--- a/plugins/pushbullet.js
+++ b/plugins/pushbullet.js
@@ -1,88 +1,188 @@
 /**
  * Created by rocketman1337345 on 8/14/16.
+ * Extended by RJPGriffin (Gryphon) on 30/5/18
  */
+
+
+/**
+Required Config:
+
+config.pushbullet = {
+  // sends pushbullets if true
+  enabled: true,
+  // Send 'Gekko starting' message if true
+  sendMessageOnStart: true,
+  // Send Message for advice?
+  sendOnAdvice: true,
+  // Send Message on Trade Completion?
+  sendOnTrade: true,
+  // disable advice printout if it's soft
+  muteSoft: true,
+  // your pushbullet API key
+  key: '',
+  // your email
+  email: 'jon_snow@westeros.com',
+  // Messages will start with this tag
+  tag: '[GEKKO]'
+};
+
+
+ **/
 
 var pushbullet = require("pushbullet");
 var _ = require('lodash');
+const moment = require('moment');
 var log = require('../core/log.js');
 var util = require('../core/util.js');
 var config = util.getConfig();
 var pushbulletConfig = config.pushbullet;
 
 var Pushbullet = function(done) {
-    _.bindAll(this);
+  _.bindAll(this);
 
-    this.pusher;
-    this.price = 'N/A';
+  this.pusher;
+  this.price = 'N/A';
 
-    this.done = done;
-    this.setup();
+  this.advicePrice = 0;
+  this.adviceTime = moment();
+
+  this.done = done;
+  this.setup();
 };
 
-Pushbullet.prototype.setup = function(done){
+Pushbullet.prototype.setup = function(done) {
 
-    var setupPushBullet = function (err, result) {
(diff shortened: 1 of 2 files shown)


askmike:
This is great!

For ages I wanted to create something so that messaging plugins in Gekko rely more useful information. Basically what this PR does!

> My first PR so I think I've done everything right - happy to change anything that's needed though!

Looks great, I'll test this out asap and merge it if I don't run into any issues.

GryphonR:
Two things I've realised:
* This will  crash for anyone who uses it with the old pushbullet config. Should I remove the extra config options to make it backwards compatible?
* I haven't included any buy/sell amounts as I don't use pushbullet for that, but realised it would probably be useful. It does start becoming a lot of information in a message, but can add them if people think it would be useful?

askmike:
Sorry for the delay, it works great!

> This will crash for anyone who uses it with the old pushbullet config. Should I remove the extra config options to make it backwards compatible?

No: while it's good to guard this. But instead I want to move these kind of instructions to the changelog when this PR hits a new Gekko release.

> I haven't included any buy/sell amounts as I don't use pushbullet for that, but realised it would probably be useful. It does start becoming a lot of information in a message, but can add them if people think it would be useful?

I have very little visibility in what people use pushbullet for, my personal opinion is:

screen shot 2018-09-27 at 16 20 11

Amount can definitely be added. I even think the cost (in percentage for example) can be interesting as well.

Going in depth: The advice price is the last of the candle on which Gekko adviced. This is not perse the market price when Gekko triggered it (the last trade price of the last previously closed candle - in Gekko this means there was at least one more trade after this, maybe at a different price - but this is hidden from your strategy). As such calling the difference between this price and execution price "slip" can be confusing, since "slippage" is considered to be something else (usually relates to market orders, which Gekko doesn't do).

-----

That said I am not that worried about it. So I'm merging this as is since it's already a big improvement. Feel free to change the body in a follow up PR!

askmike:
A quick follow up:

The paper trader (that does all simulations) always trades at the last candle price. So what you are currently logging isn't really important since the advice price will always be the trade price, and the advice time will always be the trade time (and the slip will always be 0%). But there were costs (whatever you configured as spread and slippage).

GryphonR:
Thanks Mike.  I've been meaning to have a look at the cost provided in the trade completed event. I've always felt that I want to know the 'slip' from advice price separate to the fees. Is the slippage included in this value the difference between the first placed bid/ask and the final  averaged buy/sell price?

I know that what I have called slip isn't slip in the conventional sense, but the difference between how the strategy would perform on paper to how it actually performs live. I'll see if I can find a better name for it.

> The paper trader (that does all simulations) always trades at the last candle price. So what you are currently logging isn't really important since the advice price will always be the trade price, and the advice time will always be the trade time (and the slip will always be 0%). But there were costs (whatever you configured as spread and slippage).

Agreed - some of the data is totally redundant. I'll look to change the messages for live/paper in a future PR.

askmike:
> Is the slippage included in this value the difference between the first placed bid/ask and the final averaged buy/sell price?

Unfortunately that information is currently not exposed, I see how this would be interesting.

> I know that what I have called slip isn't slip in the conventional sense, but the difference between how the strategy would perform on paper to how it actually performs live. I'll see if I can find a better name for it.

We probably need to come up with some "definitions" on things, and put them in some documentation. That way we can clearly define what we mean when we say things. Not just for this plugin but for everything coming out of Gekko.

PR #2564, opened 2018-09-27
Added descriptive comments for tulip indicators

botsedge:
Added installation procedures for a Mac that eases Newbie installations.
Also added descriptive comments/links to each of the tulip indicators in the README

It only changes description documents. It does not change the program itself

Added information especially for Newbies not familiar with CLI installations.
Also added descriptive comments/links to each of the tulip indicators in the README

* **Other information**:


docs/installation/installing_gekko_on_mac.md
```diff
@@ -0,0 +1,78 @@
...
+The recommended way of downloading Gekko is by using git. This makes keeping Gekko up to date a lot easier. Run this in a terminal:
+
+    git clone git://github.com/askmike/gekko.git -b stable
+    cd gekko
+
+This will download the latest stable version of Gekko, remove the final `-b stable` part to download the current latest release (which might not be as stable).
+
+Alternatively you can manually download the latest stable version of Gekko on the [releases page](https://github.com/askmike/gekko/releases).
+
+## Set file permissions for the Gekko folder and all subsequent folders
```
askmike: I'm actually developing Gekko on Mac and have never needed to do this. I don't think we should be recommending people to chmod 777 anything. What problem did this solve for you?

docs/installation/installing_gekko_on_mac.md
```diff
@@ -0,0 +1,78 @@
...
+
+Once you have Gekko downloaded you need to navigate to the gekko folder and make sure that file permissions are set correctly for the Gekko tree. You can do this via terminal:
+
+    sudo chmod -R 777 ~/gekko
+
+This will take care of the gekko folder and all subsequent folders.
+
+## Set Energy Saver Preferences on your Mac
```
askmike: I'm quite sure this is only a problem for macbooks (not iMacs) that are not plugged in? I'm developing on a macdbook and haven't changed these settings (I think - not sure). Gekko runs great for me.

askmike (changes requested):
As for the mac tutorial:

Gekko started with a single installation guide. But then there were too many problems on windows (people wanted to use git shell and bash and such) so a Windows guide was added. Later a similar situation happened with rasp pi.

But I'm not sure if this is a good way to go. We should probably have one installation guide. As for Mac: the default guide has never failed me before.

If people have specific issues regarding permissions or power management, maybe we should instead create a new page with "Common installation issues" (with sections per OS or something)? What do you think of that?

I do love your strategy/indicator documentation. Happy to merge that in asap.

docs/installation/installing_gekko_on_mac.md
botsedge: I think it is a great idea to add a section with "Common installation issues" instead of the added Mac version.
But i did that because i actually installed Gekko on several Mac´s, a couple of Macbooks an Imac and a Mac Mini. These were all consumer macs, so they were not changed in any way by somebody who is a developer. 

On all these mac`s the installation procedure failed after git cloning the repo to the computer because there were no write permissions set in the Gekko folder or any file within.

So obviously trying to npm install the dependecies afterwards failed on all machines. It took me a while to find the problem, since i was first looking into the program and reinstalled nodejs and things like that. 

So if you want to add a section with OS specific issues instead i think it is a great idea.

docs/installation/installing_gekko_on_mac.md
botsedge: I had this problem with a macbook that was plugged in but had the power saving set to a couple hours. So as soon as the OS went into hibernating, Gekko crashed.

It is a small problem but could save newbies tons of time, if we put this info somewhere into the installation procedure

docs/installation/installing_gekko_on_mac.md
askmike: > On all these mac`s the installation procedure failed after git cloning the repo to the computer because there were no write permissions set in the Gekko folder or any file within.

I would be interested to know how git was installed and ran (did you use sudo or not)?

> So if you want to add a section with OS specific issues instead i think it is a great idea.

Yes let's do this instead! I can draft this but you are also more than welcome to do this: create a new page called "Common_installation_issues.md" and add one section for MacOS.

botsedge: For your question no i did not use sudo i used the command in the mac terminal like it said on the installation page.

And for the new page, i will go ahead and add a new page with the mac os issues you can then add linux or windows or server issues. I will setup the next Gekko on a VPS so let´s see what i find out about that procedure.

askmike: I am interested in how you ever got permission problems, I would want to figure that part out first. The problem with `chmod 777` is that it means everyone on that computer can now read the folder, and that folder will store your API keys (which access all your money). So I really don't want tell people to do this from a security perspective. Chmod 700 should work unless your user is not the owner (which, if you didn't use sudo is very strange).

botsedge: Interesting, i am not a Mac OS expert but this is how i encountered the problem. I was logged in as a normal user to those macs i installed Gekko on. As a user i then followed exactly the procedure in your installation guide. Installing nodejs worked as advertised (although i later reinstalled nodejs several times because i thought the nodejs version has something to do with my problem).
Then i git cloned the repo onto the macs exactly using the command in your installation procedure without using sudo and being logged in as a normal user on those macs.

I then had the Gekko folder on the mac and all files in it but installing the dependencies failed. Investigating this i first thought it might be the nodejs version then i thought i made mistakes when copying the commands from the installation procedure but that was not the problem. It took me forever to think about the permissions and when i finally checked those they were set to read only for the Gekko folder and all files in it.

At that point i had the solution at least for me being the normal user on those macs. I manually set all permissions to the Gekko folder and the files in it to write and read and voila now installing the dependencies worked and i could start using Gekko. 

Since i am the only guy using my macs i don´t have a problem with setting the write permissions globally but i see your point in not allowing this for somebody who is sharing the mac with others.

How would you like to proceed?

askmike: > I was logged in as a normal user to those macs i installed Gekko on. As a user i then followed exactly the procedure in your installation guide. Installing nodejs worked as advertised (although i later reinstalled nodejs several times because i thought the nodejs version has something to do with my problem).
> Then i git cloned the repo onto the macs exactly using the command in your installation procedure without using sudo and being logged in as a normal user on those macs.

Very strange, it should not be set to read only.

--

> How would you like to proceed?

- For the mac chmod issue: I really don't want to recommend anyone to chmod 777, exactly when I don't understand what is going wrong. It's probably better to open a new issue where we can go into your exactly installation steps (I'm specifically interested in how you installed git, and into what directory you are downloading Gekko). Once we understand why the gekko folder is read only we can prevent this in a more secure way.
- For all your indicator documentation: I would love to merge this asap.
- For the mac guide: can you put the mac specific stuff in another file?

It would probably be cleanest, quickest to now remove the Mac installation guide so I can just merge the indicator stuff. While we create an issue for the chmod stuff and a new PR for the "Common installation issues" page. What do you think?

botsedge:
Deleted the installation procedure on a mac for now will open up a new issue

askmike:
Apologies for the late merge. Thanks a lot for splitting it up :)

botsedge:
No Problem Mike, glad that i could help.

Issue #2563, opened 2018-09-27
Backtest import in GUI doesn't import to the correct date. Loader starts in negative.


rkingy:
**I'm submitting a ...**
[ X] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Went to import from binance, set my date range clicked import.
Progress started from negative. And never imported to the date (current time) as requested.

**Expected result** (what you hoped would happen)
Import imports to when I say

**Actual result** (unexpected outcome)
Doesn't import the correct amount of data.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

I discovered that in core/markets/import.js the dates need to be converted to UTC. This fixed the issue for me.. the old version had .utc() around the to and from dates. Not sure why this was removed?

askmike:
> never imported to the date (current time) as requested.

All times are supposed to be in UTC, so if you are not in a timezone equal to UTC the time you fill in there is not what Gekko will import.

> Progress started from negative.

Yes it's a timezone issue. I think the bug was introduced here: https://github.com/askmike/gekko/pull/2263#discussion_r199354415 @mark-sch maybe you can weigh in? EDIT: That only seems to impact kraken importing, while this bug is preset everywhere. Might not be that PR - but I'm sure something changed along the lines where we dropped an UTC conversion somewhere.

rkingy:
It looks like the .utc() conversions were removed in commit #1ce10f455ebab261e3ab43a121b2058236ff11aa 

I'll push up a fix.

askmike:
this is fixed in latest develop.

PR #2566, opened 2018-09-28
Fix to importer dates

rkingy:

Fix date issues in importer.
https://github.com/askmike/gekko/issues/2563#issuecomment-425038916 

Dates from the GUI are wrong.. 

Everything works as expected again in UTC format.

* **Other information**:

### DIFF
--- a/core/markets/importer.js
+++ b/core/markets/importer.js
@@ -9,12 +9,12 @@ var gekkoEnv = util.gekkoEnv();
 var adapter = config[config.adapter];
 var daterange = config.importer.daterange;
 
-var from = moment(daterange.from);
+var from = moment.utc(daterange.from);
 
 if(daterange.to) {
-  var to = moment(daterange.to);
+  var to = moment.utc(daterange.to);
 } else {
-  var to = moment();
+  var to = moment().utc();
   log.debug(
     'No end date specified for importing, setting to',
     to.format()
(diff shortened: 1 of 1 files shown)


mark-sch:
It is some time ago, but I remember similar import issues when dealing with PostgreSQL db. To get it working there had been some moment and utc changes necessary. Which db are you using? Maybe it needs deeper testing with different db environments...

rkingy:
Hi @mark-sch  I'm just using the standard install. I believe it uses "sqlite3": "^4.0.0",
This has only been an issue I've noticed since upgrading a fresh install to the latest 0.6.6 version. I had been using the older version for quite a long time.

Your probably right if people are using different dbs that handle dates differently. But for those that clone the repo and run for the first time it also needs to work out of the box.

Without the .utc treatment this is what the GUI looks like when doing a import. Starts off in negative % bar and then never imports up to the date you want.

screenshot 2018-09-29 11 43 37
screenshot 2018-09-29 11 43 42

askmike:
@rkingy @mark-sch thanks for providing a fix and context. I will test this with both sqlite and postgres asap.

askmike:
this PR fixes the importer using both databases for me. I am currently in UTC+8 and I definitely noticed the offset, I think always. With this PR the problem goes away for me.

@mark-sch if you encounter problems after this PR please let me know!

Issue #2568, opened 2018-09-28
some browsers not showing gekko UI (IE 10, OPERA mini Android)


eusorov:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[x] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
start latest version of gekko ui. Open with browser, IE10 or Opera mini on my android.  I didnt try with latest IE. 

**Expected result** (what you hoped would happen)
the page with gekko ui is loaded.

**Actual result** (unexpected outcome)
error in javascript, blank page.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

There is a problem with file: state/reduceState.js
It is excluded in babel.config.js, if it is included, than there is a problem with webpack which can not handle ES6 module and commonjs module. This file is a commonjs module and it will be imported in server: 
gekkoManager.js with require () 
At the same time this file is included in client
mutations.js also with require ()

Now some browsers can handle this and some not.
I think to exclude the file from babel transpiler seems also like a hack.

askmike:
> I think to exclude the file from babel transpiler seems also like a hack.

Yes this isn't ideal. Note I'm building a new UI. Though still not sure how to work around this:

I've been struggling with how to share javascript code between main gekko (nodejs) and the frontend (babel vue project). In the past I've always used browserify to bundle up (commonjs) nodejs code so it can be used in the frontend. However the babel used my my frontend framework doesn't use commonjs but es 6 export syntax (they are not compatible). While developing the current method I tried to figure out the proper way:

https://stackoverflow.com/questions/51082287/problems-importing-a-shared-script-in-babel-node

But I didn't found many answers.

------

If anyone can help out to get a better solution? The problem is that I want to have 1 module (that handles gekko update events and merges them into an object) and use that in both the frontend and the backend.

eusorov:
there are conflicts with dist folder (index.html etc), I will make another branch without dist files.

askmike:
@eusorov awesome that you figured out a better solution! When you submit a PR, can you not fully compile the frontend in that PR? I'm trying to keep the git repo size down, and everytime we recompile the frontend that adds another MB to the repo size. I'm only recompiling now with new full releases.

PR #2573, opened 2018-09-30
bugfix #2568 transforms commonjs modules to ES2015

eusorov:
Bugfix #2568

see #2568

commonjs modules get transpiled with babel.

* **Other information**:

### DIFF
--- a/web/vue/babel.config.js
+++ b/web/vue/babel.config.js
@@ -4,6 +4,6 @@ module.exports = {
   ],
   ignore: [
     'node_modules',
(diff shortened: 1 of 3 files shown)


askmike:
This works great! Thanks a lot. I will update my stackoverflow question with your answer.

EDIT: @eusorov this works but I'm not sure why. As the plugin you added does:

> This plugin transforms ES2015 modules to CommonJS.

While the reduceState function is written in CommonJS. This PR worked in my tests, but I'm not understanding why.

Issue #2548, opened 2018-09-19
Method.Update Function Not called with multiple Tulip Indicators


bvanhou:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[X ] bug report
[ ] question about the decisions made in the repository

newest Stabe build

**Action taken** (what you did)
Attempted to add multiple tulip indicators (any tulip indicators). 
Then run strategy via command line.

**Expected result** (what you hoped would happen)
The method.update function will be called.

**Actual result** (unexpected outcome)
The method.update function is not called.

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

You can mix and match any of the tulip indicators. Declaring multiple of them just does not work.
```
this.addTulipIndicator('_mfi', 'mfi', {optInTimePeriod: 14});
  this.addTulipIndicator('_bb', 'bbands', {optInTimePeriod: 20, optInNbStdDevs:2});
  this.addTulipIndicator('_stoch', 'stoch', {optInFastKPeriod: 3, optInSlowKPeriod: 3, optInSlowDPeriod: 3});
  this.addTulipIndicator('_macd', 'macd', { optInFastPeriod: 10, optInSlowPeriod: 21, optInSignalPeriod: 9});
```

askmike:
Thanks for the report, will check asap.

bvanhou:
Thanks, this unfortunately is a big show stopper. I just tested and found that the issue occurs in v0.6.6 and not v0.6.5. Just verified.

#if this helps
config.tradingAdvisor = {
  enabled: true,
  method: 'tulip-multi-strat',
  candleSize: 1,
  historySize: 10,
}

**Note: Issue will always occur doesnt matter what you choose

eusorov:
I can confirm, in standalone mode nothing happens but with UI-Mode it works like expected.

Added the 4 tulip indicators to tulip-multi-strat.js from above and the update method is called as expected.

In general it is probably difficult to see logs in console, because log.js prevents in UI-mode. But if you add to _write method a console.log(message); like this: 
```javascript
Log.prototype = {
   _write: function(method, args, name) {
   ...
    console.log(message);
    this.output[method](message);
}
```
than you see that the update method is called.

eusorov:
[...]

PR #2577, opened 2018-10-03
bugfix #2548 if no strategy.historySize found, than get it from tradingAdvisor

eusorov:
Bugfix #2548

When using tulip indicators the update function is not called. see  #2548
this behavior is only in standalone mode, because no historySize is in sample-config is provided. That's why isAllowedToCheck is always undefined and this.completedWarmup is always false.

Now checking after init() method if historySize is provided. If no than take the historySize from tradingAdvisor.

* **Other information**:
To  provide extra historySize in settings is not very convinient, it is better to have just one place in tradingAdvisor.historySize but with the option to override in settings.historySize.

### DIFF
--- a/plugins/tradingAdvisor/baseTradingMethod.js
+++ b/plugins/tradingAdvisor/baseTradingMethod.js
@@ -37,7 +37,6 @@ var Base = function(settings) {
   this.settings = settings;
   this.tradingAdvisor = config.tradingAdvisor;
   // defaults
-  this.requiredHistory = 0;
   this.priceValue = 'close';
   this.indicators = {};
   this.asyncTick = false;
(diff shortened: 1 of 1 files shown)


mark-sch:
nice

askmike:
Great!

The difference between requiredHistory and historySize does not make sense anymore. I want to move towards only having 1 property, and it being defined in the strategy. But until that we have can use this. Thanks

plugins/tradingAdvisor/baseTradingMethod.js
```diff
@@ -69,6 +68,11 @@ var Base = function(settings) {
   // let's run the implemented starting point
   this.init();
 
+  //if no requiredHistory was provided, set default from tradingAdvisor
+  if (!this.requiredHistory){
```
askmike: Would you mind checking if it's actually undefined? This current check would also reset it if the strategy has defined it as `0`. A simple way would be to use lodash to check if requiredHistory is a number (`_.isNumber`).

askmike (changes requested):

askmike:
perfect, thanks!

PR #2578, opened 2018-10-04
Fixed getFee() for binance exchange

olexiyb:
Bug fix

binance.getFee() - 0.1
bitfinex.getFee() - 0.002

binance.getFee() - 0.001
bitfinex.getFee() - 0.002

* **Other information**:
Here is binance raw response. As you can see binance gives us fee as 10, when I would expect 0.1 
{ makerCommission: 10,
  takerCommission: 10,
  buyerCommission: 0,
  sellerCommission: 0,
  canTrade: true,
  canWithdraw: true,
  canDeposit: true,

So we need to divide by 10000 to get decimal fee for binance

### DIFF
--- a/core/pipeline.js
+++ b/core/pipeline.js
@@ -6,7 +6,7 @@
   all enabled plugins are actually supported by that market.
 
   Read more here:
-  https://gekko.wizbit/docs/internals/architecture.html
+  https://gekko.wizb.it/docs/internals/architecture.html
 
 */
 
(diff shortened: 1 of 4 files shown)


olexiyb:
@askmike BTW why gekko does not use?

https://www.npmjs.com/package/ccxt

this library looks like implements interface to all exchanges

askmike (approved):

askmike:
nice catch, thanks!

askmike:
> this library looks like implements interface to all exchanges

It's a huge library and doesn't play nice with Gekko Broker (Gekko's internal order execution library).

PR #2581, opened 2018-10-06
ExchangeChecker: adding missing config reference

abenmrad:
Bug fix

When assessing whether the market can provide full market history when the selected exchange has a `exchangeMaxHistoryAge` property (as of now, only coingi has that property), the `ExchangeChecker` refers to the gekko process `config`, whose reference was removed in 29b0dde584a3ef432f396b32cb93b673c4c5b. The gekko process will therefore crash.    
I haven't witnessed this behavior, as I uncovered the bug while performing a rebase from gekko into an other project. But I assume this is what would happen when someone wants to run an `importer` market from coingi.

The gekko process will assess correctly whether it can or cannot fetch full market history. Assumed, untested. But seems rather straightforward.

* **Other information**:

### DIFF
--- a/exchange/exchangeChecker.js
+++ b/exchange/exchangeChecker.js
@@ -2,6 +2,8 @@ const _ = require('lodash');
 const fs = require('fs');
 const moment = require('moment');
 const errors = require('./exchangeErrors');
+const util = require('../core/util');
+const config = util.getConfig();
 
 const Checker = function() {
   _.bindAll(this);


askmike:
Thanks for this PR!

The problem is that everything inside the `exchange` directory is also a standalone lib called Gekko Broker.

- Read more here: https://gekko.wizb.it/docs/gekko-broker/introduction.html
- It can be downloaded from npm: https://www.npmjs.com/package/gekko-broker

Meaning that importing things from outside the `exchange` directory will only work for Gekko, but nowhere else.

A better solution would be to refactor these lines:

https://github.com/askmike/gekko/blob/70cdf5705a631ba9f2f41092a0da00195586cfda/exchange/exchangeChecker.js#L65-L69

The context for that: Since Gekko 0.6 I have been pulling out all exchange related integration into Gekko Broker. This is now completed with the exception of:

- All the importers
- BudFox (module that outputs realtime candles from an exchange)

Until we move both of these into Gekko Broker I think it's better to move these checks into Gekko for now. What do you think?

askmike (changes requested):
see ^.

PR #2593, opened 2018-10-13
performance analyzer add  ratio roundtrips

julesGoullee:
Add in report ratio roundtrips: (roundtrips with pnl > 0) ÷ roundtrips


askmike:
Loving this!

PR #2597, opened 2018-10-16
Add support for ZRX on GDAK

g1tt:
Adds support for ZRX on GDAX - USD, ETH and BTC


g1tt:
Mike - my apologies! Still working out git here.
I have a different mod under test and realised that submitting this (ZRX support) would submit my other mods as well.
So whilst it's a bit messy the end result should be that just the ZRX mods are submitted and my other mods stay here!
Won't do it again, promise! :)

askmike:
No problem! Thanks for the fix.

askmike:
TODO for next time:

Gekko is able to build an object of all supported markets, it does this for other exchanges already (binance, kraken, coinfalcon). Here is an example: https://github.com/askmike/gekko/blob/develop/exchange/util/genMarketFiles/update-binance.js

Based on this endpoint we can do the same for coinbase: https://api.pro.coinbase.com/products

g1tt:
Thanks for the commit and the pointer. Leave it with me and I will sort out an update-coinbase.js file! :)

PR #2603, opened 2018-10-21
Auto update gdax/coinbase currencies

g1tt:
Auto-generated assets/currencies/pairs for GDAX/Coinbase

* **Other information**:
As per convo here - https://github.com/askmike/gekko/pull/2597.

I've tested a couple of imports and market watchers - it seems to work fine. I'd give it a good test though!! :)

Two issues -

1/ The importer shows an error as follows -

Some markets were unscannable, click here for details.
Unable to find datasets in the following markets:
- gdax:GBP:ZRX

I know this to be correct and when you import date/start a bot then it doesn't offer up GBP/ZRX as an option so not sure why this is, but it doesn't seem to affect functionality.

2/ The last issue is the gdax/coinbase identy crisis! I've left the file as gdax.js but the new .json is coinbase-marketdata.js. I guess this is the next thing to look at?


g1tt:
Oh and I forgot... also this updates the kraken-markets,json... but also this will need the updated kraken.js as I updated min qtys in there first...

g1tt:
I need to sort some bits out here where I seem to have working on an older version. Back shortly!

Issue #2616, opened 2018-10-28
Twitter plugin doesn't work - "Bad Authentication Data"


crypto49er:
**Note: this is the technical bug tracker, please use other platforms for getting support and starting a (non technical) discussion. See the [getting help page](https://gekko.wizb.it/docs/introduction/getting-help.html) for details.**

**I'm submitting a ...**
[x] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Enable Twitter plugin in the config file, filled in my consumer API key, API secret, Access Token and Access Token Secret

**Expected result** (what you hoped would happen)
Tweet the message "I've started watching exchange Fiat/Crypto I'll let you know..."

**Actual result** (unexpected outcome)
Got this error message no tweet: Pushbullet ERROR: [ { code: 215, message: 'Bad Authentication data.' } ]

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

The plugin seem to have been templated from the Pushbullet plugin. That's why it was generating the a Pushbullet error even though it wasn't from Pushbullet.

askmike:
fixed by yourself :)

PR #2617, opened 2018-10-28
Fixed Twitter Plugin

crypto49er:
Updated setup function to correctly reference the Twitter API keys in the config file.

Bug fix

Twitter plugin doesn't authenticate properly.

Twitter plugin authenticates and is able to tweet.

* **Other information**:

Fixes Issue #2616


askmike:
:+1: thanks

Issue #2628, opened 2018-11-04
An small issue in the documentation


douglascrp:
The sample code in the check-function documentation is wrong
https://gekko.wizb.it/docs/strategies/creating_a_strategy.html#check-function

There is a missing , in the sample, just after the 'long' parameter:

**I'm submitting a ...**
[x] bug report
[ ] question about the decisions made in the repository

**Action taken** (what you did)
Tried to create a custom strategy

**Expected result** (what you hoped would happen)
The code works

**Actual result** (unexpected outcome)
Got an error:
trigger: { // ignored when direction is not "long"
            ^^^^^^^
SyntaxError: Unexpected identifier

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc)

There is a missing , in the sample, just after the 'long' parameter:

```
this.advice({
  direction: 'long' // or short
  trigger: { // ignored when direction is not "long"
    type: 'trailingStop',
    trailPercentage: 5
    // or:
    // trailValue: 100
  }
});
```

should be:

```
this.advice({
  direction: 'long', // or short
  trigger: { // ignored when direction is not "long"
    type: 'trailingStop',
    trailPercentage: 5
    // or:
    // trailValue: 100
  }
});
```

askmike:
Nice catch! Would you mind submitting a PR for this?

bmarchesini:
I created a PR to solve the problem ;)

douglascrp:
@bmarchesini LOL
You were faster than me

I have just created a new one :grin:

askmike:
Ah! Well the important thing is that the issue is now fixed :)

@douglascrp thanks for the report and the PR!
@bmarchesini thanks for the PR!

PR #2629, opened 2018-11-04
Update creating_a_strategy.md

bmarchesini:
Resolve the issue #2628

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix #2628

* **What is the current behavior?** (You can also link to an open issue here)
Got an error:
trigger: { // ignored when direction is not "long"
SyntaxError: Unexpected identifier

* **What is the new behavior (if this is a feature change)?**
The code works

* **Other information**:

PR #2630, opened 2018-11-04
Included the missing , in the sample code for the check function

douglascrp:
docs update

As described https://github.com/askmike/gekko/issues/2628

This simple PR fixes the issue #2628 

* **Other information**:


askmike:
Thanks for the PR, but as noted this was also fixed in another PR in #2629. I do value your effort, it's unfortunate we can't merge this anymore.

PR #2657, opened 2018-11-19
Online strategy editor

mnofresno:
- **What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)**
feature: Strategy editor support so we can edit out strategies code directly from the gekko web UI
I could add a new navigation menu item to Home/LiveGekos/Backtest..../Strat Editor

- **What is the current behavior? (You can also link to an open issue here)**
no strat editor is included

- **What is the new behavior (if this is a feature change)?**
Online strategy editor for the gekko's web UI

- **Other information:**
Hi @askmike ! I want to ask you, what do you think if we could add a new feature to gekko web UI.


mnofresno:
It goes like this:

strat editor

askmike:
A few things:

- I'm working on a new UI, it's not open source yet but you can see it live here: https://app.gekkoplus.com/bot/Philson-Neuralnet - I don't think we should be putting new features in to the current one.
- In my current explanations and guides I push people to use a code editor and a command line, for example this video: https://youtu.be/6-74ZhrG0BE

Before we start implementing these kind of features, can we think about an ideal "strategy creation flow"? Like how do we want the "experience" to work?

mnofresno:
Thank you Mike, 

I see, but the problem I've been trying to cope developing this feature is not replaceable with a code editor. Because if you have the instance of gekko installed on a cloud server, and you only have web access to the UI, there is no way to launch an editor directly over the server's strategy js files.

The proposed workflow then will be:

1-You create a new JS strategy file in the online strat editor module
2-You back test it in the backtest module
3-Lately you can run this strat with a new gekko

Additionally I've been working in other possible improvements like:

1-Adding some text code highlighting to the editor
2-The ability to rename or copy a strategy
3-That the editor can also help in creating the .toml config files in the settings directory

Let me know what you think to continue working on this to be released pull-requested on the master or just leting it at this state just in my fork that is improved enough for my current personal use.

Thank you!

PR #2660, opened 2018-11-25
[DOCS] about_gekko.md: add consistent information for the risk of someone having access to your gekko

bmarchesini:
Add consistent information for the risk of someone having access to your gekko by establishing what can happen.

Docs update

None information about this subject

Add consistent information for the risk of someone having access to your gekko by establishing what can happen.

* **Other information**:

### DIFF
--- a/docs/introduction/about_gekko.md
+++ b/docs/introduction/about_gekko.md
@@ -47,6 +47,16 @@ Read more in the [architecture documentation](../internals/architecture.md).
 * The title is inspired by [Bateman](https://github.com/fearofcode/bateman).
 * This project is inspired by the [GoxTradingBot](https://github.com/virtimus/GoxTradingBot/) Chrome plugin (which in turn is inspired by [Goomboo's journal](https://bitcointalk.org/index.php?topic=60501.0)).
 
+## The risk of someone having access to your gekko:
+
+There are various risks associated with the appropriation of your Gekko account by someone else:
+
+- they can see your bots
+- they can start/stop bots
+- they can run many backtests that will basically DOS your server
+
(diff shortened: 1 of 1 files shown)


askmike:
I like this disclaimer, but the general about me isn't the right place. Because while the issuers are valid at that point people reading this have no idea what "someone having access to your gekko" means.

bmarchesini:
Can I put this in "Getting help"? It may seem like a better place to me

askmike:
I don't think that's a place people would go to when they go looking for this information. A better way would be to create a new document under installation that first discusses those risks and than talks about how you can prevent people from "having access to your gekko" (which depends on if you run Gekko on your own computer, or server, etc).

We can then link to this document in the installation guides (a place good place where people need to be reminded of these kind of risks).

What do you think of that?

bmarchesini:
We could call this section "Possible Risk" and in this document I would put the content of this PR and add a TODO for the prevention part until I write it. How are the docs folders and files processed to add them?

bmarchesini:
Up

askmike:
When it comes to automated trading "risk" always refers to your money at risk (measured one way or another). What about something like "Security" or "Security risk"?

bmarchesini:
Yes it seems to me a good idea, apart from creating the folder and.md file for the doc, is there another file to modify to refer this file and that it appears in the right place in the doc?

askmike:
You can simply create one new document inside the installation folder here: https://github.com/askmike/gekko/tree/develop/docs/installation

I think the installation files should be sufficient :)

askmike:
Will be superseded by #2672.

PR #2707, opened 2019-01-21
Code style

THETCR:
This PR fixes code-style issues, bugs and errors: 

Unterminated statements.
Attempts to assign to const variables.
Duplicate declarations.
Using 'var' instead of 'const' or 'let'.
Redundant local variables. 
Unnecessary semicolons.
Usage of variables that are not initialized.


askmike:
Huge PR! See comment

core/budfox/budfox.js
```diff
@@ -33,46 +33,46 @@ var BudFox = function(config) {
   // relay a marketUpdate event
   this.marketDataProvider.on(
     'marketUpdate',
-    e => this.emit('marketUpdate', e)
+    e => this.emit('marketUpdate', e),
```
askmike, line 36: What do you think this improves over the old style?

askmike (commented):
^

core/candleBatcher.js
```diff
@@ -6,22 +6,22 @@
...
 
-var _ = require('lodash');
-var util = require(__dirname + '/util');
+const _ = require('lodash');
+const util = require(__dirname + '/util');
 
-var CandleBatcher = function(candleSize) {
-  if(!_.isNumber(candleSize))
+const CandleBatcher = function(candleSize) {
+  if (!_.isNumber(candleSize))
```
askmike, line 17: this is not inline with any other if! All the ifs in the codebase have `if(` without a space!

core/budfox/budfox.js
THETCR: All the styling is conform one basic style guideline. If the code should have trailing comma's after the last statement is a different discussion.

I could add a PR to use ESLint. The project will be better of if all the code conforms to one style.

core/candleBatcher.js
THETCR: I see the problem, will change the styling for the if statements to be conform.

askmike: thanks!

core/budfox/budfox.js
askmike: I don't think the current project uses trailing commas anywhere else, would you be okay with conforming to a style with no trailing commas?

THETCR: I suggest using `eslint-config-airbnb-base,` `eslint-plugin-import`, `eslint-plugin-vue` and the recommended `ESLint` rules. 

This would enforce the code to be conform the `AirBnB` style guideline as follows:

```
Semicolon: Required
Trailing Commas: Required
Template strings: Prefer
Import Extensions: None
Space before function parentheses: None
Object Curly Spacing: Yes
Array Bracket Spacing: None
Underscored functions: None
Object Destructuring: Prefer
React Ordering: Opinionated
React Prop Validation: Required
Arrow Functions Return Assignment: No
Object Property Shorthand: Prefer
```

core/candleBatcher.js
THETCR: According to the style guidelines from `AirBnB` and `Google`, as well as the `common` guideline, there should be a space between the `if` and the `(`.

THETCR:
It seems to me that you don't really have the time to maintain this repository?

PR #2712, opened 2019-02-01
Fix install issue with sqlite3 in Raspbian 9

juanvenado84:
node-sqlite3 4.0.6 fixed a build break that manifested in Raspbian 9

Bug fix

Can't install gekko in Raspbian 9 because of a  build break with sqlite3 

Gekko installs successfully and sqlite3 compiles fine.

* **Other information**:


askmike:
:+1:

PR #2718, opened 2019-02-05
Added trade blotter plugin

crypto49er:

feature

There's no way to save the trades on a file.

A plugin called "Blotter" can be enabled to create a trade blotter, capturing all the buy/sell trades.

* **Other information**:

Some exchanges like Coinbase Pro list limit trades in multiple pieces, making it hard to understand when the trade took place. This trade blotter is better at tracking the buy/sell trades to determine if a strategy is working as intended.

### DIFF
--- a/plugins.js
+++ b/plugins.js
@@ -241,6 +241,17 @@ var plugins = [
       version: '0.18.0'
     }]
   },
+  {
+    name: 'Blotter',
+    description: 'Writes all buy/sell trades to a blotter CSV file',
+    slug: 'blotter',
+    async: false,
+    modes: ['realtime'],
+    dependencies: [{
+      module: 'fs',
+      version: '0.0.1-security'
+    }]
+  },
 ];
 
 module.exports = plugins;
--- a/plugins.js
+++ b/plugins.js
@@ -247,10 +247,6 @@ var plugins = [
     slug: 'blotter',
     async: false,
     modes: ['realtime'],
-    dependencies: [{
-      module: 'fs',
-      version: '0.0.1-security'
-    }]
   },
 ];
 
--- a/plugins/blotter.js
+++ b/plugins/blotter.js
@@ -1,0 +1,87 @@
+const fsw = require('fs');
+const _ = require('lodash');
+const log = require('../core/log.js');
+const util = require('../core/util.js');
+const config = util.getConfig();
+const blotterConfig = config.blotter;
+
+var Blotter = function(done) {
+  _.bindAll(this);
+
+  this.time;
+  this.valueAtBuy = 0.0;
+  this.filename = blotterConfig.filename;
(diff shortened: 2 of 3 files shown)


plugins.js
```diff
@@ -241,6 +241,17 @@ var plugins = [
...
     }]
   },
+  {
+    name: 'Blotter',
+    description: 'Writes all buy/sell trades to a blotter CSV file',
+    slug: 'blotter',
+    async: false,
+    modes: ['realtime'],
+    dependencies: [{
+      module: 'fs',
```
askmike: fs is a build in module, you don't need to specify it as a dependency here.

plugins/blotter.js
```diff
@@ -0,0 +1,87 @@
...
+  } else {
+    this.outtxt = this.outtxt  + "\n";
+  }
+
+  // If a trade date is from 1969 or 1970, there was an error with the trade
+  if (trade.date.format('YY') == '69' || trade.date.format('YY') == '70') {
+    log.error('Received 1969/1970 error, trade failed to execute, did not record in blotter');
+  }
+  else {
+    fsw.appendFileSync(this.filename, this.outtxt, encoding='utf8');
```
askmike: Would you mind not using sync fs functions? If for whatever reason this slows down Gekko is unable to monitor the plugin (or do anything else).

plugins/blotter.js
```diff
@@ -0,0 +1,87 @@
...
+  fsw.readFile(this.filename, (err, _) => {
+    if (err) {
+      log.warn('No file with the name', this.filename, 'found. Creating new blotter file');
+      fsw.appendFileSync(this.filename, this.headertxt, encoding='utf8'); 
+    } 
+  });
+
+};
+
+Blotter.prototype.processTradeCompleted = function(trade) {
```
askmike, line 42: Can't you rely on the roundtrip event? That way you don't have to calculate the pnl yourself. See here: https://gekko.wizb.it/docs/internals/events.html#roundtrip-event

askmike (commented):
Few questions!

plugins.js
crypto49er: You're right. Dependency removed.

plugins/blotter.js
crypto49er: Makes sense. Changed to async function.

plugins/blotter.js
crypto49er: I just did. I didn't realize the roundtrip event had p&l until after I wrote up this plugin, although I practically had to rewrite the plugin to use the roundtrip event. The benefit is the roundtrip event was more accurate for when sell amount doesn't match buy amount.

crypto49er:
All questions addressed.

mark-sch:
Which external software/service are you going to use with the exported blotter csv data?

PR #2720, opened 2019-02-06
BFX HF Indicators (44pcs)

rengel8:
PR contains a demo strategy showing how all available indcators (https://www.npmjs.com/package/bfx-hf-indicators) work including a compact doc on these package.  

I was not keen enough to find the document containing the side bar main navigation. Therefore this integration is missing. 

* **Other information**:


mark-sch:
Nice showcase of the honey framework indicator library. Will test it to replace some talib/tulip calculations for faster speeds...

rengel8:
> Nice showcase of the honey framework indicator library. Will test it to replace some talib/tulip calculations for faster speeds...

I made a comparison of the internal EMA against the ones of tulip and this HF. Tulip has some different rounding 6 digits after the comma as I recollect. It would be nice to create some table (./doc) as an overview on all available indicators over talib, tulip, internal and HF. So that it would be clear, which ones are missing and which ones produce expected results via automated tests.

Related to the performance I aspected some advantage beyond non native JS libs. But for me the calculation of all 44 indicators of this package rendered faster than in comparison to only one indicator from tulip.

mark-sch:
This is another node.js lib which could be interesting for indicator comparison: https://github.com/anandanand84/technicalindicators

PR #2721, opened 2019-02-06
Rejected trades are now recreated

crypto49er:

Bug fix

When a trade is rejected in Coinbase Pro because the price moved after Gekko sent the trade to the exchange, a new trade isn't recreated (can't retry the trade as the exchanged deleted it). Gekko holds asset that wasn't sold until the user intervenes by manually selling (and restarting Gekko). Rejected buy orders also require user intervention as Gekko thinks it bought the asset and won't respond to a new buy order until a sell order is issued.

Gekko will note the error in terminal, recalculate the price based on ticker info, and recreate the trade using new price information. 

* **Other information**:

### DIFF
--- a/exchange/orders/sticky.js
+++ b/exchange/orders/sticky.js
@@ -332,6 +332,13 @@ class StickyOrder extends BaseOrder {
       return false;
     }
 
+    if (error.message == 'post only error') {
+      this.price = this.calculatePrice(this.data.ticker);
+      console.log('Reissue trade order at', this.price);
+      this.createOrder();
+      return true;
+    }
+
     console.log(new Date, '[sticky order] FATAL ERROR', error.message);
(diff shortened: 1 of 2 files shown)


crypto49er:
Keeping this pull request alive. This is a serious issue that has lost money for me and others that use Coinbase Pro.

sboydell:
Agree with crypto49er here, this fix is needed for coinbase trades.

crypto49er:
This issue might have been fixed with this recent commit (https://github.com/askmike/gekko/commit/a143345a6a42f60bbb3c5e6520fce16fa1da9a39), but I want to keep this open until I or some one can confirm by forcing a trade that will be rejected in Coinbase Pro, instructions here: https://forum.gekko.wizb.it/thread-57832.html

PR #2740, opened 2019-03-13
Feature/mobile friendly

comlaterra:

New Feature

* **What is the current behaviour?** (You can also link to an open issue here)

The web client does not adapt to small screens, the menu is not fully visible and the content is not well visualized.

* **What is the new behaviour (if this is a feature change)?**

Includes some small changes in the web client to make it **more** mobile friendly by making the menu collapsible and adding a bit more padding in the content.

* **Other information**:

This is a really small suggestion, not introducing big changes but enabling better visualization of the Gekko web client.


askmike:
Will need to test effects on desktop first!

comlaterra:
I did test it on the desktop.
I don't expect to be any change. 
There is any kind of testing/validation I should have done before submitting the PR?

PR #2763, opened 2019-04-22
Allow turning on SSL in Docker

interfect:

This PR adds a feature (`USE_SSL` environment variable to control the Docker entrypoint) and documentation for it.

Currently there's no way to tell the docker entrypoint script to use SSL, so the Docker always produces a frontend that wants to talk to a "ws://" endpoint. This doesn't work behind an SSL-terminating reverse proxy.

Now you can set `USE_SSL=1` along with the `HOST` and `PORT` variables, to tell the Docker to serve a frontend that expects to be behind HTTPS.

* **Other information**:

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,3 @@
 HOST=localhost
 PORT=3000
+USE_SSL=0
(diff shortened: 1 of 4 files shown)


askmike:
:+1:

PR #2801, opened 2019-06-23
asynchronous strategy update function, backward compatible

massaroni:
New feature.

Strategy update function is strictly synchronous.

Now, your strategy's update function can be asynchronous or synchronous.  This uses a callback function, as per the convention in this project, and it's backward compatible with synchronous update functions.

Asynchronous update functions should return "true" to indicate that it's operating in asynchronous mode, and call the provided callback when it's done.  For example:

update: function (candle, done) {
  setTimeout(() => {
    // do stuff
    done();
  });
  return true;
}

### DIFF
--- a/plugins/tradingAdvisor/baseTradingMethod.js
+++ b/plugins/tradingAdvisor/baseTradingMethod.js
@@ -126,15 +126,29 @@ Base.prototype.calculateSyncIndicators = function(candle, done) {
       i.update(candle);
   },this);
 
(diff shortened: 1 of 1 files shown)


massaroni:
This feature would be super helpful if you need to hit a file, database, web service, ML service, etc, from within your strategy.

Anyone want to review this PR?

luciusinspace (approved):

PR #2807, opened 2019-06-27
Fix broken build on latest talib version

IAMtheIAM:
This previous script defaulted to using the latest version of `talib` available on NPM, however, the version as of now `1.1.2` is broken and won't build with gekko. This commit fixes the build by sticking to a known working version based on the DOCKERFILE `1.0.2`.

* **Other information**:


IAMtheIAM:
@askmike The TALIB versioning is still an issue that breaks the build. Please merge so new users can build successfully.

PR #2841, opened 2019-08-26
Addition of CCXT

MrEngineer13:
Update from original project


dcsan:
@MrEngineer13 why did you close this out of interest?

MrEngineer13:
The basics were covered in this PR but it still has some work to be done with updating the wrappers/importing to backtest/updating config and the market broker, have to get live trading up and running first

PR #2880, opened 2019-12-30
Fix vulnerable dependencies

jnikles:
Update lodash dependency due to npm audit warnings. Had to adjust .each loops primarily, since binding is not support in anymore 4.x, using arrow functions instead.

Gekko runs correctly

no new behavior introduced. Tests pass

* **Other information**:
- introduce arrow functions in _.each
- replace _.first(<array>, <counter>) with _.slice
- adjust variable names here an there


jnikles:
have to do some manual checks on top before this is ready

